PR #15, opened 2013-05-02
Test win2

garyfreder:
corrected indentation


halatmit:
@josmas @jisqyv @afmckinney 

What shall we do about this pull request?  finish it?  drop it?

jisqyv:
I've assigned myself. I'll look into it. I'm not sure we need this change (or maybe we already added it) as I have the buildserver running under Windows just fine...

PR #99, opened 2014-05-01
Label Format - HTML / Plain Text (default)

manishkhanchandani:
I created an additional property for Labels, i.e. LabelFormat, where user can choose which format to show for label. If user chooses HTML Text, then all the contents in label will render the text as html else it will render normally in plain text.

I uploaded my changes to google appengine, here is the url:
http://1.manishkk74.appspot.com/
If you create any label, you will see LabelFormat Option on the properties panel and if you choose HTML Text then the text will render in html in android phone.

I am not able to build server in google appengine, and no clear process is defined about how to do that.

Let me know if above explanation makes sense.


halatmit:
@josmas 
What shall we do about this pull request?  Finish it?  Drop it?

puravidaapps:
my suggestion is to add the label functionality as Manish prepared it without supporting the image tag, this is btw similar to the Notifier.ShowAlert block, an image also can't be displayed there ..
and to support the image tag also is not necessary, because we have the image component for that...
see also https://groups.google.com/d/msg/app-inventor-open-source-dev/ve5otz7mKFc/vwwNL-tIQfoJ
Taifun

jisqyv:
I also believe we should merge this change. I can finish up the cruft that needs to be done (add the Property to the TranslateComponentProperty module and related changes due to internationalization).

-Jeff

halatmit:
Thanks



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.


On Sat, Jan 24, 2015 at 11:04 AM, Jeffrey Schiller <notifications@github.com

jisqyv:
@manishkhanchandani I am going to close the PR shortly. I have taken the code and added support in the Designer (so people see in the browser what things will look like (sort of) on the phone). I will submit a new PR with both this code and my additions, rebased onto the latest version of App Inventor.

jisqyv:
This code has been updated and a new PR has been submitted (#785). Closing this one.

PR #121, opened 2014-09-26
typed blocks now auto fill default arguments

evanrthomas:
currently, blocks that have default arguments (like notifier.showChooseDialog) make default arguments when they are dragged form the pop-up area ,but not when they are typed in.

### DIFF
--- a/appinventor/lib/blockly/src/core/typeblock.js
+++ b/appinventor/lib/blockly/src/core/typeblock.js
@@ -17,6 +17,7 @@ goog.require('goog.events.KeyCodes');
 goog.require('goog.events.KeyHandler');
 goog.require('goog.ui.ac');
 goog.require('goog.style');
+goog.require('Blockly.Drawer');
 
 goog.require('goog.ui.ac.ArrayMatcher');
 goog.require('goog.ui.ac.AutoComplete');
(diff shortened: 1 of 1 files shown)


evanrthomas:
I handled "make color" and "Lookup in pairs". Also deleted the useless comment

appinventor/lib/blockly/src/core/typeblock.js
```diff
@@ -458,15 +459,7 @@ Blockly.TypeBlock.createAutoComplete_ = function(inputText){
...
-        if(!goog.object.isEmpty(blockToCreate.mutatorAttributes)) {
-          //construct xml
-          var xmlString = '<xml><block type="' + blockToCreateName + '"><mutation ';
-          for(var attributeName in blockToCreate.mutatorAttributes) {
-            xmlString += attributeName + '="' + blockToCreate.mutatorAttributes[attributeName] + '" ';
-          }
-
-          xmlString += '>';
-          xmlString += '</mutation></block></xml>';
+        if(xmlString = Blockly.Drawer.getDefaultXMLString(blockToCreate.canonicName, blockToCreate.mutatorAttributes)) {
```
josmas: Missing `var` declaration for xmlString variable <-- leaking into window object as a global; @evanrthomas can you please check?

evanrthomas:
@josmas I don't actually think it would leak, because this whole block of code is in one anonymous function. But it's bad practice anyway. Fixed

josmas:
@evanrthomas it's easy to check; just see if the value exists in the window object:

leaking_var_js

evanrthomas:
yea, it didn't end up on the window object because it was in a function. But I prefer using var anyway. I'm sure not using var is bad for some reason. Probably in some "best java script practices book" somewhere. lol. Anyway, I put the var in. Fixed

jisqyv:
@evanrthomas @josmas LGTM, Sending to Gerrit

jisqyv:
MERGED.

PR #123, opened 2014-09-29
Orientation

kkashi01:
Updated ScreenOrientation to include additional items such as ReverseLandscape, SensorLandscape, ReversePortrait, SensorPortrait, FullSensor


halatmit:
The way it is now, only the common orientations appear in the designer dropdown.   You can specify other ones with the [set orientation] block.   But  don't list these other ones in our documentation, and refer people to the Google Android docs for a full list and explanation.   I think that's an OK way to handle it.

Closing this issue.

PR #124, opened 2014-09-30
Updated ListView to return synched ListView.Selection and ListView.Selec...

kkashi01:
Updated ListView to return synched ListView.Selection and ListView.SelectionIndex. These were out of sync when Filtering was used.


jisqyv:
@kkashi01 LGTM. Note I added back the AI lists comment. I also had to make a small modification to deal with a change Hal made. The items in the AdapterArray are now Spannable Strings. I've pushed it up to Gerrit prior to merging and have asked Hal to take a quick look. But it looks OK to me.

-Jeff

jisqyv:
@kkashi01 MERGED

PR #134, opened 2014-10-06
Updated code to prevent auto triggering of AfterPicking on initialization

kkashi01:
Updated code to prevent auto triggering of AfterPicking on initialization


jisqyv:
@kkashi01 Sorry for the very long delay. I approved this change and it is merged into the master branch (after rebasing).  LGTM and Merged.

-Jeff

kkashi01:
@jisqyv Thanks

PR #185, opened 2014-12-20
New screen options: enable & disable ShowTitleBar & ShowStatusBar

weihuali0509:
I added new options in the screen, which allow App Inventor users to enable/disable ShowTitleBar & ShowStatusBar. These methods are in the form of the getter and setter methods in the block editor and also in the designer property list. But there won't be any change in the designer viewer if you decide to change them. Future work, make the changes reflected on the designer viewer.

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
@@ -2210,6 +2210,10 @@ public interface OdeMessages extends Messages {
   @Description("")
   String AboutScreenProperties();
   
+  @DefaultMessage("ShowTitleBar")
+  @Description("")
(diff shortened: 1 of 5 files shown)


halatmit:
@weihuali0509 
You might want to check with Dan Sandler at Google about 

(a) What are considered "best design pracrtices" in Enabling and Disabling the status bar and activity bar.  Is it consdered OK for apps change the visibility ar runtime.

(b) Is it worth it or even possible to make this work on pre-Gingerbread versions?

weihuali0509:
@halatmit  I am reading the doc about it, will resume working on this issue tomorrow. Thanks for your note. 

Wei

weihuali0509:
I will retire this. I will separate the ShowTitleBar & ShowStatus features into two pull requests. - Wei

PR #186, opened 2014-12-21
New Screen1 option: set AppName

weihuali0509:
In the current implementation, the default app name is the project name. In the current internationalization framework, we can't name the project to something other than English. This is not friendly for the people who want to name the app to something non-English. 

I added a new property, Screen1.AppName and default to the project name. This property will be saved in the project.properties file within the youngandroidproject folder and src/../../../xxx/scm file in the aia file. 

The app name is default to project name. But when Screen1.AppName is set, the display name of the installed app will use Screen1.AppName. I checked on the Android document, there is no restriction on the app name as long as it is a string (android:label). 

I have tested this with the following case - I uploaded an old project which doesn't have AppName property and the project loaded fine. If the AppName property is triggered, the system would write the values to the project property file. 

@halatmit @jisqyv @afmckinney @josmas @kkashi01 @app-inventor-dev@googlegroups.com


weihuali0509:
The previous error was from the misconfiguration of phantomJS at the local machine. All the Unit tests are  passed now. This is ready for review.  @halatmit @jisqyv

halatmit:
What should the behavior be when the App Name is blank?    If I load an old project, the App Name appears as blank.  And when I load that project onto the Phone, the Display Name is the project name.   If I then change the App Name to be Foo and reinstall the project, the Display Name appears (correctly) as Foo.   If I now edit the project to make the App Name blank and reinstall, the Display Name is still Foo.  Is this what we want?

Also, fix the merge conflicts before submitting another pull request.

weihuali0509:
Thanks Hal's comments. There was a bug. Here is the list of the correct behaviors - 

1. Load an old project, regardless of importing, or load from the designer, the default app name is the project name. 
2. Create a new project file, the default app name is the project name. 
3. Change the app name to "foo",  other than the project name, the displayed name of the installed app is  "foo" . 
4. Change the App Name to an empty string, the displayed name is the project name by default. 
5. If you have the App Name as the empty string, whenever you reload the designer page, the filed inside the AppName property will be changed back to the project name by default. 

I think this is ready for code review again. @halatmit

puravidaapps:
there are sometimes questions in the forum like this
https://groups.google.com/d/msg/mitappinventortest/1QAgwHVTPAs/bbRLdwyFpZMJ
so will app names in chinese or arabic or hebrew now be possible? 
what about spaces or special characters in the app name?
Taifun

halatmit:
Wei is working on this.

appinventor/components/src/com/google/appinventor/components/runtime/Form.java
```diff
@@ -1019,6 +1019,19 @@ public void VersionCode(int vCode) {
...
   }
+  
+  /**
+   * Specifies the App Name.
+   *
+   * @param aName the display name of the installed application in the phone
+   */
+  @DesignerProperty(editorType = PropertyTypeConstants.PROPERTY_TYPE_STRING,
+    defaultValue = "")
+  @SimpleProperty(userVisible = false,
```
halatmit: I'd expand this to add "if the App Name is blank, it will be set to the name of the project, when the project is built.

appinventor/components/src/com/google/appinventor/components/common/YaVersion.java
```diff
@@ -543,7 +545,9 @@ private YaVersion() {
   // - AboutScreen property was added
   // For FORM_COMPONENT_VERSION 13:
   // - The Screen.Scrollable property was set to False by default
-  public static final int FORM_COMPONENT_VERSION = 13;
+  // For FORM_COMPONENT_VERSION 14:
+  // - The Screen1.AppName was added.
+  public static final int FORM_COMPONENT_VERSION = 14;
```
halatmit: Make sure to add a comment saying the the AppName property was added and no block need to be change.  See the analogous comment in upgrading to versions 7, 8, and  9 above

halatmit:
Don't forget to regenerate the component documentation and update 
appinventor/docs/reference/components/userinterface.html

are you also going to include the Chinese names?

appinventor/components/src/com/google/appinventor/components/runtime/Form.java
weihuali0509: Fixed.

weihuali0509:
@halatmit, for the component documentation regeneration part, shouldn't the admin who in charge of the release doi that?

appinventor/components/src/com/google/appinventor/components/common/YaVersion.java
weihuali0509: Fixed.

weihuali0509:
@puravidaapps, you can name the app anything you want except empty name. The default name and the replacement for the empty field is the project name.

halatmit:
The we've been doing it is that the pull request invludes the updated
documentation (html) file.

Use

ant ComponentDocumentation

and the move the correct documentation line to the right file --
unserinterface.html, in this case, I think.

See

How to Generate the Appinventor Documentation



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

weihuali0509:
@halatmit updated the component documentation.

halatmit:
LGTM.  Can be moved to Gerritt.  But please check that this is internationlized correctly.
@afmckinney @jisqyv

jisqyv:
LGTM Merged

PR #198, opened 2014-12-28
Add margins to labels

halatmit:
@afmckinney @jisqyv 
I had thought I needed to wait for api4 to do this so it worked with different densities, but that's not necessary.

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
@@ -2370,6 +2370,10 @@ public interface OdeMessages extends Messages {
   @Description("")
   String GoogleVoiceEnabledProperties();
   
+  @DefaultMessage("HasMargins")
+  @Description("")
+  String HasMarginsProperties();
+
   @DefaultMessage("Heading")
   @Description("")
   String HeadingProperties();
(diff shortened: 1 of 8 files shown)


jisqyv:
@halatmit Should the default really be true for hasMargins? This results in existing projects "growing" margins when they are loaded for the first time into the new version (the version with this change). To me this violates the principle of least surprise.

Perhaps the right thing to do is to have the form upgrader (which we do have) update loaded projects to set hasMargins to false when it loads an existing project. That way newly created labels will default to having margins while existing projects will load and look proper.

-Jeff

halatmit:
Having the upgrader set margins to false sounds like a good solution.

But it adds another issue to the list of issues waiting for the upgrader.

Lyn and Andrew:  Can we put on a full court press and get the upgrader done
in the next couple of days?




== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
@halatmit This change doesn't have to wait for the upgrader. It can be done in YoungAndroidFormUpgrader.

halatmit:
Super!  that would be great.

Please do it.  Or do you want me to figure out how? I'm not exactly sure
how to do this.





== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
I'll take care of it. You can then do the final review.

-Jeff

halatmit:
Thanks.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
MERGED.  -Jeff

PR #201, opened 2014-12-30
MediaUtil Case-insensitive search caching

### LINKED ISSUE #188: MediaUtil is slow when filenames have case-insensitive match
When loading an image (say to a sprite's picture), if the image file and request string are case-different, it triggers a super-slow case insensitive search, but only when compiled to APK.

For a full discussion, see https://groups.google.com/forum/#!searchin/app-inventor-dev/app$20really$20slow$20when$20installed/app-inventor-dev/sPFLptUj_mc/EL22HVtMsbAJ

I just validated that this bug is still present, with messages such as this one in the logcat:
I/Choreographer( 5089): Skipped 40 frames!  The application may be doing too much work on its main thread.

We currently believe that 
* getting the list of all assets in compiled mode takes a long time
* the case-insensitive search once that list arrives is sub-optimal (but may be trivially so)
* the case-insensitive search is being called twice (that's a fact, and can be improved)

New information: the bug is most obvious in *newer* devices. The old 2.3.7 nexus 1 shows slight delay, but a fantastically powerful Samsung 2014 tablet has it much more so, to the point of getting in the way of app usage. Only newer devices generate the Choreographer logcat error. I submit that newer devices have made the critical slowness *even slower*, where we suspect that critical slowness comes from getting the list of assets from the OS. 

This apk demonstrates the problem: http://www.cs.uml.edu/~msherman/files/MediaUtilBugFinder.apk
(and is built in modern AI2).

@halatmit @jisqyv @mark-friedman @fgmart

marksherman:
Fixes #188 

Speeds are better, even in the correct case!

(average operation times)

Before change:
Good names: 9 ms
Bad names: 704 ms

After change:
Good names: 6 ms
Bad names initial load: 202 ms
Bad names subsequent: 6ms

Cases where the path is not found are not cached. Average op time for this case is 175 ms.
(This is the null-return case from findCaseinsensitivePath)

This uses a ConcurrentHashMap, with no eviction mechanism. The hash table can grow without bound, adding an entry for each unique path that is requested.

### DIFF
--- a/appinventor/buildserver/src/com/google/appinventor/buildserver/BuildServer.java
+++ b/appinventor/buildserver/src/com/google/appinventor/buildserver/BuildServer.java
@@ -147,20 +147,30 @@ public class BuildServer {
   // The zip file where we put all the build results for this request.
   private File outputZip;
 
-  // Set to true if we are shutting down
-  private static volatile boolean shuttingDown = false;
+  // non-zero means we are shutting down, if currentTimeMillis is > then this, then we are
+  // completely shutdown, otherwise we are just providing NOT OK for health checks but
+  // otherwise still accepting jobs. This avoids having people get an error if the load
+  // balancer sends a job our way because it hasn't decided we are down.
+  private static volatile long shuttingTime = 0;
 
   private static String shutdownToken = null;
+
+  private enum ShutdownState { UP, SHUTTING, DOWN };
 
   @GET
   @Path("health")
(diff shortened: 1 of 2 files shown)


jisqyv:
@marksherman Under review now.

jisqyv:
@marksherman LGTM. Merge imminent.

jisqyv:
MERGED.

PR #212, opened 2015-01-07
Trac 617: change in MockComponents to let Button height in designer  cha...

osmidy:
...nge with text
@halatmit @afmckinney In the method getPreferredSizeOfDetachedWidget of MockComponentsUtil, a value of 4 is added to getOffsetWidth; I did the same with getOffSetHeight, using 10 as an arbitrary value.  I'm not sure how to choose a correct value to offset, or how adding something in this class affects the appearance other designer components using MockCompoenentsUtil.


halatmit:
@afmckinney  @osmidy 

Don't you also need to change getPreferredSizeOfDetachedWidget to match.  While you're at it, your should define constants for that 4 and that 10, to make the code cleaner.

I'm not sure what the implications of your change are.   So do a lot of testing to make verify that things look OK.  Test on ordinary screen size (in the designer) and tablet size as well.   In the end, it won't be possible to make what you see in the designer completely match what's on the phone or tablet.  Just try to do OK.  And note the new option for HasPadding on labels.   That should also look ok in phone  vs. designer -- but I would not worry about modifying the mock component to account for the padding property.

halatmit:
@osmidy @jisqyv @afmckinney 
There is something broken about this commit because it claims there are 121 changed files.  Please work with andrew or jeff to clean this up, and then one of us can review it.   thanks.

PR #217, opened 2015-01-12
Add ZXing QR Code Library to MIT App Inventor.

jisqyv:
Add ZXing QR Code library (version 2.1) directly to App Inventor so we
can scan QR Codes without requiring an additional Scanning App.

Change-Id: I3dbb603e20129fa25116555b8ae71a018e68cd0d


jisqyv:
@halatmit @afmckinney @weihuali0509 Here is the pull request to add ZXing code internally to App Inventor. We add a "UseExternalScanner" property to the BarcodeScanner component. It defaults to "true", so an external scanner is still used. If the App Inventor programmer clears this flag, either in the designer or at runtime via the blocks, an internal version of ZXing is used.

This code has been in production in the Companion for about 2 years. It works :-)

weihuali0509:
Thanks! I will take a look later tonight. - Wei

afmckinney:
LGTM Merged

PR #222, opened 2015-01-16
New screen option: TitleVisible

weihuali0509:
I added new option in the screen, which allowes App Inventor users to hide/show Title. This method is in the form of the getter method in the block editor and as property in the designer property list. But there won't be any change in the designer viewer if you decide to change them. Future work, make the changes reflect on the designer viewer. 

We can't really implement the method which Dan (from Google) has suggested -  use Window.requestFeature() with FEATURE_NO_TITLE . Because the property is called after the setContentView. This approach will crash the app. We can change it in the manifest, "via the window theme, e.g@android:style/Theme.Black.NoTitleBar", but this hides all screens' title bar. 

I want to give the maximum design freedom to App Inventor users, so I have proposed the following method. All the API calls exist since API level 1. 

The old commit is here - 
https://github.com/mit-cml/appinventor-sources/pull/185


weihuali0509:
@halatmit @jisqyv @afmckinney

halatmit:
1.  is this called the title bar or the action bar?

2. I tried this and setting showtitlebar to false removes the title, but it
does not remove the gray bar itself.
can you show me a demo of how it is supposed to work?

3. I know you want to give people maximum control, but I suggest that the
useful simple case is to either have no bars (fullscreen option) or to have
both bars.




== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

weihuali0509:
1. This is called Titlebar in the old device (<= 2.3), Action bar in the newer device. 
2. Are you running on Lollipop? 
3. How about I do it at the manifest level? so it is one for all the screens? It is much easier, simpler that way?

FullScreen in Android means different thing => no title bar (action bar) + no status bar + plus no home key panel. Do you mean by no title bar + no status bar? @halatmit

halatmit:
If we have to choose, I suggest the modern nomenclature (action bar).   But think about my suggestion to just have either both bars or none (fullscreen).   I don't know about implentation.  It would be good to dynamic contol rather, which I'm guess means it cannot be done in the manifest.  See if this runtime method works:

http://stackoverflow.com/questions/2868047/fullscreen-activity-in-android

weihuali0509:
Thanks for the pointer. There is no action bar in App Inventor, only title bar... 

Dynamic control can't be done at the manifest. 

I already saw and tried the openstack approach. It doesn't really work in App Inventor. The code " requestWindowFeature(Window.FEATURE_NO_TITLE);" must be called before the "setContentView" (within Form.onCreate()). If I have a property "ShowTitleBar" in the property list in the Designer and has the corresponding setter method "requestWindowFeature(Window.FEATURE_NO_TITLE)", this will crash the app. The method "requestWindowFeature(Window.FEATURE_NO_TITLE);", can't be call after the setContentView. 

The approach I am using now seems have compatible issue with Lollipop, which I didn't test. Did you try out anything lower than 5.0 to see if that works? @halatmit

halatmit:
The emulator I tried it on is an old system, pre-lollipop. It's the
standard emulator we ship with app inventor.  (gingerbread, or earlier?)

Let's find a dynamic approach that works, if possible.  And is there a good
reason to have separate properties rather than just fullscreen vs. regular?







== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

weihuali0509:
@halatmit I verified this and this approach doesn't work on Lollipop. I will try to finish this pull request next week.

halatmit:
"title bar" is obsolete.  We should be using "action bar" and making sure
this operation is consistent with Android design practices.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

kkashi01:
ActionBar unfortunately doesn't work with API3. I believe its API 11

weihuali0509:
At this point, I don't see a clean solution to set it dynamically. Give the fact that we don't have action bar in App Inventor (still title bar), I decide to move this setting to the manifest. The ShowTitleBar will be an option of the Screen1.property.

halatmit:
We should talk with the Android people about suggestions for handle
this.   Let's not build in things that will be obsolete with Lollipop or
future versions.




== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

weihuali0509:
@halatmit what do you mean? Back in Dec, we had emails with Dan from Google. He recommends to do it in the manifest. Title bar is in App Inventor, Action Bar is NOT. Doing in the Manifest (set the right theme) just involves in checking in the android version: theme A for pre material, and theme B for material or later.

halatmit:
I'm getting confused.

Do modern (Lollipop) android apps have title bars?  I thought they did
not.   Let's talk through the complete design a little, before you go ahead.

Also, see here as a possible alternative:

http://www.androidsnippets.com/how-to-make-an-activity-fullscreen




== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

weihuali0509:
I double checked again. The previous solution works on Android 5.0, 
lollipop as well. Can you try the attach apk and provide comment?

Wei
wli17@mit.edu

weihuali0509:
I double checked again. The previous solution works on Android 5.0,
lollipop as well. Can you try the attach apk and provide comment?

https://www.dropbox.com/s/azks3y9sww1ldnq/AndroidTitleBar.apk?dl=0

Wei
wli17@mit.edu

josmas:
If the API 4 code is merged, we can start using [Support Library](http://developer.android.com/tools/support-library/index.html) which give access to ActionBar. That could solve the compatibility issue here.

weihuali0509:
@halatmit try the new code again (for ShowTitleBar) ?

If you prefer, use the following aia and apk for testing:
https://www.dropbox.com/sh/nuvwd6snrfra1oq/AAAPfN11WCwjt-70Du_2w-_ua?dl=0

Thanks,

Wei

weihuali0509:
@halatmit @jisqyv This is updated and should be ready to review and merge.

weihuali0509:
From the last conversation, we agreed to NOT have a special property - ShowTitleBar; instead, we are going to have a *check* box right next to the *Title* property, this checkbox will serve as the function for the *ShowTitleBar* .

weihuali0509:
The updated aia and apk files are here - 
https://www.dropbox.com/sh/nuvwd6snrfra1oq/AAAPfN11WCwjt-70Du_2w-_ua?dl=0

weihuali0509:
We agreed to have a "checkbox" right next to the *Title* property in the designer earlier; however, this is not doable because gwt doesn't allow me to overloading the method in the designer - Title(boolean) / Title(string). The error message is the following - 

* Inconsistent types java.lang.String and boolean for property Title in component Form * 

Instead, I renamed the method to *TitleVisible*, which hope to be consistent with the *Title* property. In the designer, the property *Title* should be next to the *TitleVisible*.

halatmit:
Works fine.
I also tested it on 2.2.2 and it works there.
Do you want to add the translations for TitleVisible ?

jisqyv:
@weihuali0509 Also check on whitespace issues in the code

weihuali0509:
@halatmit @jisqyv added the translations.

weihuali0509:
@jisqyv can you point out the issues with the while spaces?

jisqyv:
@weihuali0509 @halatmit Never mind about the whitespace if you cannot find it. I'll take care of it when I merge the change (likely today or tomorrow assuming my tests work).

weihuali0509:
@halatmit @jisqyv I believed I have cleaned all the whitespaces.

halatmit:
LGTM

halatmit:
@weihuali0509 @jisqyv
I think this needs some more discussion.  I don't like the idea of not being able to use the action bar as we go forward.   Will this change prevent use of the action bar?

jisqyv:
@halatmit @weihuali0509 I have verified that this PR does not conflict with the work on MenuItems. We should merge it.

(I meant to do that a few days ago!)

jisqyv:
@halatmit @weihuali0509 MERGED into the ucr branch for the next component release.

PR #223, opened 2015-01-16
New screen option: ShowStatusBar

weihuali0509:
I added new option in the screen, which allowes App Inventor users to enable/disable ShowStatusBar. This method is in the form of the getter method in the block editor and as property in the designer property list. But there won't be any change in the designer viewer if you decide to change them. Future work, make the changes reflect on the designer viewer. 

I am implement the method in the following way - use getWindow().clearFlags(WindowManager.LayoutParams.FLAG_FULLSCREEN). We can also change it in the manifest,  but this hide all the status bar. 

I want to give the maximum design freedom to App Inventor users, so I have proposed the following method. All the API calls exist since API level 1. 

The old commit is here - 
https://github.com/mit-cml/appinventor-sources/pull/185


weihuali0509:
@halatmit @jisqyv @afmckinney

halatmit:
I tried this, and it does not seem to work for me, testing on Lollipop.  If I have an app with ShowStatusBar checked, I don't see a status bar.  However, I can swipe down over the title bar and the status bar is shows temporarily.   This behavior is the same whether or not showSatusBar is checked.

On the emulator, the status bar is shown, but the thing controlled by ShowStatusBar appears to be the Title, not the status bar.
can you give me a demo of how this is supposed to work?   Also, as I think about it, is there a reason to have separate controls for title and status?  Why not have a single FullScreen property that controls them both?

weihuali0509:
@halatmit I verified this and this approach doesn't work on Lollipop. I will try to finish this pull request next week.

weihuali0509:
I double checked again. The previous solution (for status bar, this pull request) works on Android 5.0,
lollipop as well. Can you try the attach apk and provide comment?

https://www.dropbox.com/s/fvkwfycz0rdrbr5/AndroidStatusBar.apk?dl=0

Wei
wli17@mit.edu

weihuali0509:
I double checked again. The previous solution (for status bar, this pull request) works on Android 5.0,
lollipop as well. Can you try the attach apk and provide comment?

https://www.dropbox.com/s/fvkwfycz0rdrbr5/AndroidStatusBar.apk?dl=0

Wei
wli17@mit.edu

appinventor/docs/reference/components/userinterface.html
```diff
@@ -930,6 +930,8 @@ <h2 id="Screen">Screen</h2>
   <dd>The requested screen orientation. Commonly used values are unspecified (-1), landscape (0), portrait (1), sensor (4), and user (2).  See the Android developer docuemntation for ActivityInfo.Screen_Orientation for the complete list of possible settings.</dd>
   <dt><code>Scrollable</code></dt>
   <dd>When checked, there will be a vertical scrollbar on the screen, and the height of the application can exceed the physical height of the device. When unchecked, the application height is constrained to the height of the device.</dd>
+  <dt><code><em>ShowStatusBar</em></code></dt>
+  <dd>The status bar is the toppest bar on the screen. You have the option to hide the status bar. This function returns you the visibilty of the status bar.</dd>
```
halatmit: toppest --> topmost

halatmit:
The apk you provided seems to work.  But when i pull down this commit and build it, I have problems.

(1) If I start a new project then showStatusBar is checked in the designer, but the phone doesn't show a status bar, and if I go to the blocksk editor and evaluate the showStatusBar property it returns false.

(2) Also, there is no setter for the showstatusBar property in the blocks.   Is that indended?

weihuali0509:
I think I know where the confusion is from. 

(1) Are you using the MIT AI companion to develop? There is no problem for the installed the apk, but I see the same issue you provided on the MIT AI companion. 

(2) Yes, back then, you told me we should not add more methods to the setter methods draw until we figure out a better way to show them?

halatmit:
OK.  Fix the companion issue so this works in live deevelopment.

I wasn't clear when  made that comment.  I didn't mean the setter in
particular, but rather that we should not add more property blocks (setters
or getters) and just put them in the
dropdowns.  But maybe I'm wrong and we should also have the blocks (and
also have them in the dropdowns).

Let me know when I can do another review.





== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

weihuali0509:
@halatmit  try the new code again?

If you prefer, use the following aia and apk for testing:
https://www.dropbox.com/sh/4hl50hbd32kln77/AACXRkBnuN3zRFm9dLX0O6wia?dl=0

Thanks, 

Wei

appinventor/components/src/com/google/appinventor/components/runtime/Form.java
```diff
@@ -783,6 +784,38 @@ public String AboutScreen() {
...
     this.aboutScreen = aboutScreen;
   }
+  
+  /**
+   * ShowStatusBar property getter method.
+   *
+   * @return  showStatusBar boolean
+   */
+  @SimpleProperty(category = PropertyCategory.APPEARANCE,
+      description = "The status bar is the toppest bar on the screen. You have the option to hide the status bar. This function returns you the visibilty of the status bar.")
```
halatmit: Change description to:
The status bar is the topmost bar on the screen.  This property reports whether the status bar is visible.

appinventor/docs/reference/components/userinterface.html
```diff
@@ -936,6 +936,8 @@ <h2 id="Screen">Screen</h2>
   <dd>The requested screen orientation. Commonly used values are unspecified (-1), landscape (0), portrait (1), sensor (4), and user (2).  See the Android developer docuemntation for ActivityInfo.Screen_Orientation for the complete list of possible settings.</dd>
   <dt><code>Scrollable</code></dt>
   <dd>When checked, there will be a vertical scrollbar on the screen, and the height of the application can exceed the physical height of the device. When unchecked, the application height is constrained to the height of the device.</dd>
+  <dt><code><em>ShowStatusBar</em></code></dt>
```
halatmit: Regenerate the comments, so this gets the new description of the property.  Or just change this <dd> manually.

weihuali0509:
@halatmit regenerated the documentation.

appinventor/components/src/com/google/appinventor/components/common/YaVersion.java
```diff
@@ -571,6 +573,8 @@ private YaVersion() {
   // - The Screen.Scrollable property was set to False by default
   // For FORM_COMPONENT_VERSION 14:
   // - The Screen1.AppName was added and no block need to be changed.
+  // For FORM_COMPONENT_VERSION 15:
+  // - The Screen.ShowStatusBar was added.
   public static final int FORM_COMPONENT_VERSION = 14;
```
jisqyv: FORM_COMPONENT_VERSION = 15 (!!)

jisqyv:
You also have to update versioning.js. Search for "Screen" (near line 1701). What you need to change should be obvious.

-Jeff

weihuali0509:
@halatmit @jisqyv This is updated and should be ready to review and merge.

halatmit:
works fine, also works fine on Froyo
please add the translations for the new property

weihuali0509:
@halatmit @jisqyv added the translations.

halatmit:
LGTM

jisqyv:
@weihuali0509 MERGED onto the ucr branch for the next component release

PR #225, opened 2015-01-18
 Add host[:port] property in appengine-web.xml to tell build server app host url when deploy it on other GAE implementation such as AppScale.

wanddy:
Add host[:port] property in appengine-web.xml to tell build server app host url when deploy it on other GAE implementation such as AppScale.
Signed-off-by: wanddy <wanddy@163.com>


halatmit:
Thanks!



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
The logic here looks wrong. If you set the default value of appengine.host to "localhost:8888" then it will use it, even in production (because you test with appengineHost.get()==""). Perhaps the default should be the empty string both in the code and in the appengine-web.xml file. A comment in appengine-web.xml can explain how to use it.

-Jeff

wanddy:
I have reviewed my code and configuration,and test on Appscale and GAE,it works fine by setting appengine.host or remain blank.

appinventor/appengine/src/com/google/appinventor/server/project/youngandroid/YoungAndroidProjectService.java
```diff
@@ -114,6 +114,9 @@
   // host[:port] to use for connecting to the build server
   private static final Flag<String> buildServerHost =
       Flag.createFlag("build.server.host", "localhost:9990");
+  // host[:port] to tell build server app host url
+  private static final Flag<String> appengineHost =
+      Flag.createFlag("appengine.host", "localhost:8080");
```
jisqyv: @wanddy Perhaps the default here should be "" instead of localhost:8080. In theory this default will never get used because the appengine-web.xml file sets the property to "". This is why I was confused about whether or not the code would work.

-Jeff

P.S. Sorry for the delay in all of this. Been busy :-)

appinventor/buildserver/src/com/google/appinventor/buildserver/Compiler.java
```diff
@@ -319,7 +320,7 @@ private boolean writeAndroidManifest(File manifestFile, Set<String> permissionsN
 
     // TODO(user): Use com.google.common.xml.XmlWriter
     try {
-      BufferedWriter out = new BufferedWriter(new FileWriter(manifestFile));
+      BufferedWriter out = new BufferedWriter(new OutputStreamWriter(new FileOutputStream(manifestFile), "UTF-8"));
```
jisqyv: @wanddy This change should probably be in its own PR so it winds up in its own commit.

jisqyv:
@wanddy Sorry for the delay, LGTM. Currently squashed and submitted to Gerrit. Should be merged shortly.

jisqyv:
@wanddy Note: I fixed some whitespace issues

jisqyv:
MERGED

PR #226, opened 2015-01-18
Fix taglines

halatmit:


halatmit:
@jisqyv 
I decided to go ahead and let people remove the Yandex tagline along with the MIT tagline, i.e., treat them both the same by addinng an HTML open comment in the about string.

jisqyv:
@halatmit LGTM Moved to Gerrit (Jenkins building now)

jisqyv:
@halatmit MERGED

PR #227, opened 2015-01-19
Enhance the Blocks Upgrader Code

jisqyv:
Add a test for YOUNG_ANDROID_VERSION and bypass upgrade code if
loading project has same value as the system. Some minor style fixups
in versioning.js. Also updated BlocklyEvalTests so that the blocks
language and young android version do not have to be hardcoded into
the units tests. They are read from YaVersion.java at runtime.

Change-Id: I593a652a36373573125a3b2e28d9c842455e0499


jisqyv:
@fturbak @halatmit @afmckinney Please review. This change bypasses most of the upgrade path when a project doesn't need to be upgraded, makes loading a bit faster.

halatmit:
LGTM

PR #229, opened 2015-01-20
Add caching to frequent server operations

jisqyv:
Reduce the number of calls to the data store by using memcache to hold
active files. Upgrade Objectify library to version 3.1

Change-Id: Iac4fb43e205c4f4fe54e7646d463e694bae5fe38


appinventor/appengine/src/com/google/appinventor/server/storage/ObjectifyStorageIo.java
```diff
@@ -222,7 +222,7 @@ public void run(Objectify datastore) {
           user.setUserTosAccepted(userData.tosAccepted || !requireTos.get());
           user.setSessionId(userData.sessionid);
         }
-      });
+      }, false);
```
halatmit: It might be useful to put comments at the places where useTranscaction is false, noting this, and saying why transactions are not necessary in those cases.

appinventor/appengine/src/com/google/appinventor/server/storage/StoredData.java
```diff
@@ -136,7 +141,7 @@
 
   // Project files
   @Unindexed
-  static final class FileData {
+  static final class FileData implements Serializable {
```
halatmit: Please add a comment explaining why the change.

halatmit:
A couple of suggestions to add comments, but otherwise LGTM

Issue #208, opened 2015-01-04
Transfrom CRLF to \n


halatmit:
Change blocklyeditor/src/generators/yail.js, quotifyForREPL  so that CR followed by LF is treated the same as LF, i.e., transformed to an unescaped \n

This is to handle lines read in from Windows systems.
See forum issue

https://groups.google.com/forum/#!category-topic/mitappinventortest/tzoNdh4HyYY

cdalsass:
I think the proper place to fix this is within the CSV parsing functions such as CsvUtil.java. quotifyForREPL deals with escaping strings from Blockly passed to REPL, and doesn't deal with what's inside CSV files. 

FYI: feel free to assign this issue to me.

halatmit:
I agree:  It seems like a simple change to quotifyForREPL

halatmit:
Assigned (and invited as a developer)



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.
MIT Dept. of Elec. Eng. and Comp. Sci.
MIT Media Lab, Second Class Faculty

cdalsass:
I don't think this is a bug. Rather, the original poster seems to have been confused about the API. (He should not have been putting CRs or LFs into a CSV "row" unless it was surrounded by quotes). This led to discussion which got somewhat off track from original problem. See my response at https://groups.google.com/d/msg/mitappinventortest/tzoNdh4HyYY/Q1hc3xMoUOgJ

halatmit:
Whether or not it's the original issue with CSV, the problem to fix here arises from reading a file from a Windows machine, which ends lines with CR LF, unlike Mac and Linux, which ends them with just LF.   The solution we decided on is for the parser to transform occurrences of CR followed by LF into and unescaped LF, just as if it had read an LF without the CR.

cdalsass:
I do understand the problem. However, I disagree that quotifyForREPL in yail.js is the best place to make changes. (But maybe I am misunderstanding the code - I apologize if that's the case). 

Question: Has there has been a persistent problem in AI2 with newline issues, prior to this post? (From reading the original Google post, it appears to me the bug was just noticed in December - which contradicts that.)

In my test case, I show that both Windows and Unix files are both handled properly in the context of CSV row parsing. But I think you're saying that **any time a file is read**, we should normalize return characters. 

[...]

PR #230, opened 2015-01-21
Bug fix for git issue #208.

cdalsass:
I'd like someone to review changes for:

File ReadFrom: Normalize end of line characters to make dealing with text files uniform across platforms.

It's a straightforward change, but still a bit concerned that:

1. People may want to detect Windows files (and find they no longer can, because we've obscured any differences)
2. Someone reading in a file, then writing it out, can expect a change in size or checksum if it contained CRLF when opened.
3. Someone reading in a file from different source, like URL, may expect same functionality. (They won't get it with this fix).

Please note that **AI Companion must be upgraded** for this change to take effect.

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/File.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/File.java
@@ -240,6 +240,20 @@ public class File extends AndroidNonvisibleComponent implements Component {
     });
   }
 
+  /**  
+   * Replace Windows-style CRLF with Unix LF as String. This
+   * allows end-user to treat Windows text files same as Unix or Mac. In future, 
+   * allowing user to choose to normalize new lines might also be nice - in case someone really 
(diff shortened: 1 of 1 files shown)


halatmit:
@afmckinney @jisqyv @cdalsass 
LGTM

jisqyv:
@cdalsass The commit author has test@test.com as your email address. Can you update the commit to contain a real email address. Thanks.

-Jeff

cdalsass:
Sure Jeff. I'll update in about hour when I get home.

-----Original Message-----

cdalsass:
Just updated the commit author email in (github) remote, but not sure if I need to resubmit pull request. Can you see updated email?

jisqyv:
@cdalsass No problem, I'll fetch the e-mail address from the commit.

jisqyv:
@cdalsass MERGED.

PR #234, opened 2015-01-23
List view font size

cooper-sloan:
Added fontsize property to ListView.  Font size is capped at 1000.

@halatmit


jisqyv:
OdeMessages, and TranslationCOmponentProperty also have to be updated to add a new Property to a component. I can take care of this though.

-Jeff

halatmit:
Thanks.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
@cooper-sloan Why did you use "float" for the textSize property?

-Jeff

cooper-sloan:
I looked at some other components (eg Label) and they used a float for the
fontSize, so I was just following.  Integer would probably make more sense.

    - Cooper

jisqyv:
@cooper-sloan Indeed. But in Label, the low level TextView accepts a float, while in this case the SpannableString takes an integer. I would suggest making the change.

-Jeff

cooper-sloan:
I changed textSize to an integer.

appinventor/components/src/com/google/appinventor/components/runtime/ListView.java
```diff
@@ -405,5 +412,33 @@ public void TextColor(int argb) {
...
+
+  /**
+   * Specifies the ListView item's text font size
+   *
+   * @param integer value for font size
+   */
+  @DesignerProperty(editorType = PropertyTypeConstants.PROPERTY_TYPE_NON_NEGATIVE_FLOAT,
+      defaultValue = Component.FONT_DEFAULT_SIZE + "")
+  @SimpleProperty
+  public void TextSize(float fontSize) {
```
jisqyv: You should probably change this argument to int as well (and update the PropertyTypeConstants to integer as well.

-Jeff

cooper-sloan: Will do.  ( I was worried about users overflowing the integer input but I
see that PropertyTypeConstants deals with that. )  I changed this as well
as the default font size to an integer value.  ( Component uses a float
currently)

jisqyv:
@cooper-sloan I'm getting ready to merge this change. I'm going to change the default text size to 22. This appears to be the value needed to have things look as they had before this change (if the programmer doesn't explicitly set the TextSize property).

halatmit:
Sounds good.   Somewhere, there's a android list of a very small number of
recommended sizes.    We should use those eventually, but not till after we
switch to API leveel 4 (and maybe density independent fonts).



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
@cooper-sloan LGTM, MERGED.

cooper-sloan:
Great!

PR #239, opened 2015-02-01
Add a getter for ContactUri from ContactPicker's result

Eagles2F:
I propose this getter because of the issue below:
    I was creating a android system intent "contact" based app. The contactPicker which is provided by AppInventor is working fine. Howvever, I want to click a button and view a specific contact. To achieve that, I need to use Activity Starter to do an action which is: android.intent.action.VIEW.   This activity will need a input: DataUri.      
        Here is the problem. I cannot find out what is the DataUri for a contact so I cannot do the call to view a contact.     
        Thus, since we already have this contactPicker and inside the code I learned that we actually have the contactUri but we somehow haven't expose it. I propose to make a getter for this ContactUri. And then we could view the contact in the program. 
         Here are the screen shots of my test:
             1.png->2.png->3.png
         The new block in ContactClicker: 4.png

1.png

2.png

3.png

4.png

       ant tests also built successfully.    

By Yifan(Evan) Li (thueeliyifan@gmail.com)


Eagles2F:
I'm aware of the problem in the encoding and extra eclipse files added into the code base. I will close the request and do the fix.

Eagles2F:
I'm aware of the problem in the encoding and extra eclipse files added into the code base. I will close the request and do the fix.

PR #240, opened 2015-02-01
add a getter of ContactPicker for getting contactUri

Eagles2F:
I propose this getter because of the issue below:
I was creating a android system intent "contact" based app. The contactPicker which is provided by AppInventor is working fine. Howvever, I want to click a button and view a specific contact. To achieve that, I need to use Activity Starter to do an action which is: android.intent.action.VIEW. This activity will need a input: DataUri.

Here is the problem. I cannot find out what is the DataUri for a contact so I cannot do the call to view a contact.

Thus, since we already have this contactPicker and inside the code I learned that we actually have the contactUri but we somehow haven't expose it. I propose to make a getter for this ContactUri. And then we could view the contact in the program. 
Here are the screen shots of my test:
1.png->2.png->3.png
The new block in ContactClicker: 4.png

1.png

2.png

3.png

4.png

ant tests also built successfully.

By Yifan(Evan) Li (thueeliyifan@gmail.com)

### DIFF
--- a/appinventor/aiplayapp/youngandroidproject/project.properties
+++ b/appinventor/aiplayapp/youngandroidproject/project.properties
@@ -7,12 +7,12 @@ icon=AI2Companion-Icon.png
 # The Version code is an integer. Each new version of an App uploaded
 # to the Play Store must have a version greater then the version in
 # the store
-versioncode=223
+versioncode=224
 # The Version Name is displayed to the user and can contain numbers
 # and letters. It generally should be congruent to the Version Code
 # Terms used by AI2:
 # Version code XYZ turns into X.YZ. ai2 indicates a Companion for AI2
 # and an optional "zx1" indicates it has internal Zebra Crossing (QR
 # Code) scanning builtin.
-versionname=2.23ai2zx1
+versionname=2.24
 useslocation=False
--- a/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
@@ -3684,6 +3684,10 @@ public interface OdeMessages extends Messages {
   @Description("")
   String LongClickEvents();
 
+  @DefaultMessage("RequestFocus")
+  @Description("")
+  String RequestFocusMethods();
+
   @DefaultMessage("LostFocus")
   @Description("")
   String LostFocusEvents();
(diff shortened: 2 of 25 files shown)


halatmit:
@jisqyv @afmckinney @Eagles2F @josmas 

Yifan:
Nice job coding this.
I'm a little wary of putting something into App Inventor that will be useful only with the activity starter, and only for viewing the contact.  What are some other things that App Inventor users could do, once they can get the ContactUri? If it's not a lot, then maybe the operation should be to view the contact, rather than return the Uri.  That might be simpler for our users, although it's not as flexible.

Let us know and we can do a proper review.

José, do you have an opinion on this?

Thanks!!

josmas:
@halatmit @Eagles2F We are going to think about a way of making this less low level. We both agree it is something not many users would go for. We'll have a think about it.

Eagles2F:
@halatmit @josmas    guys, I've created a proposal for a more easy-to-use solution.
https://docs.google.com/document/d/14u_HMTyrzB9AE3Tr4Y85Jc9kPXlvQ7ZDkCU6SJfArSk/edit?usp=sharing

halatmit:
I'm cc-ing this to app-inventor-dev because I think it needs wider
discussion.

This feels a little too restrictive to me.   I'd like to be able to do
something with the selected contact other than show it immediately at the
time of picking.   Maybe what's needed is an operation that can view a
contact.   The  issue then is how to represent the contact.  One idea would
be to have the contact be a string (the URL).   Another would be to make it
a new data type.   This needs a bit of mulling over.


== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

halatmit:
Here's a variation on the Yifan's second idea.

Implement a showContact method, that takes a contact, represented as a
string (the url).

Change after-picking so that it has a callbacks the url, the contract name,
email, and email address list.  And deprecate these three as properties.




== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

Eagles2F:
I agree with Hal about the latest method.    For me it now looks like adding two thing: showContact for user easy usage and contactUri from the picking activity .    Does this make more sense?    

Probably, we need to bring other people's idea on this.  I would love to write the code if we get an agreed solution.

halatmit:
That would be great if you wrote the code.   Look at something like the ImagePicker block to see how to return a value as a callback.   Meantime let's see what other people suggest.

Eagles2F:
Hi ,   I added a ViewContact method which take a string URI to open the contact detail. Plz have a look.

jisqyv:
This code is now handled in #445.

PR #241, opened 2015-02-01
Speed Up Blockly on Firefox

jisqyv:
This small fix comes from Neil. see:

 https://github.com/google/blockly/commit/f79b42ab70e2c6eda7beb6777cec76db654332a4

Change-Id: I2559fe33fb261243e5a4c985bddfda59ff103a3a


jisqyv:
@afmckinney @halatmit Here is the firefox speedup fix from Neil

halatmit:
LGTM

Issue #242, opened 2015-02-02
Type blocking is broken on current master


halatmit:
Typeblocking does not work.  And any When blocks created by type blocking appear as black blobs,
This problem exists on the current master, but no on the current release.
What things got merged between then and now?

jisqyv:
@halatmit I'll have a pull request to fix this shortly. Just testing it out now.

jisqyv:
Fix has been merged.

PR #243, opened 2015-02-02
Bugfix to typeblock, fix component-event blocks.

jisqyv:
Bugfix to typeblock.js to work correctly on component-event blocks.

Change-Id: I4e152bf7086ab3b6c4a646a8b7cde8868feb5ebc

### DIFF
--- a/appinventor/lib/blockly/src/core/typeblock.js
+++ b/appinventor/lib/blockly/src/core/typeblock.js
@@ -459,18 +459,22 @@ Blockly.TypeBlock.createAutoComplete_ = function(inputText){
         blockToCreateName = blockToCreate.canonicName;
         // components have mutator attributes we need to deal with. We can also add these for special blocks
         //   e.g., this is done for create empty list
-        var xmlString;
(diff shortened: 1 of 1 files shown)


jisqyv:
@halatmit @josmas This fixes the problem referenced in #242

halatmit:
@jisqyv 

Tried it and it seems to work.
LGTM

by the way, what broke things?

jisqyv:
@halatmit Evan converted the code to call Blockly.Drawer.getDefaultXMLString(). However that call doesn't provide default code for every block, in which case it returns null. His code for constructing the block was just not right (and I didn't catch it in the review). The code there now is the correct way to construct the block if there isn't a default XML String to start with.

jisqyv:
@halatmit I meant to say his code for constructing the block when when getDefaultXMLString doesn't provide one was where the problem was.

halatmit:
thanks



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

Issue #219, opened 2015-01-12
SetFocus for textboxes


halatmit:
Here's a suggestion form Enis for a new useful feature. 

Here's the reference from the SDK:

Added in API level 1

Call this to try to give focus to a specific view or to one of its descendants. A view will not actually take focus if it is not focusable (isFocusable() returns false), or if it is focusable and it is not focusable in touch mode (isFocusableInTouchMode()) while the device is in touch mode. See also focusSearch(int), which is what you call to say that you have focus, and you want your parent to look for the next one. This is equivalent to calling requestFocus(int, Rect) with arguments FOCUS_DOWN and null.
Returns

    Whether this view or one of its descendants actually took focus. 

And here's the actual suggestion for App Inventor

        Would love a SetFocus method for textboxes.

        Say for example, you evaluate the contents of textboxes before submitting to something, and you find one incorrect.  So you pop up a message box that says the address is missing... Ideally, after eh message box closes, you put the user back to the offending address textbox for new entry, but in AI, you're left in a bizarre never-never land.

        And there are times when you want to pop users back to the top of an input screen, but there's no way to do it in AI...

ReachAayush:
Hello @halatmit 

I have started working on this and have come up with a setter and a getter for the Focus property of textboxes. 

This is a screenshot of what the options look like from the textbox methods menu:
![screenshot](https://cloud.githubusercontent.com/assets/2350321/5995170/14dcec14-aa5a-11e4-9457-18e1ea40b214.png)

halatmit:
Looks good.  Thanks!



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.
MIT Dept. of Elec. Eng. and Comp. Sci.
MIT Media Lab, Second Class Faculty

halatmit:
Like the looks of this... This will REALLY help....

ReachAayush:
Added pull request #244

jisqyv:
Pull Request #244 merged.

PR #244, opened 2015-02-04
added focus for textboxes

ReachAayush:
I have added a Focus Property for textboxes, relating to post #219.

The TextBox will have a Focus property which will either be true or false. If the textbox focus property is set to true, the textbox will become active and the user's default keyboard will be opened. 

From the blocks, you may now select this from the TextBox property options:
screenshot

If you choose to set the TextBox.Focus property to true, maybe after a screen.initalize or button.onclick, the textbox will become the active one for the user to type into. 

Ant tests is also successful.

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/TranslationComponentProperty.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/TranslationComponentProperty.java
@@ -115,7 +115,7 @@ public class TranslationComponentProperty {
     map.put("CharacterEncoding", MESSAGES.CharacterEncodingProperties());
     map.put("Checked", MESSAGES.CheckedProperties());
     map.put("ColorChangedEventEnabled", MESSAGES.ColorChangedEventEnabledProperties());
-    map.put("Focus", MESSAGES.FocusProperties());
+    map.put("RequestFocus", MESSAGES.RequestFocusProperties());
     map.put("Country", MESSAGES.CountryProperties());
     map.put("ConsumerKey", MESSAGES.ConsumerKeyProperties());
     map.put("ConsumerSecret", MESSAGES.ConsumerSecretProperties());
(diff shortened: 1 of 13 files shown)


EnisM:
Since I was the person that brought this up...my only comment on this is that the set block doesn't need a true/false... only a true, so it can be a fixed block that says Set TextBox1.Focus.  No need for the socket.  Why would you ever set it to false?  The next control that grabs focus would technically handle that...

kkashi01:
Agreed with Enis. How about just below?

    public void RequestFocus()

ReachAayush:
I agree with you guys, I have changed the fix accordingly. Instead of a setter and a getter, you can now make a call to requestfocus. 

screenshot

ant tests builds successfully.

EnisM:
Looks much better to me... nice!  Thanks...

appinventor/appengine/src/com/google/appinventor/client/TranslationComponentProperty.java
```diff
@@ -115,6 +115,7 @@ public static String getName(String key) {
     map.put("CharacterEncoding", MESSAGES.CharacterEncodingProperties());
     map.put("Checked", MESSAGES.CheckedProperties());
     map.put("ColorChangedEventEnabled", MESSAGES.ColorChangedEventEnabledProperties());
+    map.put("RequestFocus", MESSAGES.RequestFocusProperties());
```
jisqyv: I believe this needs to be moved to TranslationComponentMethods.java now that it is a function instead of a Property

appinventor/components/src/com/google/appinventor/components/runtime/TextBox.java
```diff
@@ -153,6 +154,16 @@ public void HideKeyboard() {
...
   }
 
+    /**
+     * Request focus to current textbox.
+     */
+    @SimpleFunction(
+            description = "Sets textbox active."
+    )
+    public void RequestFocus() {
+        TextViewUtil.requestFocus(view);
```
jisqyv: Why create a new function in TextViewUtil. Why not just do "view.requestFocus()"?

jisqyv:
@ReachAayush See line comments.

ReachAayush:
I have made the aforementioned fixes. I apologize for not having realized that earlier.

jisqyv:
@ReachAayush LGTM. I have squshed it, fixed some minor whitespace (added the documentation) and sent it off to Gerrit prior to merging.

EnisM:
This is a great addition!

jisqyv:
MERGED.

Issue #220, opened 2015-01-13
Screen Switching should send all Yail in one shot


jisqyv:
This should make screen switching both faster and less error prone.

PR #249, opened 2015-02-06
Combine Yail Forms sent to the Companion

jisqyv:
When we have a queue of forms ready to go to the Companion, combine
them into one large Ajax request. Unfortunately we lose track of which
block each form is form. However combining forms is likely only going
to happen when this isn’t as important. Keeping track of blocks
requires a new support in the Companion, which is forthcoming.

Change-Id: I312729fda02155fc0b142e2149fab1bdbadb3921

### DIFF
--- a/appinventor/blocklyeditor/src/replmgr.js
+++ b/appinventor/blocklyeditor/src/replmgr.js
@@ -28,6 +28,8 @@ goog.require('goog.crypt.Sha1');
 goog.require('goog.crypt.Hmac');
 goog.require('goog.crypt.base64');
 
+top.loadAll = true;             // Global for debugging!
+
 // Repl State
 // Repl "state" definitions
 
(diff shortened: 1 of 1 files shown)


jisqyv:
@halatmit @afmckinney This addresses #220. It eliminates the "Ghost Component" that we see when loading a project into the Companion.

jisqyv:
@halatmit @afmckinney OK. Now we are ready for a review...  -Jeff

halatmit:
I tried this and it seems to work.   Is there some specific test I should try?
If not, then LGTM.

jisqyv:
@halatmit There isn't a specific test. I have been running it on largish projects and it appears to work.

I'll arrange to merge it.

jisqyv:
Latest commit is just a squash of the previous commits.

PR #252, opened 2015-02-08
Added a Progress Dialog method to the Notifier component fixes #245

### LINKED ISSUE #245: Really Need a Processing Dialog Message
This probably isn't that difficult to do and it would be VERY helpful, especially in situations when you know what you're going to do in AI Code will be time consumptive...

prog1

Could have two blocks... one to turn it on and add a message:
waitingblock1

And a simple block to hide or turn it off...
waitingblock2

gitars:
So this is my take on the solution to the issue "Really Need a Processing Dialog Message" which is issue #245. In the discussion there, names for the methods/block were suggested as

ProgressShow(String title, String message) //if empty title, then remove title -- always cancellable
ProgressDismiss() // cancels if already on

But I ended up using
ShowProgresDialog(String message, String title)
DismissProgressDialog()

Also, the requester said there is no need for a title, just the message. But as the standard android ProgressDialog takes a message, and as it could be useful to others, I decided to leave it in.

Either way, changes to the block names and what arguments are taken can very easily be changed. Let me know what you all think, and what you would like changed.

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/Notifier.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/Notifier.java
@@ -8,6 +8,7 @@ package com.google.appinventor.components.runtime;
 
 import android.app.Activity;
 import android.app.AlertDialog;
+import android.app.ProgressDialog;
 import android.content.Context;
 import android.content.DialogInterface;
 import android.graphics.Color;
@@ -17,11 +18,11 @@ import android.text.Html;
 import android.text.SpannableString;
 import android.util.Log;
 import android.view.Gravity;
+import android.view.View;
 import android.view.inputmethod.InputMethodManager;
 import android.widget.EditText;
 import android.widget.TextView;
 import android.widget.Toast;
-import android.view.View;
 
 import com.google.appinventor.components.annotations.DesignerComponent;
 import com.google.appinventor.components.annotations.DesignerProperty;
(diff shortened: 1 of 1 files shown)


gitars:
screen shot 2015-02-08 at 4 26 18 pm

screenshot_2015-02-08-16-26-05

kkashi01:
I suggest using ProgressShow and ProgressDismiss. This way, whey user does typeblocking, both methods of notifier will show.

kkashi01:
Also, might be better to swap parameter order like: String title, String message
Usually thinking is from top-down from title then message itself

jisqyv:
@gitars If you want to submit this, you need to edit OdeMessages.java and a bunch of other files when you add Methods/Properties to Components. You can see how it is done in this commit:

https://github.com/mit-cml/appinventor-sources/commit/9cc43be1d7132e23cdb9545ae8a1d59d5dec0472

-Jeff

gitars:
@jisqyv Yea, I realized I was lacking some parts to this, I'll get on it. Thanks for the example.

@kkashi01 The order between message/title seems to be that way the other notifier component methods do it as well, so I decided to stick with it. (But I agree, I would have preferred the normal order). Also, I'll think about the naming suggestion.

halatmit:
Why was this closed and deleted?   Is it not going to be worked on any more?

gitars:
I kind of messed something up with my git repo and I wasn't sure how to fix it, so I deleted the branch and repushed. Probably not the best way to handle it, but I wasn't really sure what to do. I added a post in the App Inventor Open Source Development group to get more feedback. I can't open this pull request again because of the poor way I handled this, but I'll open a new one when I have a more finished feature.

PR #253, opened 2015-02-09
Make sure blocks editor code compiles properly.

jisqyv:
Remove compilation errors from compiling blockly-all.js

Change-Id: I5bb3ddad6d2dca7e97ffe27c391b8cc57fb8ba9b


jisqyv:
@halatmit @afmckinney Note: This fixes errors I caught while compiling blockly-all.js. When we do "development" builds, we don't compile the blockly javascript, so these errors are not caught. Note: I am putting the release up on ai2testing and will include these fixes (I have to, it won't otherwise compile :-) ).

halatmit:
Super.  I never realized this.

Thanks!


== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
@halatmit I'll take that as a LGTM and merge it...

PR #268, opened 2015-02-16
Added a Progress Dialog method to the Notifier component fixes #245

### LINKED ISSUE #245: Really Need a Processing Dialog Message

gitars:
(I closed my original pull request and can't reopen it, but this is a new copy).
@jisqyv Will be helping me out with the internationalization updates and YaVersion.

------------- Stuff from before ---------
So this is my take on the solution to the issue "Really Need a Processing Dialog Message" which is issue #245. In the discussion there, names for the methods/block were suggested as

ProgressShow(String title, String message) //if empty title, then remove title -- always cancellable
ProgressDismiss() // cancels if already on

But I ended up using
ShowProgresDialog(String message, String title)
DismissProgressDialog()

Also, the requester said there is no need for a title, just the message. But as the standard android ProgressDialog takes a message, and as it could be useful to others, I decided to leave it in.

Either way, changes to the block names and what arguments are taken can very easily be changed. Let me know what you all think, and what you would like changed.

screen shot 2015-02-15 at 2 05 39 pm

screenshot_2015-02-08-16-26-05


gitars:
So I merged newer code in to my branch, I didn't realize it would cause this problem here. But I think I finished all the stuff with internationalization and versioning.

gitars:
So it looks like it made this pull request very cluttered because of all the other commits. Is there something I should do about this?

jisqyv:
@gitars Don't worry about it, I'm pretty good at squashing and sorting things out. I'll let you know if I run into a problem.

jisqyv:
@gitars I just added a pull request against your fork which modifies this work. If you accept that pull request it should automatically update this pull request. Once that is done, I can merge it all together. I did it that way so you can see what I had to change.

jisqyv:
@gitars LGTM, MERGED (squashed)

PR #269, opened 2015-02-16
Fix BuildServer aapt failded error when set AppName with an unicode string on Windows OS

wanddy:
Fix BuildServer aapt failded error when set AppName with an unicode string on Windows OS


jisqyv:
@wanddy LGTM, MERGED

PR #271, opened 2015-02-17
Fix for initial scrollable state in Designer #203

### LINKED ISSUE #203: Designer startup glitch on ScreeScrollable
Create a project with a canvas and make it's vertical fill_parent.

You'll see that it doesn't fill in the designer.

To fix this, go to the screen proprties in the design.  Check Screen "scrollable" and then uncheck it and the canvas should fill.    It looks like the startup for the Form(?) assumes the screen is Scrollable even though the designer says that it's not.

josmas:
After having changed the default value of Screen.Scrollable (this was done a few months ago) now we need to force the setting of the property when creating a MockForm. Fixes #203.

Please review @jisqyv @halatmit


jisqyv:
LGTM, Moving to Gerrit

jisqyv:
MERGED

PR #275, opened 2015-02-20
Serve fonts with appropriate content-type and filenames

jisqyv:
Quote the filenames so they properly load on some windows systems.

Change-Id: Ib8afdda5915ef7dc8c9be387a847cffab15da121


jisqyv:
@halatmit @afmckinney This fixes the cases where people say project names and some buttons don't render.

halatmit:
Awesome.  I'm super impressed that you found this!

LGTM

PR #276, opened 2015-02-20
Update activity starter documentation

halatmit:
I played around but i can't find a frameheight to use in line 455 that gets rid of the internal scrollbar.
Let this go as it is, if you don't see how to fix it.


halatmit:
I got rid of the internal scrollbar, but this might not be a longterm solution, for the case where we edit the doc to add a lot on content:  the iframe height might need to be increased.

jisqyv:
@halatmit LGTM. I set scrolling=yes. So if the document height exceeds the 6000 value, a scrollbar will appear. But otherwise it won't be there.

Pushing to Gerrit now.

jisqyv:
MERGED

PR #283, opened 2015-02-24
Added property in Slider component to all for Slider Thumb removal

gitars:
Earlier, I wanted to see the addition of a progress bar, and recently it was brought up to me that the Slider component can easily be modified to be a progress bar. So basically, what I have here is the addition of a property to the Slider component that will remove the Thumb Slider, and you will simply be left with a progress bar. The way it works is, when the property, ThumbEnabled = true, nothing is changed. When it is false, the Thumb Slider is hidden and the ability to change the progress (from the UI) is disabled (you can still change the progress from the code/blocks). 

The reason I went with "ThumbEnabled" was because there is another property "ThumbPosition" so I figured they kind of go together.

Also, this is not a complete feature because it lacks Internationalization which I am still not 100% sure how to do. If anything else is missing, please let me know.

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/common/YaVersion.java
+++ b/appinventor/components/src/com/google/appinventor/components/common/YaVersion.java
@@ -313,8 +313,10 @@ public class YaVersion {
   // For YOUNG_ANDROID_VERSION 119:
   // - TEXTBOX_COMPONENT_VERSION was incremented to 5
   // - WEBVIEWER_COMPONENT_VERSION was incremented to 6
+  // For YOUNG_ANDROID_VERSION 120:
+  // - SLIDER_COMPONENT_VERSION was incremented to 2
 
-  public static final int YOUNG_ANDROID_VERSION = 119;
+  public static final int YOUNG_ANDROID_VERSION = 120;
 
   // ............................... Blocks Language Version Number ...............................
 
--- a/appinventor/components/src/com/google/appinventor/components/common/YaVersion.java
+++ b/appinventor/components/src/com/google/appinventor/components/common/YaVersion.java
@@ -726,7 +726,9 @@ public class YaVersion {
 
   // For SLIDER_COMPONENT_VERSION 1:
   // - Initial version.
-  public static final int SLIDER_COMPONENT_VERSION = 1;
+  // For SLIDER_COMPONENT_VERSION 2:
+  // - Added the property to allow for the removal of the Thumb Slider
+  public static final int SLIDER_COMPONENT_VERSION = 2;
 
   // For SPINNER_COMPONENT_VERSION 1:
   public static final int SPINNER_COMPONENT_VERSION = 1;
--- a/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
@@ -2738,6 +2738,10 @@ public interface OdeMessages extends Messages {
   @Description("")
   String ThumbPositionProperties();
 
+  @DefaultMessage("ThumbEnabled")
+  @Description("")
+  String ThumbEnabled();
+
   @DefaultMessage("Day")
   @Description("")
   String DayProperties();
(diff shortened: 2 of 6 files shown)


marksherman:
I tested locally. Looks great to me! No runtime errors in companion or apk build.
Notes:
- tested thoroughly: the thumb handle is NOT touchable/draggable when invisible
- bringing the thumb handle back to visible instantly makes it interactive again, vice-versa
- The PositionChanged event still fires regardless of handle visibility (as expected)

I have an app that exercises the feature that I'll post to the google group.

appinventor/components/src/com/google/appinventor/components/common/YaVersion.java
```diff
@@ -726,7 +726,9 @@ private YaVersion() {
 
   // For SLIDER_COMPONENT_VERSION 1:
   // - Initial version.
-  public static final int SLIDER_COMPONENT_VERSION = 1;
+  // For SLIDER_COMPONENT_VERSION 2:
+  // - Added the property to allow for the removal of the Thumb Slider
+  public static final int SLIDER_COMPONENT_VERSION = 2;
```
jisqyv: You also need to update YOUNG_ANDROID_VERSION. Search for it in the code, what you need to do is obvious (or should be :-) )

jisqyv:
There is an interesting argument that we should implement a "real" progress bar, aka use the one provided by Android. By using the Android provided version the user will get a same "feel" of the version of Android running on her device. For example with Lollipop, it will be a material design version.

jisqyv:
Perhaps #268 should be updated to provide both a spinner and the progress bar.

gitars:
@jisqyv So, I think this will also provide the same look and feel as the Android Progress Bar, because the Slider component is an Android Seekbar, which is an extension of the Android Progress Bar. I think the argument could still be made for a Progress Bar component, but it would have all the same methods as the Slider component, which might be undesirable.

In regards to #268, it could use a progress bar, but then we would have to add a good number of methods to that progress bar functional, by again adding all the methods in the Slider component.

jisqyv:
@gitars Good point.

appinventor/components/src/com/google/appinventor/components/common/YaVersion.java
gitars: Good note, thanks! I updated it, hopefully that was all.

halatmit:
Make sure the new property is internationalized.




== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

gitars:
Yup, I noted that above. I was hoping that @jisqyv could help me with the Internationalization of #268 and afterwards I would be more capable of doing so for this feature.

halatmit:
See the comment in YaVersion at line 392 ff



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

gitars:
@halatmit Yes, I have seen those docs before, but it seems a little undocumented to me. It says to add a message to OdeMessages, so I did that. Then it says, You then have to add a mapping to TranslationComponentProperties.java (or ..Methods or ..Events, etc). As I based what I added the ThumbPosition method, which is only in TranslationComponentProperty.java and TranslationDesignerProperties.java, I added a mapping for ThumbEnabled in those two files. 

Then, it says "If you edit the description of a component (but not yet a property,method or event of that component) you must also find and update the description in OdeMessages.java." But I did not edit the description of the component. 

That's all it says to do, and I don't know what else I am missing. If you have more information, that would be helpful.

halatmit:
That sounds like you did the right thing. We'll look carefully when we do
the code review.  == Thanks




== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
@gitars @halatmit I'll complete the review (and merge if no issues) tonight. So far it looks good.

jisqyv:
LGTM. MERGED

@gitars I have squashed your changes to one commit, edited the commit message and MERGED the result into master. Note: I made two edits. I edited versioning.js to deal with the component update (a new requirement that isn't well documented) and I also updated the on-line documents. You can review the commit for details.

gitars:
Awesome thanks!

PR #293, opened 2015-03-01
Update Text to Sppech to provide lists of the available languages and countries

halatmit:


halatmit:
@jisqv @afmckinney 
This is a bit unpleasant because the TTS engine needs to be initialized before you can get the lists
of available languages and countries.  I added the initialization check to the property getters.  That seems to provide enough time for the initialization, and I put in a error check, but am not implementing a retry-delay loop.

jisqyv:
MERGED

PR #306, opened 2015-03-05
Updated the Storage documentation to reflect the changes to the Fusion T...

ajcolter:
...ables component that allows Service Authentication. Added the additional necessary steps to make this possible. Also proofread the other entries on the Storage page.

@afmckinney @jisqyv

### DIFF
--- a/appinventor/docs/reference/components/storage.html
+++ b/appinventor/docs/reference/components/storage.html
@@ -463,7 +463,7 @@
 
 <h2 id="File">File</h2>
 
-<p>Non-visible component for storing and retrieving files. Use this component to write or read files on your device. The default behaviour is to write files to the private data directory associated with your App. The Companion is special cased to write files to /sdcard/AppInventor/data to facilitate debugging. If the file path starts with a slash (/), then the file is created relative to /sdcard. For example writing a file to /myFile.txt will write the file in /sdcard/myFile.txt.</p>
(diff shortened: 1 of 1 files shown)


halatmit:
Doc looks good, and can be merged.
@jisqyv 
@afmckinney 

Does DoQuery work at all now?  If not, do we want to use the upgrader to mark  the block bad and put an error message telling people toi use SendQuery?  If so, I'd merge this current doc fix, and file the DoQuery thing as a new issue.

josmas:
It was deprecated in the api so not working, should be deprecated in the component too.

This happened before we moved to AI2, so the chances of finding one of those blocks is rather slim, but still, should be marked as bad.

jisqyv:
We can deprecated. In fact I *did* deprecate when I was testing the bad block code (but I only deprecated it in my testing) so I know that the deprecation code will do the right thing.

halatmit:
sounds good.  Does DoQuery work at all?  If not, we should clearly
deprecate.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
Let me test it, I'm pretty sure it doesn't work.

-Jeff

afmckinney:
DoQuery is obsolete now, and should be deprecated. But, I will double check just to be 100% sure.

Andrew F. McKinney
andrew.f.mckinney@gmail.com <mailto:andrew.f.mckinney@gmail.com>
+1 978-255-3979

jisqyv:
I just checked. You get a 404 Error if you attempt to use it.

jisqyv:
LGTM, Sending to Gerrit...

ajcolter:
I will say that I found a few more errors in the text. But I can just
submit them as another pull request.

jisqyv:
MERGED

PR #308, opened 2015-03-06
Added target for generating guid so it doesn't happen for ant clean

lizlooney:
I think we shouldn't generate a guid during ant clean.
@jisqyv @josmas


jisqyv:
LGTM

PR #313, opened 2015-03-14
Added generic event handlers

gitars:
So I want to add a way to dynamically add new components, and the way I wanted to do that is after the new component is added, the user can interact with them through the generic blocks. However, the generic blocks lack generic events, so I think this covers that. I wanted to make sure this was all good rather than just having one huge pull request with everything else as well.


appinventor/blocklyeditor/src/blocks/components.js
```diff
@@ -61,21 +68,35 @@ Blockly.Blocks.component_event = {
...
         .appendField(this.componentDropDown, "COMPONENT_SELECTOR")
         .appendField('.' + window.parent.BlocklyPanel_getLocalizedEventName(this.getEventTypeObject().name));
-    this.componentDropDown.setValue(this.instanceName);
+      //for non-generic blocks, set the value of the component drop down
+      this.componentDropDown.setValue(this.instanceName);
+    } else {
+      this.appendDummyInput('WHENTITLE')
+        .appendField(Blockly.Msg.LANG_COMPONENT_BLOCK_TITLE_WHEN + this.typeName + '.' + window.parent.BlocklyPanel_getLocalizedEventName(this.getEventTypeObject().name));
+      var compInput = this.appendValueInput("COMPONENT")
+        .setCheck(this.typeName).appendField(Blockly.Msg.LANG_COMPONENT_BLOCK_GENERIC_METHOD_TITLE_FOR_COMPONENT)
```
gitars: I was not sure if I should have created a new message "Blockly.Msg.LANG_COMPONENT_BLOCK_GENERIC_EVENT_TITLE_FOR_COMPONENT" (changing "METHOD" -> "EVENT"), but I was not sure if it was necessary or really exactly how to do it.

gitars:
Also, this is what the generic event block looks like, and the blocks for the sample app that I tested it on:

screen shot 2015-03-13 at 8 47 50 pm

puravidaapps:
this looks great!
but how can I use one of the advanced blocks with the current button
component inside that event?
in the example you are using the regular block

set Button2.Text to

how would I use the following block from the "Any component" drawer instead

set Button.Text
of component
to

Taifun

Am 13.03.2015 um 18:48 schrieb gitars:

gitars:
I'm a little confused on what you mean. You can do something like this:

screen shot 2015-03-13 at 10 14 18 pm

puravidaapps:
yes, sorry, you are right, that would be possible...
But I still do not understand the advantages of the suggested generic block...

For example let's assume there are 4 buttons. After each button click, you like to call a procedure with the button component. Currently you have to define a click event handler for each button like this
unbenannt2


imho it would be nice to have a generic event handler, which gives me automatically the component, which triggered the event as local variable.

unbenannt

Taifun

gitars:
Oh yea, you're right. I noticed this myself and how inconvenient it could be because for each button you will have to initialize a event handler specifically for that button. The reason this has some advantage, is because I want to add a way to dynamically add new components, and the way I wanted to do that is after the new component is added, the user can interact with them through the generic blocks. The one problem is that that the generic blocks lack generic events, so I added this. 

What you stated certainly is also applicable to my end goal, as with this current solution, if you dynamically add buttons, you have know exactly how many you want to add, as for each one, you would have to create a new event handler. But with what you said, you could add any arbitrary number of them.

That being said, I see the value in what you suggest, but I think I might try to focus on dynamically adding new components to the screen and later come back to this issue (but I will start looking into it now).

halatmit:
We eventually want to have both of these features: (generic handlers and adding components dynamically).    Be careful that your design for adding new components does not make it overly difficult to have generic event handlers.

gitars:
@halatmit when you refer to generic handlers, do you mean as @puravidaapps refers to them? I can try to work on something of that form in place of what I have, as I do agree that such a block would be more generalized.

puravidaapps:
so how will the user be able to use a button component of a dynamically
generated button?
where does that button component block come from?
Taifun

Am 14.03.2015 um 10:56 schrieb gitars:
> The reason this has some advantage, is because I want to add a way to
> dynamically add new components

halatmit:
@guitars:  When I refer to generic handlers, I mean the same thing Taifun described.
Also, it would be better if you finished this pull request, rather than starting on dynamic addition of components.   Having pull requests open for a long time ends up creating merge conflicts and more work for reviewers.

gitars:
@halatmit okay sure, I think it is actually a better idea to continue to work on this first anyway. Also, should there be two types of generic handlers, the way I implemented it and the way Taifun described, or do you think it would be better to have generic handlers only the way Taifun described?

halatmit:
I don't see a reason for two types of generic handlers.   Is there any advantage to that?

gitars:
Well maybe in the future when do have dynamically added components. In that case, if we only have the way Taifun described, there would be only one event handler for all of the dynamically added components, and you would have to case on which component triggered that event (if you want to do something different for each of them). So you would get something like this:

screen shot 2015-03-14 at 7

So in such a case, it might cause complicated event handler blocks, where a user might just want a different one for each component.

Overall, I'm not really sure if both are needed, as you can get the job done purely with the way Taifun described, but I thought I might play devil's advocate here. I think I may need to think about it more.

halatmit:
That's a good point.  Maybe we would like both of them.

puravidaapps:
The button variable in your screenshot is a component, isn't it? But you are comparing the button component with a text or is this something else we currently do not have? This brings me back to my previous questions: how will the user be able to use a button component of a dynamically generated button? Where does that button component block come from?

Yes, probably these are 2 different feature requests, but the probably(?) easier task is to a provide generic event handlers as shown in my screenshot?
For the task "create dynamically new objects" you also might want to take a look at the ideas here https://code.google.com/p/app-inventor-for-android/issues/detail?id=197

For the simple 4 button example of course it could be done as I showed... However let's assume we have 20 sprites. Currently we have to define for each sprite a separate CollidedWith event, EdgeReached event, Touched event etc, which results in loads of different event handlers very quickly...

gitars:
That is currently something we don't have. I was just getting at the idea that if only have a generic event handler in the way you described, all dynamically added components of the same type would have to share event handlers. 

The way I was thinking to add components dynamically, is a have a method for the screen, and any container, that takes in an existing component, creates a new one of the same type and returns the new component. The reason why I think this is reasonable (to take in an existing component as an argument), is because suppose you don't have any buttons on the screen and you somehow dynamically add one. Then you won't have the generic blocks for buttons to use to interact with the new dynamically added button. Also, as I want to add a remove component method as well, you can remove the existing component if you really don't want it on the screen.

Yes I agree, they are two different pull requests. That's why I started on for generic event handlers here. I will continue work on this one by trying to add the generic event handlers you described.

gitars:
I was wondering if someone could help me out with how the Yail part of the project works. I know that the blocks are all converted to Yail, and somehow the events are registered from in runtime.scm / EventDispatcher.java, but I'm not quite sure how that happens. I'm also not quite sure how the events are actually dispatched. Is there any documentation on this? Having these new generic events will require some addition/change to this code (as the events do not have a specific component tied to them as others do), and I really need a better understanding of this to make progress.

gitars:
Also, currently I have blocks that look like this

screen shot 2015-03-25 at 6 09 17 pm

The left block which was developed earlier (I call this the generic event block) works. The right block (which I call the multi generic event block because it handles events for multiple components), is not actually functional yet.

mark-friedman:
I can probably help you out with the YAIL and Java event dispatching code.  Unfortunately, there's no general documentation on it and the code comments are pretty sparse.  Mea culpa ;-(  I'll try and re-familiarize myself with the code and get back to you with an overview of the process.  In the meantime, if you have any specific questions I can try and answer them.

gitars:
So, I guess my general question is, where/when are events registered with that function in runtime.scm (https://github.com/mit-cml/appinventor-sources/blob/master/appinventor/buildserver/src/com/google/appinventor/buildserver/resources/runtime.scm#L447) that eventually registers the event with "registerEventForDelegation" in EventDispatcher?

gitars:
Actually, I think I figured that out. Right now I'm kind of confused on what happens in runtime.scm when the the apk is being built and how an event is actually dispatched in the actual app.  Pretty much what is happening here https://github.com/mit-cml/appinventor-sources/blob/master/appinventor/buildserver/src/com/google/appinventor/buildserver/resources/runtime.scm#L399

mark-friedman:
Thanks.  I'm hoping to get around to this in the next few of days.  Feel free to bug me if I don't ;-)

halatmit:
Thanks.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

gitars:
Hey I was wondering if you got a chance to look at this. I think I could really use an overview of the build process in general. (Thanks by the way!)

gitars:
So I started thinking about ways to implement this feature without having to edit runtime.scm and trying to conform to the normal event registration/dispatch flow (so easier to implement...maybe). And I think it is possible (maybe not 100% clean, but doable), and I was wondering if you think it would be something that is acceptable.

Basically, we still have the "multi-generic" event handler that returns a the component when it is called, but we register it as a normal event to one of the components on the screen, under a different name such as "MultiGeneric-" + EventName. Then, when a normal event occurs, we check all the components to see if any of them have a event closure for "MultiGeneric-" + EventName. Then, we simply add the component that triggered the event to the list of arguments and dispatch it. 

I think all of this done a lot easier, but like I mentioned before, it might not be the cleanest solution. However, I do believe that it would be easier to both implement, understand, and maintain. Let me know on your thoughts.

Currently I am blocked on the way I was working on it before, so I will start exploring this solution.

mark-friedman:
I should be able take look at this today.

-Mark

halatmit:
Thanks.   I don't know how hard this is, but I'm sure he'll need help.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

gitars:
So I worked on that other idea I had. Check it out if you get the chance. I didn't test very much at all, but it seems to work for buttons.

https://github.com/mit-cml/appinventor-sources/pull/347

halatmit:
This is obsoleted by PR #450.  closing

Issue #311, opened 2015-03-10
Bug in lookup-in-pairs


halatmit:
See

https://groups.google.com/forum/#!msg/mitappinventortest/2hEzhiJEfkQ/fbTak248bZAJ

The problem is that a string "123" is not recognized as equal to a number 123.   Because Hal programmed it to use Kawa equal? to compare keys.  Need to account for number-string conversion.

halatmit:
#315

PR #315, opened 2015-03-15
Fix lookup in pairs so that numbers in keys will match string forms of those numbers

halatmit:


halatmit:
Issue #311

jisqyv:
LGTM, moving to gerrit

PR #332, opened 2015-03-23
fix for duplicate Warning for disabled blocks

jbensal:
Currently, all event handler blocks are marked as duplicate blocks (and show a warning) if there are two or more blocks of the same type. This happens even if one or more of the blocks are disabled. e.g. if you have one disabled block and one enabled block of the same type, you won't be able to build your application.

This fix separates the duplicate checking for disabled and enabled blocks of the same type. Now, errors are only shown for enabled blocks if there are 1 or more enabled blocks of the same type. Multiple disabled blocks of the same type will show a warning, but will not have an impact building the application.

 
screenshot 2015-03-23 14 56 13


halatmit:
LGTM

jisqyv:
@jbensal MERGED

PR #333, opened 2015-03-23
Do not attempt to rename the “type” of a block.

jisqyv:
Do not attempt to rename the type of a block when it’s name is
changed. Doing so can corrupt projects is the oldname of a block is a
substring of a type name (say component_set_get). If you create a
component named “et”. Make sure it has blocks in the workspace and
then rename it to “st” the project will become corrupted because
occurances of the string “component_set_get” will turn into
“component_sst_get.”

Change-Id: I2dbcb308a6f81f67a2ebdef645e70c4dfee93f7a

### DIFF
--- a/appinventor/blocklyeditor/src/blocks/components.js
+++ b/appinventor/blocklyeditor/src/blocks/components.js
@@ -340,9 +340,6 @@ Blockly.Blocks.component_method = {
       //var title = this.inputList[0].titleRow[0];
       //title.setText('call ' + this.instanceName + '.' + this.methodType.name);
       this.componentDropDown.setValue(this.instanceName);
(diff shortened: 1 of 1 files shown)


jisqyv:
@halatmit This fixes the corruption...

jisqyv:
@halatmit Note: I'm going to push up another commit that completely removes the commented out code (and the snide comment...)

jisqyv:
@halatmit @afmckinney One of the early XML representation of components had the component name as part of its type. At some point Paul rewrote how components were represented and the names are no longer part of the component type. The code I removed was a vestige of the previous code and is no longer needed, or wanted!

-Jeff

halatmit:
LGTM
Great catch in figuring this out.

PR #335, opened 2015-03-24
Deprecate useFront camera property

### LINKED ISSUE #180: useFront property of camera no longer works
This property relied on an undocumented feature of the android camera system, so there was a well-known risk that it may stop working someday. That day came faster than anticipated.

It appears that the feature has been removed from nearly all androids, including older ones. My guess is a google play services update was pushed out that removed it, even to old 2.3 devices.

Leaving the useFront feature may cause more confusion than anything else. In current testing I, and some others on the forums, can't find a single remaining device where it works. 

What I do not know is the right way to remove it. There exist projects that use it, is there a mechanism to check projects on load to nudge the components to updated versions? I believe there is, I just don't know enough about it to complete this recommendation alone.

marksherman:
Deprecates the code relating to the property and removes it from the designer.

Fixes #180 (at least in the short term, by taking the feature away)

@josmas @jisqyv


josmas:
Please hold on doing anything with this at least until applications for GSoC15 are over. One of the challenges involves this property and people can get confused if it disappears right now. cc/ @jisqyv @halatmit

osmidy:
I was looking into #180 yesterday.  Apparently there exists no intent to specify which camera to use (other than the unofficial) non-working one; the user should decide what camera to use.  Unless we want to deal directly with android.hardware.camera, dropping the property is best. For what it's worth, Camcorder currently lacks the property.  @halatmit @josmas

halatmit:
How does the android camera app let the user pick which camera to use?



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

osmidy:
Whatever camera app is used by default on the phone should have a button to switch between the front- and back-facing cameras.

marksherman:
The system camera app, like most camera-using apps, accesses the camera
system directly, through which it can query available cameras and select
which to use. (I've looked at this api)

App Inventor has always used an intent to launch the external camera app,
and that intent does not have any way of specifying, and it definitely
cannot query device state.

Building in direct camera access into app inventor (and no longer launch an
external app) is the inevitable solution. osmidy's assessment is correct-
until we're ready to do that, we should deprecate the feature.

I have the deprecation ready to pull, and am waiting for GSoC15 to conclude
before submitting (as instructed by José).

halatmit:
If we could implement the chooser button, then couldn't we implement that
same chooser as a property of the camera component?     Or am I just
confused?



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

marksherman:
If we rewrite the camera component completely with the camera API, we will
be able to have lots of good stuff, including a switch button. Also, we'd
be able to query the cameras available from the blocks, and choose cameras
from the blocks.

But we don't use the camera API. we simply open the default camera on the
device, and that app (which varies device to device) typically has a switch
button. But we have no control over it.

A less-critical problem with our current solution is we can't write
consistent instructions on what to do when the camera is open, and our
users can't predict the experience for their users. The camera apps
interfaces vary. Samsung's camera, beyond the big button to take the
picture, looks nothing like the nexus camera, etc.

halatmit:
also see issue #338

marksherman:
@josmas is GSoC15 sufficiently over yet? This trigger awaits to be pulled.

thequixotic:
Rewriting the Camera Component is tricky because of the vast types of devices out there. There is a standard way of doing things neatly that should work on all devices, but it requires a SurfaceView  (http://developer.android.com/reference/android/view/SurfaceView.html) with a size of at least 1x1 px . SurfaceView can be achieved by upgrading the Canvas Component.

However if you ignore older devices (<API 11) we can use SurfaceTexture (http://developer.android.com/reference/android/graphics/SurfaceTexture.html) to use camera and take pictures without a preview.

Again the Camera api has so many settings which are different for different camera sensors. So images taken on some devices with improper settings can give you too dark images (due to very low exposure setting). To fix this there seems to something called AutoFocus(http://developer.android.com/reference/android/hardware/Camera.html#autoFocus(android.hardware.Camera.AutoFocusCallback)). 
Maybe we can get default camera preferences. 

Not to mention the deprecated status of android.hardware.Camera.. The newer one is android.hardware.camera2 (android.hardware.http://developer.android.com/reference/android/hardware/camera2/package-summary.html). This might not be a problem though. 

Obviously depracating useFront is easy, but I think we should wait till the Canvas Component can be upgraded to show live Camera preview. As of right now it doesnt crash the app if UseFront is checked, it simply ignores because the Camera App doesnt even know the extra intent data being passed) .. 

One camera App that does accept the UseFront intent data is Camera KK (https://play.google.com/store/apps/details?id=com.moblynx.camerakk&hl=en) .. So UseFront does work with this Camera App linked.

josmas:
@marksherman yes, GSoC applications are over (results are not out yet, though) so taking the tag off now.

halatmit:
LGTM

halatmit:
@afmckinney @jisqyv 
This should be ready to merge.
PTAL

jisqyv:
@marksherman Because you are deprecating a block, you have to update YOUNG_ANDROID_VERSION and the version of the Camerca Component. Make sure the upgrader does the right thing with projects that have the block in use AND have set the useFront property in the designer.

The code as written causes projects which set useFront in the designer to fail to load.

marksherman:
You know what's funny? The existing useFront currently works perfectly on my OnePlus One (Cyanogen 12s).

halatmit:
I'll do this as soon as we fix the deprecation code.

jisqyv:
Implemented in https://github.com/mit-cml/appinventor-sources/commit/27a58237bfb42446c4926964538afef3e1ff042b Closing this one.

PR #341, opened 2015-03-28
Fix for typeblocking in safari

josmas:
Typeblocks in Safari does not work

Safari returns a 'Node' as opposed to an 'Element' so .childNodes should be used instead of .children


jisqyv:
@josmas Have you tested this on Safari (I know, dumb question, but I have to ask :-) )?

josmas:
@jisqyv yes, was the only way to reproduce it and find a fix. tested on firefox, chrome, and safari on mac.

jisqyv:
@josmas Thanks. There is a good chance we will be doing another release this afternoon (to fix a problem with decimals see #343). This will get in that release.

jisqyv:
LGTM

josmas:
great  stuff. thanks!

jisqyv:
@josmas MERGED

PR #343, opened 2015-03-28
Yail list number to string

halatmit:


halatmit:
@afmckinney @jisqyv 
To test this, do some numeric opertations with division and check the the results appear as decimal values (not fractions) with do-it, and with putting the values in list of elements for lsitView and listPicker.
There should be no ".0" after in the case of integer values , e.g., when you divide 10 by 5.
Also set the phone to a different locale, e.g., Italy, and verify that decimals still print with periods, not commas.

jisqyv:
Is this fix important enough for us to do a new release? It would require a companion update, which is a pain for our users. But on the other hand we can make the companion update optional. 

-Jeff

halatmit:
The biggest difference will be to people whose phones are set to a european
locale.   With the way it is now, those phones will display decimals using
comma as the decimal separator; on the other had, when you input a decimal,
you have to use period as the separator.   So it's inconsistent.

The change I made freezes the separator at period regardless of the locale,
which is consistent, but not friendly to european locales.

I can't decide about another release just for this.  The person who cares
about this is Ghica.  Of course, she'd like us to fix the separator for
both reading and writing, but I don't want to tackle that now without some
hints about how our users would be able to write internationalized apps.

As I'm writing this, I"m thinking that yes, we ought to do the release,
since the current situation will be very confusing.  And sadly, with the
comapnion, too.  What do you think?



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
You change also fixes the listview displaying 2.0 when the input was 2 (I'm
testing now, but that appears to be what you are doing).

I believe we should do the release, we can make the Companion update
optional. People who have it set to auto-update from the Store will get the
new version. It actually isn't that bad because we are in what I call the
Companion "soak" period. Right now people who side-loaded the Companion or
are running the emulator are not being asked to auto-update their Companion
from Friday's release. So we can update the Companion again. Later this
week (say Tuesday or so) we'll turn on auto-updating. So people who
side-load will only see one Companion update for both releases.

-Jeff

On Sun, Mar 29, 2015 at 9:02 AM, hal <notifications@github.com> wrote:

> The biggest difference will be to people whose phones are set to a european
> locale. With the way it is now, those phones will display decimals using
> comma as the decimal separator; on the other had, when you input a decimal,
> you have to use period as the separator. So it's inconsistent.
>
> The change I made freezes the separator at period regardless of the locale,
> which is consistent, but not friendly to european locales.
>
> I can't decide about another release just for this. The person who cares
> about this is Ghica. Of course, she'd like us to fix the separator for
> both reading and writing, but I don't want to tackle that now without some
> hints about how our users would be able to write internationalized apps.
>
> As I'm writing this, I"m thinking that yes, we ought to do the release,
> since the current situation will be very confusing. And sadly, with the
> comapnion, too. What do you think?
>
>
>
> == Hal Abelson
> hal@mit.edu
> Prof. of Comp. Sci. and Eng.
> MIT Dept. of Elec. Eng. and Comp. Sci.
> MIT Media Lab, Second Class Faculty
>
>
> On Sun, Mar 29, 2015 at 8:20 AM, Jeffrey Schiller <
> notifications@github.com>
> wrote:
>
> > Is this fix important enough for us to do a new release? It would require
> > a companion update, which is a pain for our users. But on the other hand
> we
> > can make the companion update optional.
> >
> > -Jeff
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <
> https://github.com/mit-cml/appinventor-sources/pull/343#issuecomment-87404200
> >
>
> > .
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mit-cml/appinventor-sources/pull/343#issuecomment-87410734>
> .
>



-- 
_______________________________________________________________________
Jeffrey I. Schiller
Information Systems and Technology – MIT App Inventor
Massachusetts Institute of Technology
77 Massachusetts Avenue  Room E17-110A, 32-392
Cambridge, MA 02139-4307
617.910.0259 - Voice
jis@mit.edu
http://jis.qyv.name
http://appinventor.mit.edu
_______________________________________________________________________

jisqyv:
LGTM

halatmit:
Yes, the change should cause integer values to print without the decimal
point.  People who really want to produce things like "5.0" should use
format-as-decimal.

I agree with doing the release.   Your commend about companion updating
makes me realize it's a little complicated.   There should be a short note
explaining this for developers.  I'm not sure where that not should go,
though.   What do you think?



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
MERGED

PR #347, opened 2015-04-03
Added multi/generic event handlers

gitars:
Stemming off of stuff from here:
https://github.com/mit-cml/appinventor-sources/pull/313

This is a solution to this problem that works like this (copied from other pull request thread):
Basically, we still have the "multi-generic" event handler that returns a the component when it is called, but we register it as a normal event to one of the components on the screen, under a different name such as "MultiGeneric-" + EventName. Then, when a normal event occurs, we check all the components to see if any of them have a event closure for "MultiGeneric-" + EventName. Then, we simply add the component that triggered the event to the list of arguments and dispatch it.

Example. A screen with three buttons, and this block. All of the blocks will have their text changed on click.

screen shot 2015-04-02 at 8 10 31 pm


appinventor/buildserver/src/com/google/appinventor/buildserver/resources/runtime.scm
```diff
@@ -407,7 +408,8 @@
            ;; for event handlers and component names.
            (let ((registeredObject (string->symbol registeredComponentName)))
                  (if (is-bound-in-form-environment registeredObject)
-                     (if (eq? (lookup-in-form-environment registeredObject) componentObject)
+                     ;; temporarily deleted because I have no ideas how to get true out of this condition using isMulti
+                     ;;(if (or (eq? (lookup-in-form-environment registeredObject) componentObject) (isMulti))
```
gitars: So I have no idea how to write code here, everything I tried for an hour failed. Basically, I know we do this check for a reason, but because of the way I implemented multi-generic events, I want to ignore it if it such an event. So I passed in a parameter isMulti, and if it is true, I want to pass this condition. Every way I've tried to do this (by using different types for isMulti such as java.lang.Boolean/java.lang.String and comparing it to thing failed).

However, the multi-generic event handler seems to work here if I simply ignore the condition, so if we get the condition right, it should still work.

halatmit: Check with Mark Friedman.   I don't understand the implications of what you
are doing here.


== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

mark-friedman:
I haven't looked at your code yet but I've starting writing a doc on the App Inventor event handling.  So far, I've written up the part that's in runtime.scm.  Feel free to comment in it.  The doc is at https://docs.google.com/document/d/1Mpg0q5_MtmO6BtIMJjYkfDjiVXehboavx99FVfU91aU/edit?usp=sharing

mark-friedman:
I just added some more to the event handler doc.  In particular I added more on the rationale behind the code that you quote above.

mark-friedman:
I have some questions about the design and implementation of the 'generic' event block (i.e. the one that takes a component as an argument).  

On the design side, it seems to me that this block needs to be able to be placed within a block stack (i.e. have setNextStatement(true) and setPreviousStatement(true)) rather than being a top-level block like ordinary event handlers.  I say this because generally the  component that's being plugged in will determined dynamically during program execution.  I might also changing the block so that it says something like "Set Button.Click for component" to make the distinction between this and the "normal" event handler a little clearer.

I'm also going to mention a couple of things to make sure are handled correctly by your implementation.

* Make sure that you can only have one block handling a given event for a given component.  This will mean, for example, that you probably want to make sure that you can't have both a generic and non-generic event block for a given component type.  I.e. you can't have both Button1.Click and Button.Click.

* Make sure that live component renaming works correctly.  The code in runtime.scm that you quote above is part of what deals with that for the current event handling so it may need to be changed/extended.

gitars:
- On the design side, it seems to me that this block needs to be able to be placed within a block stack (i.e. have setNextStatement(true) and setPreviousStatement(true))

Yea, you are right, I didn't realize that earlier.

- I might also changing the block so that it says something like "Set Button.Click for component"

Sure, that makes sense, especially given that it will no longer be a top level block.

- Make sure that you can only have one block handling a given event for a given component.

So right off the bat, I this becomes an issue:

screen shot 2015-04-04 at 11 55 56 am

As the top block is an event handler for every button, and the bottom block is an event handler for button 1, meaning we will have 2 events handlers for button 1. I don't really know what to do about. It seems unreasonable to say if you have anything like the top block, you cannot have any other event handlers for same event. What do you think?

- Make sure that live component renaming works correctly. 

Okay good point. Can I ask how live component renaming happens? I understand the checks that are done as a result, but I don't understand how they happen.

puravidaapps:
why not? this sounds reasonable for me
you always can add some if statements into the generic event handler
like this

if component = Button1
then set Button1.Text = "do something"
else set Button.Text of component get component
                                                     to "Changed the text"

Taifun

Am 04.04.2015 um 09:57 schrieb gitars:
> It seems unreasonable to say if you have anything like the top block,
> you cannot have any other event handlers for same event. What do you
> think?

gitars:
Yea you right that it can be done. I was just wondering on what everyone thought. Same concern/question I had earlier:

https://github.com/mit-cml/appinventor-sources/pull/313#issuecomment-80759805

I'd be fine either way, but I just wanted to talk about it.

gitars:
@mark-friedman @halatmit 

What do you think about it. Would it be reasonable to not allow the multi-generic event block (the one that returns the component) and any normal event handler blocks for same event?

In my opinion, I think it would make sense to simply have a warning rather than an error. This is because both events can be dispatched, and we should just make sure that the user knows that it will happen. (And the error for two normal event handlers of the same event/component should still exist).

mark-friedman:
It would be inconsistent with the current event handling to allow more than one handler for an event.  I believe that a key reason for the current behavior is that the order of the multiple handler execution is indeterminate and therefore confusing and difficult to debug.

mark-friedman:
@gitars, you asked above about how live renaming works.  I'm not sure that you need to know all that.  I think that you just need to read through the event handling doc I wrote and make sure that the renaming scenario described there (with respect to event handlers) is dealt with for the new blocks you are adding.

That said, I just took a quick look at the current rename code and it's not so clear to me whether the old event handler issue with component renaming still exists.  It looks like maybe component renaming now causes a complete reload on the device, which would, I think, obsolete that issue.  Maybe someone else who knows the current situation with renaming could pipe in and clarify.

halatmit:
@mark-friedman @jisqyv @gitars 
I believe Jeff made the change that Mark is referring to, to force reload on renaming.  Jeff?

gitars:
Okay, so I will work on making sure that if you have a "multi-generic" event block for an event, you can have no other event handlers for the same event for any component of the same type.

gitars:
Also, I agree, I don't think that I really need to know much about live renaming, but thanks for the info.

jisqyv:
Any change in the designer, including component renaming, forces a reload in the Companion. I would like to change that... and its on the list.

-Jeff

mark-friedman:
The restriction on multiple event handlers for a given component should apply to "generic" event blocks as well as "multi-generic" blocks.  Note that this means that you can't have a "generic" event block coexistent with any other event blocks of that event type for that component type.  That's unfortunate, but I don't see good way around that.

halatmit:
That sounds right.  I agree it's unfortunate, but probably necessary.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

gitars:
Okay sure. I think this is all very doable, and I hopefully I can have something done soon.

gitars:
So suppose there is a multi-generic event handler for all buttons, and also Button1.Click event handler. Should both of them getting the warning, or just the multi-generic event handler? Also, should the warning be the same duplicate event warning, or should it be something new, to explain how the concept of the multi-generic block works?

halatmit:
I'm not sure which one should get the warning.  Probably teat it the same
as the case of two ordinary event blocks for the same event and the same
component.

I'd suggest making this a different error message.  For now, these can have
the same text, but we might want to change that.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

mark-friedman:
I agree with Hal.  And note that it is currently error, not just a warning.

gitars:
It seems to me that the name of each component is simply the component's class name. I just wanted to make sure that this always the case, right?

appinventor/buildserver/src/com/google/appinventor/buildserver/resources/runtime.scm
gitars: @mark-friedman 
So previously the condition here is
(if (eq? (lookup-in-form-environment registeredObject) componentObject)

I want this condition to be something like this:
(if (or (eq? (lookup-in-form-environment registeredObject) componentObject) (isMulti))

where is isMulti is a boolean passed into the function. However, something about this is not working correctly, and I haven't been able to figure it out. I was wondering if you have an idea of what the issue is.

Also, the reason why this will be okay is because multi-generic events will be registered to some arbitrary instance of the right type. So when the event is dispatched, the registered object and the one that we pass in won't always match. However, it doesn't matter, because the two objects are of the same type (will be guaranteed by the event name, which will be of the form MultiGeneric-TypeName-EventName). Since, the multi-block handles events for all components of that that type, that is enough to know.

halatmit:
I believe that's the case, but I'm not positive.   You'll have to check the
source.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

mark-friedman:
The name of a component is the name as defined by the user, e.g. 'Button1' or 'ClickyBox'.

appinventor/buildserver/src/com/google/appinventor/buildserver/resources/runtime.scm
mark-friedman: On the surface, the line of code you mention does't work because you have parentheses around 'isMulti'.  In Scheme those parentheses cause a procedure call, so '(isMulti)' means call the procedure that is currently bound to isMulti, and since isMulti will actually be bound to a boolean it fails.

I should point out, though, that your rationale for why it's ok to avoid the lookup isn't really correct, since you're not really taking into account the purpose of the check (i.e., the whole renaming issue).  Perhaps rather than registering the handler to an arbitrary instance (which is kinda hacky anyway and what happens when that component gets deleted) you can extend the dispatch mechanism to support registering event handlers to component types in addition to component instances.

gitars: Yea, I knew it was a workaround, but I only attempted this way because I was blocked on other things in runtime.scm. I thought I'd see what you all thought about this, but I guess I will try to go back to the way that I was trying to do things.

gitars:
Sorry, I was referring to the type name of the component. The instance name is the one that can renamed like this, right?

appinventor/buildserver/src/com/google/appinventor/buildserver/resources/runtime.scm
```diff
@@ -434,12 +443,37 @@
                          registeredComponentName eventName)
                        #f))))
 
+       ;; For the HandlesEventDispatching interface
+       (define (dispatchMultiEvent eventName :: java.lang.String
```
gitars: I was wondering if you could help me debug this function. I'm not really sure how to even begin to do this for this file and I'm not really knowledge about this language at all. 

Basically I have multi-events registered separately, and whenever a normal event is called, the correct multi-event will also be dispatched if it exists (currently hardcoded only for buttons right now). Currently, an app using the multi-event block fails when the event is dispatched. If in the EventDispatcher, if I comment out the line "dispatchDelegate.dispatchMultiEvent(multiEventName, newArgs);" then no error occurs.

mark-friedman: Just wanted you to know that I haven't forgotten about this.  I just haven't gotten much time to take a close look.  I have been thinking about approaches that might simplify things for you but I need to think it through.  I do have a clarifying question, though.  When you refer to multi-events here do you mean the same as what you've called multi-generic in other places?

I'll point out that you say that the "multi-event will also be dispatched ...".  I hope that, as discussed, you plan to only allow for a single handler to be dispatched for any given event.

gitars: Yea, sorry for that confusion, I've been using "multi-generic event" and "multi-event" pretty interchangeably.

Also, yea, I created a new Yail "define-multi-event" for those blocks (still in the same function that creates the Yail for normal events and uses "define-event"). And then, I pretty much tried to mimic everything that happens for define-event, for define-multi-event, which doesn't happen an instance associated with it, just a the component type.

halatmit: It would be more consistent to call these generic events rather than
multievents.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

gitars: Actually it seems to me that, that blocks like this are called "generic method blocks" (https://github.com/mit-cml/appinventor-sources/blob/master/appinventor/blocklyeditor/src/drawer.js#L221)

screen shot 2015-04-12 at 12 26 24 pm

Which makes me feel like it makes sense to call these "generic event blocks"

screen shot 2015-04-04 at 2 14 13 pm 2

But since these are inherently different blocks, I call these "multi-generic event blocks" and/or "multi event blocks"

screen shot 2015-04-06 at 7 49 26 pm

I don't think it make sense to call those generic events as well, but another name is certainly possible

halatmit: What's the difference between the second block and the third block?

gitars: The second block is like a normal event handler except you also have to give it the component that you want to use it for. So these two blocks behave the same (hence the error):

screen shot 2015-04-12 at 1 00 31 pm

The third block handles events for all instances of specific component and returns the instance that actually triggered the event.

halatmit: Sorry to be dense, but I still don't get it.  what is the different between the second block and the third block?  That is, the second block says "set button click handler for [component} to ..."   and the third block says
"when there is a button.click on [component] do ..."
What's the difference between these two things?

gitars: Suppose you have 3 buttons. With the third block,

screen shot 2015-04-12 at 1 11 37 pm

when you click any of the 3 buttons, the text of that button will change to "Clicked"

With the second block, a block like this

screen shot 2015-04-12 at 1 13 12 pm

will only handle events for Button1. If Button1 is clicked, it's text changes.

halatmit: See my previous email just now.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

puravidaapps: you said "The second block is like a normal event handler except you also have to give it the component that you want to use it for. So these two blocks behave the same"

let me ask a probably silly question:  when do you use the new block and when do you use the already existing block? or: what exactly is the advantage of the new block? Everything can be done currently without that new block?
new block:
example
current block:
unbenannt


Taifun

gitars: @halatmit I don't think I've gotten an email from you

@puravidaapps I think in general, these generic event blocks are really for later on when we have dynamically added components.

halatmit: @gitars 
There's no email -- I meant my comment above that starts "Oh, I get it".   What are your thoughts in reasponse to "Think about changing the syntax ...."

gitars: Sorry, missed that comment. In regards to syntax, I changed it because of this: https://github.com/mit-cml/appinventor-sources/pull/347#issuecomment-89458942

Also, yes the multi-generic event returns the instance of the component that was clicked as a callback variable.

halatmit:
OK about comment in #347.  Mark is correct, as usual.

appinventor/buildserver/src/com/google/appinventor/buildserver/resources/runtime.scm
mark-friedman: @puravidaapps and @gitars 

It's also useful even without dynamically created components.  For example, the expression in the "for component" socket doesn't have to just be a constant (e.g. "Button1");  it could be a variable, procedure call, etc.

gitars: @mark-friedman I was wondering if you had gotten a chance to look at this code/think about it. My progress is getting severely slowed down, and I'm not quite sure what to do. Also, if you have any tips on how I can debug this code, that would really be useful because right now I know of no way of doing that other than building an app and seeing it works.

appinventor/buildserver/src/com/google/appinventor/buildserver/resources/runtime.scm
```diff
@@ -456,6 +490,17 @@
...
 
+         ;; Register the multi events with the Simple event dispatcher
+         (define (register-multi-events events)
+           (define-alias SimpleEventDispatcher
+             <com.google.appinventor.components.runtime.EventDispatcher>)
+           (for-each (lambda (eventName)
+                       ;; Tell the Simple event dispatcher to delegate dispatching of this event to this class
+                       (SimpleEventDispatcher:registerMultiEventForDelegation
+                        (as com.google.appinventor.components.runtime.HandlesEventDispatching (this))
+                        eventName))
```
mark-friedman: This won't quite work.  What this for-each statement does is execute the lambda procedure over each of the elements of the variable 'events'.  Each of those elements is a pair (called an event-info) comprised of a component name and an event name.  You need to extract the event name from the event-info in order to pass it to registerMultiEventForDelegation.  To do that you use the 'cdr' procedure from Scheme.  If you look at the register-events procedure above you can see how to do that.  Just remove the '(car event-info)' call since registerMultiEventForDelegation only takes an event name as an argument (unlike registerEventForDelegation, which takes a component name and an event name)

appinventor/buildserver/src/com/google/appinventor/buildserver/resources/runtime.scm
mark-friedman: @gitars 

One way of debugging without having to build apps is to write and run unit tests.  That has the added advantage of adding to the general test suite for everyone to use.  Unfortunately, there aren't any existing such tests for the event handling stuff for you to just add to - you would have to create your own.  But if you're experienced with Java and junit you should be able to write tests for the stuff in EventDispatcher.java.

It's harder for the stuff in runtime.scm, of course, since that requires knowledge of Scheme.  There is some stuff in [YailEvalTest.java](https://github.com/gitars/appinventor-sources/blob/master/appinventor/buildserver/tests/com/google/appinventor/buildserver/YailEvalTest.java) for testing Scheme code using junit that might be worthing looking at.

appinventor/buildserver/src/com/google/appinventor/buildserver/resources/runtime.scm
gitars: Actually, I'm pretty sure that is not a problem. This function is called with a different list of events - the multi events:

(register-multi-events multi-events-to-register)

and this is how that event list is created: 
+       (define (add-to-multi-events event-name)
+         (set! multi-events-to-register
+               (cons event-name multi-events-to-register)))

It is in fact just a list of event names.

appinventor/buildserver/src/com/google/appinventor/buildserver/resources/runtime.scm
gitars: Okay sure, I'll take a look into this.

appinventor/buildserver/src/com/google/appinventor/buildserver/resources/runtime.scm
mark-friedman: @gitars

Oops, you're right.  Can you tell me more about what errors, log messages and behavior you're seeing?

gitars: Sorry for the delay. Basically I have multi-events registered separately, and whenever a normal event is called, the correct multi-event will also be dispatched if it exists (currently hardcoded only for buttons right now). Currently, an app using the multi-event block fails when the event is dispatched. When an app using this block is built and you click on a button, the app will crash and it will simply says "Unfortunately APPNAME has stopped." If in the EventDispatcher, if I comment out the line "dispatchDelegate.dispatchMultiEvent(multiEventName, newArgs);" then no error occurs. Similarly, if in the function dispatchMultiEvent in runtime.scm, if I just change it so that it simply returns without doing anything, no error occurs. So this leads to believe that it is somewhere in this snippet of code that everything breaks, but I am not sure what yet.

gitars: I think I just solved this issue. It was actually a relatively small bug somewhere else. I will push some changes soon.

gitars:
So I think the multi-generic event blocks are good, so I would ask you guys to test it out and let me know what you think / if you find issues.

I did notice that the normal generic event handlers (the ones that a component) do not work any more. The reason for it was because I simply set setNextStatement(true) and setPreviousStatement(true)) for that block. When it was a top level block it worked, but I also need to look into it now that it is not a top level block.

Also, I'm not sure on the current state of the warning/errors for duplicate events, I have not yet worked/tested all that out.

halatmit:
It sounds like you are saying that the reason the normal generic event
handler don't work is simply the block shapes.  If that's the case, then
you can fix this simply by creating to different blocks.  Why don't you do
that so we have something more complete to test?   Also include a demo-app
or two that we can use for testing so we can be sure we're using this
correctly.


== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

gitars:
So what I am actually is saying is that, from what I understand, is that it was assumed that any event definition would be a top level block when the code was written. This means, it will go through and register events that are on the top level. As a result, having the normal generic event block no longer be a top level block, when we want to register this block (whether it be in another event block or in a procedure), there is nothing defined to do the registering.

gitars:
Actually, the more that I look at it, the less sure I am. So there is a define-syntax rule for taking the Yail for events and basically validating/registering it (define-syntax define-event). So this is the Yail

(define-event Button1 Click()(set-this-form)
    (define-generic-event Button2 Click()(set-this-form)
      (set-and-coerce-property! 'Button2 'Text "Clicked" 'text)))

for this:

screen shot 2015-04-20 at 11 03 41 pm

And this does not work as you would expect. If you click the first button, and then the second, nothing really happens to the second one. You would expect, based on these blocks, that after the first button is clicked, clicking on the second will change it's text to "Clicked" but it does not. I'm not really sure on what the reason for this is. 

@halatmit  You mentioned an idea earlier, but I wasn't quite sure what you meant. Could you elaborate?

halatmit:
Sorry, I don't remember what Idea I mentioned.

And the behavior you describe with the two buttons means there's still a
bug, so this isn't ready for other people to test
Right?


== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

gitars:
Well I would say that the multi-generic block is pretty big in and of itself to test, and it is ready. However, if you want to wait for everything, you could do that too.

gitars:
Hey, so I've been working on the making the normal generic block work as a non-top level block, and I realized something that you all may consider to be an issue. Consider something that looks like this:

screen shot 2015-04-27 at 2 33 24 pm

Then, every time Button1 is clicked, the event handler for Button2 can be changed to something else. Now this is something is hard to check for, so I don't know if we can find all ways that people can misuse the block and issue warnings in exactly the right cases. What do you all think?

mark-friedman:
That would be working as intended, from my point of view.  The functionality for 'generics' is fairly advanced.  That's why it's semi-hidden within the "Any" category.

By the way, if you have a test server up and running somewhere it would be great to send us a link so we can at least take a look at the multi-generic block, which I think you've said you have working.

gitars:
Okay, I guess that is true. Yea I think you are right, that is a fairly advanced technique that I will have to think about more of how to implement. For it to work, we then have to unregister the old event so we so when the event is dispatched, the right one gets dispatched.

Also, sure I will spin up a test sever soon.

mark-friedman:
Note that there is an [example of unregistering](https://github.com/mit-cml/appinventor-sources/blob/master/appinventor/buildserver/src/com/google/appinventor/buildserver/resources/runtime.scm#L432)  in dispatch-event in runtime.scm, although it might just work to unregister it within [registerEventForDelegation](https://github.com/mit-cml/appinventor-sources/blob/master/appinventor/components/src/com/google/appinventor/components/runtime/EventDispatcher.java#L109) in EventDispatcher.java.

gitars:
@mark-friedman Yea, thats a good idea.

gitars:
Hey, so I'm running app inventor on apppsot here: http://rare-citadel-93023.appspot.com/
But, I'm not quite sure how to get buildserver to work. Do you know if there is any documentation on this?

halatmit:
See

How to build App Inventor from the MIT sources
<https://docs.google.com/document/d/1Xc9yt02x3BRoq5m1PJHBr81OOv69rEBy8LVG_84j9jc/edit>


== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

gitars:
I decided to run the buildserver locally and pointed the test server to ip address. It seems to work for me. You can try it out, and I will try to leave the build server running as often as I can.

appinventor/buildserver/src/com/google/appinventor/buildserver/resources/runtime.scm
gitars: @mark-friedman Hey have you gotten a chance to look this over? I'll be working on it for a good amount of time later today and would really appreciate some feedback/help. Thanks!

appinventor/buildserver/src/com/google/appinventor/buildserver/resources/runtime.scm
```diff
@@ -636,6 +681,31 @@
...
                (add-to-events 'component-name 'event-name)))))))
 
+(define (define-generic-event component-name event-name args . body)
+  (let ((event-func-name (gen-event-name #`component-name #`event-name)))
+    (define-event-helper event-func-name args body)
+    (com.google.appinventor.components.runtime.EventDispatcher:registerEventForDelegation
+                (as com.google.appinventor.components.runtime.HandlesEventDispatching *this-form*)
+                'component-name
+                'event-name)))
+
```
gitars: Sorry, I meant here.

@mark-friedman Hey have you gotten a chance to look this over? I'll be working on it for a good amount of time later today and would really appreciate some feedback/help. Thanks!

puravidaapps:
to be able to tests, can you provide a companion app, which could be
used together with your test server?
Taifun

Am 28.04.2015 um 18:15 schrieb gitars:

puravidaapps:
to be able to test, can you provide a companion app, which could be used
together with your test server?
Taifun

Am 28.04.2015 um 18:15 schrieb gitars:

gitars:
Sorry, yea, I can try to get to that soon. I've been busy trying to figure out how to get the normal generic block to work.

gitars:
I think I'm done with everything. I'll get you guys a way to test this.

gitars:
My aws free instance doesn't have enough memory to build apps, so I guess I will go with the companion app method. Does any one have any documentation on how I can get this for you to test with?

puravidaapps:
see here
https://groups.google.com/d/msg/app-inventor-open-source-dev/yXYAU95VVBM/0zZjWSelwnYJ

    The look in section 2.5  of the document link below on how to build
    the companion app. There is also a link for a video Jose made in
    same section.

    https://docs.google.com/document/pub?id=1Xc9yt02x3BRoq5m1PJHBr81OOv69rEBy8LVG_84j9jc
    <https://docs.google.com/document/pub?id=1Xc9yt02x3BRoq5m1PJHBr81OOv69rEBy8LVG_84j9jc>



Am 01.05.2015 um 11:38 schrieb gitars:

halatmit:
Wen you build app inventor with ant, it automatically builds a companion
APK that you can download and then distribute to people who want to play
with your code.

Look in

/appinventor/build/buildserver


== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

gitars:
https://www.dropbox.com/s/9lkaabw220fg799/MIT%20AI2%20Companion.apk?dl=0
http://generic-events.appspot.com/

puravidaapps:
I tested a little bit...
the companion app says, that this is Version 2.24?
can you update the version number next time to avoid confusion?

the "when Button.Click (component)" block works fine, great!
the "set Button.Click for component" somehow does not work? Also there
is no Do it?



Taifun


Am 01.05.2015 um 12:02 schrieb gitars:

gitars:
@puravidaapps Sorry, I don't really know much about the companion app, how do I change the version. Also, I noticed that myself using the companion app. Whenever, I reconnected, the app and generic blocks seemed to work fine.

halatmit:
Don't worry about changing the version.  Jeff handles that when we do
distributions.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

gitars:
If you guys get a chance, it would be great if you could test this soon. My term ended yesterday, and I really want to try to get this merged as soon as possible.

halatmit:
Is this ready to test?  I thought the normal generic block wasn't working yet.
Is your plan to release a system without that?

gitars:
As far as I can tell it is working.

-Yash

gitars:
Again, here is the info:

Companion App:
https://www.dropbox.com/s/9lkaabw220fg799/MIT%20AI2%20Companion.apk?dl=0

Site:
http://generic-events.appspot.com/

Let me know if you have any comments.

halatmit:
Please also post a simple AIA and associated APK that illustates the features we should be testing.

gitars:
Here are some tests: 

https://www.dropbox.com/sh/3bimsqatvnyj7vc/AAClV3Ta0-8vRPXXs0qkzJNFa?dl=0

halatmit:
This is looking nice, but there are still bugs.  Here's an example:

Create an app with 5 buttons and the following blocks.  Start the app and touch button 5.  Then touch button1 and the app will crash:

screen shot 2015-05-02 at 1 49 44 pm

Here's what is in logcat:

D/AndroidRuntime(26777): Shutting down VM
E/AndroidRuntime(26777): FATAL EXCEPTION: main
E/AndroidRuntime(26777): Process: edu.mit.appinventor.aicompanion3, PID: 26777
E/AndroidRuntime(26777): java.util.ConcurrentModificationException
E/AndroidRuntime(26777): 	at java.util.HashMap$HashIterator.nextEntry(HashMap.java:787)
E/AndroidRuntime(26777): 	at java.util.HashMap$KeyIterator.next(HashMap.java:814)
E/AndroidRuntime(26777): 	at com.google.appinventor.components.runtime.EventDispatcher.delegateDispatchEvent(EventDispatcher.java:248)
E/AndroidRuntime(26777): 	at com.google.appinventor.components.runtime.EventDispatcher.dispatchEvent(EventDispatcher.java:219)
E/AndroidRuntime(26777): 	at com.google.appinventor.components.runtime.Button.Click(Button.java:51)
E/AndroidRuntime(26777): 	at com.google.appinventor.components.runtime.Button.click(Button.java:43)
E/AndroidRuntime(26777): 	at com.google.appinventor.components.runtime.ButtonBase.onClick(ButtonBase.java:656)
E/AndroidRuntime(26777): 	at android.view.View.performClick(View.java:4780)
E/AndroidRuntime(26777): 	at android.view.View$PerformClick.run(View.java:19866)
E/AndroidRuntime(26777): 	at android.os.Handler.handleCallback(Handler.java:739)
E/AndroidRuntime(26777): 	at android.os.Handler.dispatchMessage(Handler.java:95)
E/AndroidRuntime(26777): 	at android.os.Looper.loop(Looper.java:135)
E/AndroidRuntime(26777): 	at android.app.ActivityThread.main(ActivityThread.java:5254)
E/AndroidRuntime(26777): 	at java.lang.reflect.Method.invoke(Native Method)
E/AndroidRuntime(26777): 	at java.lang.reflect.Method.invoke(Method.java:372)
E/AndroidRuntime(26777): 	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:903)
E/AndroidRuntime(26777): 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:698)

When you test, make sure to check the case where the components local variables and when they are passed as parameters.

puravidaapps:
some more feedback:

generic_event_test1
I get the following error after clicking onto Button1:
*Argument (false) to 'apply-to-args' has wrong type (java.lang.Boolean)
(expected: procedure)*



generic_event_test2
same error after clicking onto Button1

and: There is no Do it functionality available for the setButton.Click
for component block... why?

the multi event test worked fine
I only tested together with your companion app (Version 2.24)

Taifun


Am 02.05.2015 um 10:47 schrieb gitars:

puravidaapps:
my test project still does not work
see test project attached
Taifun




Am 02.05.2015 um 09:36 schrieb gitars:

gitars:
Okay thanks for the feedback, I will start working on this issues.

@halatmit Thanks for the example. That log information seems to be really, really useful. How did you get those logs? Where do they come from?

@puravidaapps What do you mean by "Do it functionality"? Also, what "test project attached"?

josmas:
HI @gitars and all, trying to catch up with this thread...

Hal's logs are from logcat: http://developer.android.com/tools/help/logcat.html

*Do it* is available by right clicking in a block, but as far as I know there's no 'do it' for higher level handlers like Click and so forth, so might not be the best way to test this.

gitars:
@josmas Oh thanks!

Also, yes. It seems like there is no "Do it" functionality for even the normal event handlers. @puravidaapps

puravidaapps:
sorry, I was wrong... ;-)
I had to learn, that I have to adjust my blocks like this to see the
result "something" for button1 , works fine now
Taifun




Am 02.05.2015 um 12:31 schrieb Taifun:

halatmit:
Those log excerpts I included were the Android system log.  See  the App Inventor section on debugging:

http://appinventor.mit.edu/explore/learn/live-development-testing-and-debugging.html
If you look in runtime.scm, you'll see uses of the comments android-log, that writes entries to the log.
And you'll see Log.e used in the Java code similarly.

You'll also be wanting to look at the Java log.   And also be writing entries to the logs for debugging.

halatmit:
PS:   It sounds like you haven't used "do-it".  See the  App Inventor
documentation on debugging.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

puravidaapps:
yes I know, normal event handler do not have the "Do it" functionality
but all blocks which "set" something, for example set Label1.Text etc.
have it
that's only a "nice to have" feature... let's forget it...
Taifun
PS: it seems to be, you can't attach aia files in here...


Am 02.05.2015 um 13:33 schrieb gitars:

puravidaapps:
well, I have to come back to my first example somehow...
what is the expected result? see the following simple 2 button - 1 label
example:



the expected result is, after clicking button1 the label will display
"Yeah"...
this works fine!
the Button1.Click event was successfully overwritten by the generic event

Now see the following blocks using the multi generic event:




the expected result again is, after clicking button1 the label will
display "Yeah"
but now for both button clicks the label displays "A Button was clicked"...
_the Button1.Click event was not overwritten by the generic event...__
_this is an issue, isn't it? of course depending on "what is the
expected result"...

Taifun


Am 02.05.2015 um 13:37 schrieb Taifun:

gitars:
@puravidaapps I'm having a hard time following what you are saying. Could you provide screenshots of the blocks that go along with your example?

gitars:
@puravidaapps Also, actually, I think I need to update the warning handlers. If there is a multi-event block, I believe we are not allowing any other type of event block for the same event for that component.

gitars:
@halatmit I've started looking into example you gave, and it is interesting to me the result you got. The issue is actually deeper than what the logs indicate (that problem I think is relatively easy to fix). 

When I try to run the app. it simply doesn't work. This is because I have always assumed the parameter to the "for component" slot would be a block that I could call "block.instanceName" on, which is not the case here. As a result, the Yail for that block looks like "(define-generic-event undefined Click()...)" which really shouldn't do anything as "undefined" is not an actual instance of any component. To me it seems that I have call "blockToCode" on the block, and create Yail that looks something like (define-generic-event (lexical-value $item) Click()...), but that doesn't compile.

Anyway, I will continue to figure out this issue. It seems like I will have to make some change to runtime.scm which has historically taken me quite a bit of time to figure out. Also, this week is finals week for me, so forgive me if I am slow at responding this week.

halatmit:
Take your time to get it right.

I have not looked at your code, so I can't advise on the question you
asked.  But form looking at the thread I suspected you might have trouble
with the example I wrote.  Similarly, you need to handle the case where the
component is passed into a procedure as an argument.

In any case, any object passed around in the Yail needs to be referenceable
via a lexical lexical variable.  In your case, the object is the component
itself.    If your problem is that an expresion won't compile then Mark
Friedman is probably the best person to help with that.   Keep posting on
your progress and good luck with finals.


== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

puravidaapps:
(trying again to add images, now directly in Github)

what is the expected result? see the following simple 2 button - 1 label
t4-1

the expected result is, after clicking button1 the label will display "Yeah"...
this works fine!
the Button1.Click event was successfully overwritten by the generic event

Now see the following blocks using the multi generic event:
t4-2




the expected result again is, after clicking button1 the label will display "Yeah"
but now for both button clicks the label displays "A Button was clicked"...
__the Button1.Click event was not overwritten by the generic event...__
this is an issue, isn't it? of course depending on "what is the expected result"...

Taifun

puravidaapps:
not necessary, the warning handlers already work successfully as far as
I can see
Taifun

Am 02.05.2015 um 15:18 schrieb gitars:

gitars:
@puravidaapps  So about a month ago, we came to the conclusion that although it is unfortunate, if there is a multi-generic block, then we should not allow any other event handler for the same component/event (either generic or normal). That is something that explains the behavior you see. Following that assumption, we should not actually allow the generic event block for the button click as we already have a multi-generic event block for button clicks. From an implementation perspective, it happens because we store the multi-generic events by component name and the other events by instance name. Then, when an event is dispatched, we check if there is a normal/generic event to dispatch and dispatch it, and after we check if there is a multi-generic event to dispatch and dispatch it. In your example, it is likely that Button1.text was changed to "Yeah!", but then immediately changed to "A button was clicked" right after. For this reason, I think do in fact need update the warnings for event handlers so this can be made clear to the user.

mark-friedman:
I think that "Do it" functionality should definitely be supported for the 'generic' event handlers.  They aren't supported on regular event handlers because there wouldn't really be a point to it - the handler will get updated implicitly if the body is changed.  In the case of 'generic' event handlers you could have no change to the body but the target of the event could still get changed.  Think, for example, if the target was a variable.  If the value of the variable changes then the only way in live development to affect that change is via 'Do it'.

gitars:
I guess that is true. For now, however, I think that I will have to keep the "do it" functionality on the backlog as there are other more pertinent parts of the feature that are still not working 100% accurately yet.

mark-friedman:
That's fine.  I just want to make sure that it eventually gets done.  If
for nothing else, it helps in testing the new features.

>
-Mark

halatmit:
+1



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

gitars:
Wanted to let you guys know that I haven't forgotten about this. I finished finals and I'm going to start looking into some of these issues now.

halatmit:
thanks for letting us know.   I hope finals went well.  Keep us updated on your progress.

gitars:
@mark-friedman 
Hey, so I'm having trouble with types and I was wondering if you could clear up something for me. So if you want to set a button's text, if you do it in this way, you get the following Yail.

screen shot 2015-05-11 at 1 10 09 pm

(set-and-coerce-property-and-check! (get-component Button1) 'Button 'Text "text" 'text)


If you do it, this way, you get this Yail (for each button in the list):

screen shot 2015-05-11 at 1 14 25 pm

(set-and-coerce-property-and-check! (lexical-value $item) 'Button 'Text "text" 'text)


So it must be the case that (lexical-value $item) has the same type as (get-component Button1). First question is, what is the actual type?

Next, when defining generic events, the Yail looks something like this normally:

(define-generic-event Button1 Click()(set-this-form) .... )

However, if we are setting generic events while iterating through a list (so like the example @halatmit provided: https://cloud.githubusercontent.com/assets/789431/7442148/58a9aa0e-f0d2-11e4-9bb3-e1efa8145cd4.png), we have to generate Yail that evaluates to the same type, but I'm not really sure what that type is/how to. At first, I thought it would be (lexical-value $item), but that type seems to match up with (get-component Button1), not just "Button1", as shown above. Do you have any ideas about this?

Also, my next question is, sometimes I notice that an apostrophe right before a variable. What does that do? For example, in define-event, "(add-to-events 'component-name 'event-name)"

mark-friedman:
From the YAIL point of view (i.e. the sort of thing that is in coerce-arg in runtime.scm), the type of the "of component" slot needs to be 'component, which corresponds to the Java type com.google.appinventor.components.runtime.Component.

The apostrophe that you refer to is called 'quote' in Scheme/Lisp.  It creates a symbol.which is a datatype similar to an immutable string.  In the context of define-event, which is a Scheme macro, it's a little complex but basically if you have something like (define-event Button2 Click ...) the 'component-name evaluate to the symbol 'Button1 and the 'event-name will evaluate to the symbol 'Click

mark-friedman:
To expand a little on my earlier remark about types, the actual Java type for the specific example you give would be _com.google.appinventor.components.runtime.Button_, which is a subtype of _com.google.appinventor.components.runtime.Component_.

You also wondered about the difference between _(lexical-value $item)_ and _(get-component Button1)_.  For the examples above, they should be the same type.  In the first case you're getting the contents of the AI variable named _item_ (which will contain a Button component) and in the second case you're getting the component named 'Button1' (which will also be a Button component).  Part of the confusion might be it looks from the syntax of _(get-component Button1)_ that maybe _Button1_ refers to a Scheme variable named 'Button1' but that is not the case.  The way that the _define-generic-even_ macro is defined the first argument (i.e. Button1) is treated as a symbol, rather than a variable reference.

Hope this helps.

gitars:
Thanks! I'm sure this will help. I'll let you know if I have any more questions.

gitars:
So I am still having trouble with this. Currently I'm trying something like this:

"(define-generic-event '(lexical-value $item) Click()..."

because (lexical-value $item) gives the component and I thought maybe the quote would make it to a symbol (hoping that it would work in the same way Button1 really translates to the symbol 'Button1). 

I was hoping you could help figure this out for me: how can make the component-name argument the same as something like this:

"(define-generic-event Button1 Click()..."

if I am starting with "$item"

mark-friedman:
@gitars 
Ooh, this is actually a tough one.  The problem, as the component-name parameter in define-generic-event suggests, is that define-event-helper and registerEventForDelegation take a _component name_ as an argument, not a _component object_ as will be provided to the generic event block.  Currently, I don't think that there's a way to get a component's name from the component object.  Let me think a little more about this.

mark-friedman:
@gitars 
I think I know what to do here.  I'll update you with the details later today or tomorrow.

gitars:
Great, thanks!

halatmit:
Thanks for helping on this -- we really need it.

(And  I'm pretty stressed now, as you might appreciate.)


== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

mark-friedman:
So, the right thing to do would really be to overhaul the whole event registration and dispatching mechanism to use component objects rather than component names.  That would eliminate a lot of the complexity (and hard to understand Scheme macros) of the current approach.  However, I can certainly appreciate that at this point doing such a thing is probably more than you want to take on.

The next best thing is to just get the component's name and use the current mechanism.  The first step toward that was figuring out how to get the component's name, given the component object.  I believe that the following Scheme procedures will handle that:
```Scheme
;;; Environments are a map from names to values.  This procedure does a reverse lookup.
;;; I.e. it returns a name (or #f if none is found) given a value.
(define (reverse-lookup-in-environment value environment :: gnu.mapping.Environment)
  (let ((locations :: java.util.Iterator (*:enumerateAllLocations environment))
	(name-to-return #f))
    (let loop ()
	 (if (*:hasNext locations)
	     (let ((location :: gnu.mapping.Location (*:next locations)))
	       (if (eqv? (*:get location) value)
		   (set! name-to-return (*:getKeySymbol location))
		 (loop)))))
    name-to-return))

;; This procedure finds the name associated with the provided value in the current form environment
(define (reverse-lookup-in-current-form-environment value)
  (let ((env (if (not (eq? *this-form* #!null))
                 (*:.form-environment *this-form*)
                 ;; The following is just for testing. In normal situations *this-form* should be non-null
	       *test-environment*)))
    (reverse-lookup-in-environment value env)))
```

We still need to change `define-generic-event` a bit to handle expressions returning component objects rather than unquoted component names.  I don't have time right now, so I'll do that later today or tomorrow.

gitars:
Honestly, I think you are correct, in that this might be a bit more than I can take on. However, I'm free for a few days before I go on vacation and start work, and I wouldn't mind still trying to take a look into this until then. And I can certainly document changes I have made since the beginning.

mark-friedman:
I think that the following might work as a definition for `define-generic-event`:

```
(define-syntax define-generic-event
  (syntax-rules ()
   ((_ component event-name (arg ...) expr ...)
    (begin
     (let ((component-name (reverse-lookup-in-current-form-environment component))
	   (event-handler
	    (lambda (arg ...)
	      ;; The arguments to the handler come from the components and
	      ;; need to be sanitized before we can operate on them in Yail.  See
	      ;; the comments on sanitize below
	      (let ((arg (sanitize-component-data arg)) ...)
		(begin
		 expr ...)))))
       (if *this-is-the-repl*
	   (add-to-current-form-environment (gen-event-name component-name 'event-name) event-handler)
	 (add-to-form-environment (gen-event-name component-name event-name) event-handler))
       (com.google.appinventor.components.runtime.EventDispatcher:registerEventForDelegation
	(as com.google.appinventor.components.runtime.HandlesEventDispatching *this-form*)
	component-name
	'event-name)
       (list
	component-name
	'event-name))))))
```

Please give it (along with the the code I gave you yesterday) a try and let me know how it goes!

If this works, I would advise not worrying about overhauling the entire event mechanism for the time being.

gitars:
So now it doesn't work even for this app:

screen shot 2015-05-22 at 11 30 28 pm

producing this error:

screenshot_2015-05-22-23-29-21

The same error occurs when I create generic events from items in a list. Also, what should the Yail look like for the generic events when we set them from items in a list. Currently, I tried this two:

(define-generic-event $item Click()(set-this-form) ...)
and
(define-generic-event (lexical-value $item) Click()(set-this-form)

mark-friedman:
@gitars 

Could you send me the generated yail output for the program above?  I'll double check the code but having the exact output you'er generating will help.

As for your question about items in a list, that part of the generated code (i.e. '$item' vs. '(lexical-value $item)') should just be whatever is normally generated by the block in the 'component' socket.  You shouldn't need to add or remove anything.

gitars:
Yea that is true. Here is the Yail:

#|
$Source $Yail
|#

(define-form appinventor.ai_test.stillgood.Screen1 Screen1)
(require <com.google.youngandroid.runtime>)

;;; Screen1

(do-after-form-creation (set-and-coerce-property! 'Screen1 'AppName "stillgood" 'text)
 (set-and-coerce-property! 'Screen1 'Title "Screen1" 'text)
)

;;; Button1

(add-component Screen1 Button Button1 
(set-and-coerce-property! 'Button1 'Text "Text for Button1" 'text)

)

(define-event Button1 Click()(set-this-form)
    (define-generic-event Button2 Click()(set-this-form)
      (set-and-coerce-property! 'Button2 'Text "adsfadf" 'text)))
;;; Button2

(add-component Screen1 Button Button2 
(set-and-coerce-property! 'Button2 'Text "Text for Button2" 'text)

)

(init-runtime)

mark-friedman:
I think I had a typo in `define-generic-event` because I only tested it in a repl-type context.  Please try the following definition.  Note that the only change is the addition of a single-quote (for one of the `event-name variables)

```
(define-syntax define-generic-event
  (syntax-rules ()
   ((_ component event-name (arg ...) expr ...)
    (begin
     (let ((component-name (reverse-lookup-in-current-form-environment component))
       (event-handler
        (lambda (arg ...)
          ;; The arguments to the handler come from the components and
          ;; need to be sanitized before we can operate on them in Yail.  See
          ;; the comments on sanitize below
          (let ((arg (sanitize-component-data arg)) ...)
        (begin
         expr ...)))))
       (if *this-is-the-repl*
           (add-to-current-form-environment (gen-event-name component-name 'event-name) event-handler)
         (add-to-form-environment (gen-event-name component-name 'event-name) event-handler))
       (com.google.appinventor.components.runtime.EventDispatcher:registerEventForDelegation
        (as com.google.appinventor.components.runtime.HandlesEventDispatching *this-form*)
        component-name
        'event-name)
       (list
        component-name
        'event-name))))))
```

gitars:
For the same app, I now get this error:

screenshot_2015-05-25-20-00-37

mark-friedman:
Ok, I'll take a look.

mark-friedman:
Maybe the best thing would be for you to make your latest code available
for me, so I can build and run it.

gitars:
I put the companion app here:

https://www.dropbox.com/s/r8nsmq0la4acwwg/MIT%20AI2%20Companion.apk?dl=0

And I updated the appspot server here:

http://generic-events.appspot.com/#5730082031140864

If you try the app that I was referring to above, you get the same error.

halatmit:
I think Mark would like the source code so he can work on debugging it.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

gitars:
Oh sorry. The repo I'm working from is public, and I just added Mark as a collaborator. Is that enough?

gitars:
I also pushed the code Mark gave me and the other small changes I added.

halatmit:
That should be OK.

Mark?



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

mark-friedman:
Sounds good. I'll check it out

mark-friedman:
@gitars 

Sorry for the long delay, but I finally had a chance to download your code, build it and debug.  I think that I know what's going on and should have a fix for you soon.

mark-friedman:
@gitars @halatmit 

I believe that I have a fix for the generic event issues.  If you change the `Blockly.Yail.component_event` function in componentblock.js it appears to work (at least for me):

```
Blockly.Yail.component_event = function() {
  var body = Blockly.Yail.statementToCode(this, 'DO', Blockly.Yail.ORDER_NONE);
  // TODO: handle deactivated block, null body
  if(body == ""){
    body = Blockly.Yail.YAIL_NULL;
  }

  var header;
  var eventName = (this.isMulti) ? ("MultiGeneric-" + this.typeName + "-" + this.eventName) : this.eventName
  if (this.isMulti) {
    header = Blockly.Yail.YAIL_DEFINE_MULTI_EVENT + eventName;
  } else {
    var instanceName;
    if (this.isGeneric) {
      var selector = this.getInputTargetBlock("COMPONENT");
      console.log(selector);
      instanceName = callPrefix = Blockly.Yail.YAIL_SPACER
        + Blockly.Yail.valueToCode(this, 'COMPONENT', Blockly.Yail.ORDER_NONE)
        + Blockly.Yail.YAIL_SPACER;
    } else {
      instanceName = this.instanceName;
    }
    var defineType = this.isGeneric ? Blockly.Yail.YAIL_DEFINE_GENERIC_EVENT : Blockly.Yail.YAIL_DEFINE_EVENT;
    header = defineType
      + instanceName
      + Blockly.Yail.YAIL_SPACER
      + eventName;
  }

  var code = header
    + Blockly.Yail.YAIL_OPEN_COMBINATION
    // TODO: formal params go here
    + this.declaredNames()
          .map(function (name) {return Blockly.Yail.YAIL_LOCAL_VAR_TAG+name;})
          .join(' ')
    + Blockly.Yail.YAIL_CLOSE_COMBINATION
    + Blockly.Yail.YAIL_SET_THIS_FORM
    + Blockly.Yail.YAIL_SPACER
    + body
    + Blockly.Yail.YAIL_CLOSE_COMBINATION;
  return code;
}
```

I would have made the updates to your GitHub repo's generic-events branch but I wasn't sure if you'd be okay with that.

Also note that this is kind of the minimal required change to that funtion.  You might also want to change the `instanceName` variable to be something like `instance` or `component` since it now reflects something which could be either a component name or a component object.  You might also consider changing the name of `define-generic-event` in runtime.scm (and its references in Blockly) to something like `set-generic-event!` since it's really more of a `set`-type thing than a `define`-type thing.  Not to mention that it would better match the label on the block.

Finally, I note that there is no 'Do it' menu item available for the generic event block.  You really do want one.  Unfortunately, I don't know off-hand how to get the 'Do it' menu onto a block.  Maybe @halatmit or @jisqyv do.

gitars:
@mark-friedman Oh feel free to make changes on the repo. Also, I made the change that you posted and it seems to work great. So far I only tested it on 

https://cloud.githubusercontent.com/assets/4552296/7782266/29b6396a-00db-11e5-8d7f-4492537fd0a6.png

and 

https://cloud.githubusercontent.com/assets/4552296/7570632/9036e43c-f7df-11e4-9185-616e75470d6c.png

and they both seem to be working as expected. I'm going to update the appspot app and companion app for further testing.

gitars:
Oh and yea, I can start to make those changes. Also, yea I'm not quite sure where to even begin with the Do It functionality.

gitars:
So I updated the Companion app here:

https://www.dropbox.com/home/AI_Generic_Tests

and the appspot is once again here:

http://generic-events.appspot.com

@mark-friedman 
So actually I tested out some of the test apps I had from a while back with the companion app and I ran into some issues. Could you check out generic_event_test1.aia and generic_event_test2.aia in the same Dropbox folder?

mark-friedman:
@gitars 

That https://www.dropbox.com/home/AI_Generic_Tests dropbox link doesn't work for me, it's pointing to a non-existent folder in my home space.  I assume that you really mean https://www.dropbox.com/sh/3bimsqatvnyj7vc/AAClV3Ta0-8vRPXXs0qkzJNFa?dl=0

gitars:
Yup, sure do.

mark-friedman:
Hm.  Both of your generic test projects work fine for the version of App Inventor that I built based on the code in your repo (generic-event branch) plus my changes in both live development mode and with built apks.  If you sync the repo with your current code, I can check to see if my local code differs from that.  Also, if you haven't already you might try to do an 'ant clean' and then rebuild and rerun App Inventor and your local buildserver to see if that makes any difference.

gitars:
Yea, that must have been it. It seems to work for me now. I will start working on those changes you mentioned earlier.

gitars:
@mark-friedman So I think I made the changes you were asking for, if not let me know or feel free to push changes of your own.

mark-friedman:
@gitars 

Once again, please forgive the long delay in my response.  I've been fairly swamped.  Your changes look good.  Is there any chance you could figure out how to implement "Do it" on the generic event block or do you need more help for that?

By the way, congrats on all this work!

-Mark

gitars:
No worries. I haven't actually looked into it too much, but I can try and ask if I have any problems. Also, thanks!

gitars:
So I think what I changed makes it so that the generic events get a "Do It" functionality, and it seems to be working correctly. The one issue is that the block now has this blue question mark on it like this:

screen shot 2015-06-17 at 11 18 39 pm

and a this happens after I click do it:

screen shot 2015-06-17 at 11 18 47 pm

Does this happen with any other block (as it might help with debugging that issue)?

halatmit:
Try grabbing the (apparently) blank box at the lower right and expanding
it.

There's a bug that shows boxes that look empty, but in fact have content.




== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

gitars:
Oh okay. So this shows up:

screen shot 2015-06-18 at 2 22 54 pm

Is that an issue? What do you think should happen in the UI after the Do It button is clicked if anything?

mark-friedman:
I think that what it's displaying is a reasonable result value, i.e. simply the component and event names.

mark-friedman:
@gitars @halatmit @afmckinney 

It looks like this is ready for someone at MIT to take a look at.  The functionality appears to be done.

halatmit:
Excellent!

I'll see if I get some time over the weekend.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

gitars:
Sorry for not commenting earlier, I was gonna ask about that. Sounds great!!! Let me know if you find any  issues or feel free to make the changes yourself.

jisqyv:
This pr has been squashed and rebased to the current master in #450

gitars:
Awesome!! Thanks for everything!

PR #349, opened 2015-04-06
Double click to collapse or expand block

E-Hon:
@halatmit 

Please read the doc for details. If this change is a go, a new commit will be made so that the doc will not stay in master branch.


E-Hon:
**Description**
Double-click a block to collapse or expand it. The collapse or expand behavior is same as that from the context menu. It doesn’t collapse or expand its ancestors or descendants

**Sequence of events fired**
1. mousedown
2. mousemove, if any
3. ..
4. mouseup
5. mousedown
6. mousemove, if any
7. ..
8. mouseup
9. dblclick

**Remarks**
1. During a dblclick event, after the second mousedown and before the second mouseup event, if the block is dragged to the trash can, the associated callback is not called
2. In onMouseDown_(), the context menu of an unmovable block can be made shown so collapse or expand options are available for it. Hence, double-click can also apply to it
3. Since most clean-up work is done in onMouseUp_(), like terminateDrag_(), little is needed in the dblclick callback

halatmit:
This needs to be discussed with Neil, for compatibility with Blockly.

E-Hon:
You may try this change here.
http://ai2newfeatures.appspot.com

halatmit:
We decided not to do this.   Double click to expand a block could be confusing because people might double click accidentally.

PR #351, opened 2015-04-06
Rename media panel to files panel

E-Hon:
@halatmit 

Please read the doc for details. If this change is a go, a new commit will be made so that the doc will not stay in master branch.


josmas:
@E-Hon Please use this panel for comments instead of a pdf document.

Note that this may affect internationalisation; it can't be just simply changed in English.

halatmit:
Also note that there are lots of existing videos and tutorials that would
be affected by this change.
And "files" is rather nondescriptive and could be confusing.  You ought to
post this as a suggestion to powerusers and app-inventor-dev and
see what people say.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
Thank you for your submission. However given the comments from Hal above, we feel that the disruption caused is not justified by the potential benefit. I'm going to close this pull request.

E-Hon:
I see. I will put this aside for now. Thanks.

Issue #309, opened 2015-03-07
Need to create default initial value for Spinner


halatmit:
See this thread https://groups.google.com/forum/#!category-topic/mitappinventortest/Q2f0fyPsVk8

the user currently has to set the spinner selectionindex to a value (for example 1) to be able to do some data validation
probably a default value = 1 could be set automatically by App Inventor, in case the user forgot to do this?

EnisM:
My workaround for this was to create a dummy first item in the list... "Select Item..."

Then, when I do anything with the list, I check to make sure the selectionindex is > 1.

ReachAayush:
Hello, I was looking to fix this as I am also tackling issue #211. 

I was thinking that a possible solution would be to set the "String selection" to the empty string when the Spinner is created. Currently the selection string is null until a user selects something, which seems inconsistent with the fact that the default SelectionIndex is 0 if nothing is selected.

Please let me know if this workaround will not work.

EnisM:
Personally, I think initial selection has to be the first item.

In my case, I worked around this by adding a dummy "Select..." item at the
top of the list, and then in processing, I check the selectionindex, which
oddly enough always starts out as 1 even though the value is null, and if
selectionindex = 1 then put up an error message.

Ideally, the thing should have the first item selected, but that doesn't
trigger the AfterSelecting because a TRUE selection hasn't been made.  But
at least that way, at minimum if there's no user interaction, the first
item is the value that's returned.

A lot of people got tripped up by this because they put all of the stuff to
do for a selection in the AfterSelecting, which just doesn't trigger.  But
if they know that the first one is always selected automatically even if
there's no user interaction, they can provide the proper default settings
in their code to pertain to the first selection, without the AfterSelecting
getting triggered.

Either way, it's a tough call...

puravidaapps:
the default SelectionIndex is 0, so yes, empty string would be fine for
the default Selection imho
Taifun

Am 12.03.2015 um 17:38 schrieb EnisM:

halatmit:
#354 was closed

militrik:
Set "spinner selection index"=0 is not worked.
It always select first item from the list. And  "spinner selection index" =1 select first item from the list too.
Because of this, it is impossible to trigger the first element with Spinner.AfterSelecting
![01](https://user-images.githubusercontent.com/12719705/27949783-a3b61fc4-6307-11e7-9682-34179c657578.jpg)

PR #354, opened 2015-04-08
Need to create default initial value for spinner #309

E-Hon:
#309

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/youngandroid/YoungAndroidFormUpgrader.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/youngandroid/YoungAndroidFormUpgrader.java
@@ -293,6 +293,9 @@ public final class YoungAndroidFormUpgrader {
       } else if (componentType.equals("SoundRecorder")) {
         srcCompVersion = upgradeSoundRecorderProperties(componentProperties, srcCompVersion);
 
+      } else if (componentType.equals("Spinner")) {
+        srcCompVersion = upgradeSpinnerProperties(componentProperties, srcCompVersion);
+
       } else if (componentType.equals("TimePicker")) {
         srcCompVersion = upgradeTimePickerProperties(componentProperties, srcCompVersion);
 
@@ -1045,6 +1048,14 @@ public final class YoungAndroidFormUpgrader {
     return srcCompVersion;
   }
 
+  private static int upgradeSpinnerProperties(Map<String, JSONValue> componentProperties,
+      int srcCompVersion) {
+    if (srcCompVersion < 2) {
+      // A bunch of bugs were fixed
+      srcCompVersion = 2;
+    }
+    return srcCompVersion;
+  }
 
   private static int upgradeTimePickerProperties(Map<String, JSONValue> componentProperties,
       int srcCompVersion) {
--- a/appinventor/appengine/src/com/google/appinventor/client/youngandroid/YoungAndroidFormUpgrader.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/youngandroid/YoungAndroidFormUpgrader.java
@@ -293,9 +293,6 @@ public final class YoungAndroidFormUpgrader {
       } else if (componentType.equals("SoundRecorder")) {
         srcCompVersion = upgradeSoundRecorderProperties(componentProperties, srcCompVersion);
 
-      } else if (componentType.equals("Spinner")) {
-        srcCompVersion = upgradeSpinnerProperties(componentProperties, srcCompVersion);
-
       } else if (componentType.equals("TimePicker")) {
         srcCompVersion = upgradeTimePickerProperties(componentProperties, srcCompVersion);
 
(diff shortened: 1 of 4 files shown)


E-Hon:
@jisqyv @afmckinney @halatmit 
This issue is simply due to code inconsistency. Another issue, which is trickier, is also spotted.

AI's Spinner is old that it undesirably selects an item.

1. When adapter data goes from empty to non-empty, it selects the 1st.
2. After selecting the 3rd of a 5-item adapter, if the adapter data is changed to 2 items, the last of the 2-item adapter will be selected.

The above results in onItemSelected(...) being called undesirably

E-Hon:
You may try the test app here
https://www.dropbox.com/s/3yobkuw5y0d510i/SpinnerTest.apk?dl=0

josmas:
just a note to link this to #359

colinwirt:
This change removes line 128 from Spinner.java which was added by a poorly tested commit. That commit had the effect of breaking user selection after Elements is used to update the spinner from a list.

I have already updated my projects to avoid using Elements, now using ElementsFromString which wasn't broken by https://github.com/mit-cml/appinventor-sources/commit/094c19e4d256a3d74e1d089650305b1c2795edb8#commitcomment-10750680

halatmit:
@colinwirt  @jisqyv @josmas 
thanks.   I'll mark this as needing review.  Maybe Jeff or José can take a look.

josmas:
There's something weird happening to indices when re-setting elements. Although the issue highlighted by @colinwirt seems to be gone, if you have an scenario like the following blocks, you get an error under certain conditions:

error_setting_elements



**How to Reproduce:**
1) select number three from the first selection.
2) use Button1 to reset the values to 'a' and 'b'. At this point when setting the label, the spinner still responds with the latest selected, but because the first list has 3 elements and the second only 2, there's an error `invalid index 2`.


On a different note, I find a bit limiting the fact that the first item is always selected by default, and it cannot be selected. But that's a different discussion.

cc/ @E-Hon @halatmit @jisqyv

colinwirt:
Thanks for identifying that issue @josmas -- 

# By the way:
In my implementations I re-set elements to refresh the UI after creating or editing records.

The spinners that are relevant to this issue, allow the user to select and load the last 20 database records in a data collection system, so it's fine for the previous selection to be lost and for the first item to be selected in that use case.

# Placeholder
In this case, the first spinner element is a hint-type placeholder label for the spinner like "Open", so it doesn't need to be selected by the user.

Perhaps it would make sense for the first item in the list to (optionally) be a base item set from a placeholder property configured with a hint value like "" or "Make a selection" / "Not selected", etc.

# Restoring previous selection
Does it make sense if after the elements are set, the setter searches the list for the previous selection and restores that value if it can be found?

E-Hon:
@josmas 
I added some changes to fix the bug you mentioned. its cause is SelectionIndex is not updated when new data is fed and Selection() uses SelectionIndex for the item.

For 1st item is always selected, I think it's because the version of spinner we are using is old. I tried to write a native android4 app and found the default was select nothing. More precisely, as long as the data is not empty, our spinner always selects an item

E-Hon:
@colinwirt I think "placeholder" can be a hack from user side, meaning users knowing the first is automatically selected, they deliberately put a placeholder there.
I think a more profound solution would be to upgrade to use spinner of newer android versions

halatmit:
@E-Hon 
@jisqyv 
Will this work when we move to API level 4?

E-Hon:
@halatmit do you mean android 4, which is of at least api level 15? any guideline for me to use api 15 or above for appinventor?

halatmit:
@E-Hon @jisqyv @afmckinney 
I mean api level.
We're currently at minSDK API level 3.   We're going to move that to 4 in our test release, but we'll probably quickly increase that to api level 7 and possibly 11.  I don't think we should go higher than that for now on the spinner, and just focus on fixing the spinner issues.
If you want to see how to deal with this, look at how the Notifier handles level 14 (IceCreamSandwich) and above in setting the font size.

E-Hon:
@halatmit the bug raised by Jos has been fixed. the code can handle the issues related to "an item is always selected". it's just that this behavior is not the same as higher-api-level spinner.
this pr is ready for review

halatmit:
@E-Hon 
Look very nice.   The changes you put in are fine, but there's another tiny bug, which was there before and you should fix while you're here:   IIf you atempt to look at the spinner prompt (e.g. with DOIT get sprinner.prompt) before having set it to anything, you'll get an error.  I think the prompt simply needs to be initialized.

halatmit:
@E-Hon @liyucun 
Ethan, 
Please also include (and test) Yucun's change in #453 .  I've closed that issue.

E-Hon:
@halatmit
i added some commit to handle issue #442 
the root cause is the same so we don't need to hack the code

halatmit:
@E-Hon @afmckinney @jisqyv 
LGTM
requires a companion upgrade so I guess should go into ucr branch.

E-Hon:
@halatmit since i just changed the implementation code but didn't add or remove properties, do i need to upgrade the version? if not, i will revert the last commit

halatmit:
@E-Hon 
@jisqyv 
You should not need to change the version number for this, right Jeff?

jisqyv:
@halatmit @E-Hon A Version update is not needed. However it still needs to go on the ucr branch because it needs a new Companion and buildserver to take effect.

E-Hon:
@jisqyv thx Jeff. I just reverted the changes for versioning

jisqyv:
@E-Hon Thanks. I'm doing my final testing and integration for this change to hit the ucr branch

jisqyv:
LGTM

jisqyv:
@E-Hon MERGED onto ucr

PR #356, opened 2015-04-09
Generate Component Properties/Events/Methods translation map

jisqyv:
Generate Component Properties/Events/Methods translation map from the
source annotations.

Change-Id: I7001d14939e35283e077a54ba97edc4feb839494


jisqyv:
@afmckinney @halatmit @josmas This change eliminates the need to hand edit the translation maps. Instead we generate them dynamically. There is one catch. I create a source file at build time based on the annotations. I place this file in appengine/build/src/... There is a symlink to this file from appengine/src/com/google/appinventor/client/. I tried various ways to avoid this, but I could not find another way (other then writing into the source tree itself) that made the GWT Compiler happy.

Of course, I am open to suggestions (that work :-) ).

-Jeff

josmas:
@jisqyv haven't looked at the code, but are you doing it in the same way that the components json file is generated? It's also created at compile time by the annotation processor and needs to be available for GWT to read.

I'll have a poke at the code when I get a chance; won't make me stop complaining about internationalisation as a whole, but this is a great improvement; thanks Jeff!

jisqyv:
@josmas Yes, I leveraged the code that generates the json file. However instead of generating a json file, it generates Java source which is then compiled and fed to the GWT Compiler.

jisqyv:
MERGED

Issue #340, opened 2015-03-28
highlight selected item in listView


halatmit:
See here:

public boolean getItemsCanFocus ()
Added in API level 1
Returns

    Whether the views created by the ListAdapter can contain focusable items. 

public void ListView.setItemsCanFocus (boolean itemsCanFocus)
Added in API level 1

Indicates that the views created by the ListAdapter can contain focusable items.
Parameters
itemsCanFocus true if items can get focus, false otherwise

Once we set this, need to figure out how to make the individual element highlight when they are selected unhighlight when they are not selected.

halatmit:
see pr #357

halatmit:
#357 was merged

PR #357, opened 2015-04-11
highlight selecteitem in listView #340

E-Hon:
@halatmit 
Please have a look to see if it's enough for #340 

It only focuses the item visually. The item is not actually focused but the ListView is. There should be some code to hack it so that the item is actually focused. 

~~You may try it here.~~
~~http://ai2changes.appspot.com/~~

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
@@ -5342,8 +5342,8 @@ public interface OdeMessages extends Messages {
   @Description("")
   String FusionTablesStandardErrorMessage();
 
-  @DefaultMessage("HighlightColor")
+  @DefaultMessage("SelectionColor")
   @Description("")
-  String HighlightColorProperties();
+  String SelectionColorProperties();
 
 }
--- a/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
@@ -5342,4 +5342,8 @@ public interface OdeMessages extends Messages {
   @Description("")
   String FusionTablesStandardErrorMessage();
 
+  @DefaultMessage("HighlightColor")
+  @Description("")
+  String HighlightColorProperties();
+
 }
--- a/appinventor/appengine/src/com/google/appinventor/client/OdeMessages_zh_CN.properties
+++ b/appinventor/appengine/src/com/google/appinventor/client/OdeMessages_zh_CN.properties
@@ -476,6 +476,7 @@ SensorPortProperties = 传感器端口
 ScreenOrientationProperties = 屏幕方向
 ScrollableProperties = 允许滚动
 SecureProperties = 启用安全连接
+SelectionColorProperties = 选中颜色
 ServiceURLProperties = 服务地址
 ShapeProperties = 形状
 ShowFeedbackProperties = 显示交互效果
--- a/appinventor/appengine/src/com/google/appinventor/client/OdeMessages_zh_TW.properties
+++ b/appinventor/appengine/src/com/google/appinventor/client/OdeMessages_zh_TW.properties
@@ -477,6 +477,7 @@ SensorPortProperties = 感測器埠號
 ScreenOrientationProperties = 畫面方向
 ScrollableProperties = 允許捲動
 SecureProperties = 啟用安全連線
+SelectionColorProperties = 選中顏色
 ServiceURLProperties = 服務位址
 ShapeProperties = 形狀
 ShowFeedbackProperties = 顯示互動效果
--- a/appinventor/appengine/src/com/google/appinventor/client/TranslationComponentProperty.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/TranslationComponentProperty.java
@@ -233,7 +233,7 @@ public class TranslationComponentProperty {
     map.put("Hour", MESSAGES.HourProperties());
     map.put("Minute", MESSAGES.MinuteProperties());
     map.put("HasMargins", MESSAGES.HasMarginsProperties());
-    map.put("HighlightColor", MESSAGES.HighlightColorProperties());
+    map.put("SelectionColor", MESSAGES.SelectionColorProperties());
 
     // =========== ProximitySensor
     map.put("MaximumRange", MESSAGES.MaximumRangeProperties());
--- a/appinventor/appengine/src/com/google/appinventor/client/TranslationComponentProperty.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/TranslationComponentProperty.java
@@ -233,6 +233,7 @@ public class TranslationComponentProperty {
     map.put("Hour", MESSAGES.HourProperties());
     map.put("Minute", MESSAGES.MinuteProperties());
     map.put("HasMargins", MESSAGES.HasMarginsProperties());
+    map.put("HighlightColor", MESSAGES.HighlightColorProperties());
 
     // =========== ProximitySensor
     map.put("MaximumRange", MESSAGES.MaximumRangeProperties());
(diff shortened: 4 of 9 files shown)


halatmit:
@jisqyv 
Is there are companion we can use to test this?
Jeff, What's the right way for Ethan to provide a companion?  Do we have instructions about that?

josmas:
@E-Hon @halatmit simply provide a link to download the apk; I generally use a public dir in dropbox, but any other accessible way is valid.

E-Hon:
Thx @josmas . This will do.

Please download and install this apk to test the change
https://www.dropbox.com/s/ubz6e8fxc8apz6c/Highlight_ListView.apk?dl=0

thequixotic:
@E-Hon  , in the apk provided, I have to double tap the list view to make it focused. My device is gingerbread (2.3.6) . 
As for making it really focusable I think its not recommended when you are in touch mode. I have tried force fully focusing it, but it produces inconsistent touch behavior (touch doesn't work after focusing an item).  
(read this http://android-developers.blogspot.in/2008/12/touch-mode.html) 

When your not in touch mode (using Dpad or keyboard or trackball etc.. ) the focus works properly as its supposed to. 

I think what @halatmit meant is to give a visual feedback on the selected item. Which need not really focus the item. ?

thequixotic:
@E-Hon , you might want to try choice mode for ListView 

http://developer.android.com/reference/android/widget/AbsListView.html#attr_android:choiceMode

E-Hon:
@thequixotic Thx for the guidance. This is a better solution! I am going to try it

E-Hon:
Following @thequixotic 's suggestion, it now uses choice mode instead of focus. Further, I added a default selector color. @halatmit @josmas do you think selector color should be chosen by users?

Please try it here (apk)
https://www.dropbox.com/s/ubz6e8fxc8apz6c/Highlight_ListView.apk?dl=0

halatmit:
Nice!

Yes, I think users should be able to set the selector color as a property.


== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

E-Hon:
Please try the demo app with the latest change here
https://www.dropbox.com/s/5w69u1i4au3q929/Test0.apk?dl=0

halatmit:
Nice.
You may as well add the Chinese strings for the new property.

@jisqyv Does he need to add keys to the internationlization table, or is your code to do that automatically already merged into maaster?

thequixotic:
@E-Hon , have you tried this in on lower android versions. I have a gingerbread device  (2.3) and this is what I get ..

When I touch an item the entire listview (not just the item ) becomes highlighted with a random color and upon removing my finger all highlight goes. 

It does work well in my android 4.4 (device). I think App Inventor recommends to be consistent from android 1.5 .. but Good job. 

But I think the issue is that you setSelector for entire list view instead of just the individual view items generated by ArrayAdapter.

E-Hon:
@thequixotic Thx for your information. 

The code does setSelector for the entire list view. That method is not for individual items.

I only tried it on JellyBean as I don't have other android devices. Can you help writing native android code to test if it still occurs on gingerbread?

E-Hon:
@halatmit Besides OdeMessages_zh_TW.properties and OdeMessages_zh_CN.properties, any other files I should update?

The property is translated in Blocks view, but not in Designer
Imgur   Imgur

halatmit:
See issue #340

EnisM:
@E-Hon - Ethan, looked at the demo... exactly what I was thinking about and hoping for.  Very well done!

E-Hon:
@EnisM Thx, let me finalize it and let's hope it will get merged by then

E-Hon:
Hi @thequixotic I found out the root cause for what you described. 

According to this page, http://cyrilmottier.com/2011/08/08/listview-tips-tricks-3-create-fancy-listviews/
ColorDrawable, which I used, was buggy before Honeycomb. "it does not respect its bounds"

halatmit:
If this does not work before Honeycomb, then make sure the behavior degrades gracefully on older systems.

E-Hon:
As of commit 89aab4f, 

Test AI client 
http://ai2changes.appspot.com

Companion for the latest changes
https://www.dropbox.com/s/hxbbsi32dxun9tz/MIT%20AI2%20Companion.apk?dl=0

E-Hon:
@halatmit I changed to use GradientDrawable which worked well for that problem

EnisM:
Ethan:

Is this message about the thing we were discussing?  The ListView?  Do you
want me to test this?

E-Hon:
@EnisM ah no, I was referring to the problem pointed out by @thequixotic 
https://github.com/mit-cml/appinventor-sources/pull/357#issuecomment-92346019

EnisM:
Ok thanks...

halatmit:
Looking good!
Couple of comments:
- What is the scrollbar at the right of the listview in the designer for?  It doesn't seem to to anything. Maybe it's OK, since the scrollbar on the phone works.
- Look at how the listview behaves when the screen is scrollable:  It's very short.  That might be OK, but play with it.
- If you set the selection index programmatically (e.g., with do-it), the highlighting doesn't change.

I also tried this on Froyo (2.2.2), which has additional issues:
- the items don't remain highlighted when you lift your finger.    This is be OK if it's too hard to fix, but take a look at it. If we can't make it work back to Gingerbread, then OK, but leave a comment in the code.

EnisM:
It MAY be that that was the behavior in Froyo or Gingerbread...

halatmit:
Enis,

You should be able verify this with Genymotion.
So far, we want to support back to Gingerbread, but that might change with
API4.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

E-Hon:
@halatmit 
>What is the scrollbar at the right of the listview in the designer for? It doesn't seem to to anything. Maybe it's OK, since the scrollbar on the phone works.
I think this is the result of using GWT's VerticalPanel in MockListView

>Look at how the listview behaves when the screen is scrollable: It's very short. That might be OK, but play with it.
When Height is set to Automatic, this occurs (I guess ?). Is this behavior expected or a bug of Height property in ListView?

>If you set the selection index programmatically (e.g., with do-it), the highlighting doesn't change.
I am going to look into this

>the items don't remain highlighted when you lift your finger. This is be OK if it's too hard to fix, but take a look at it. If we can't make it work back to Gingerbread, then OK, but leave a comment in the code.
oh, I thought we just needed to support back to Gingerbread

halatmit:
>> I think this is the result of using GWT's VerticalPanel in MockListView
That's OK then.   Not critical to change

>>When Height is set to Automatic, this occurs (I guess ?). Is this behavior expected or a bug of Height property in ListView?
I think it's because screen height might return 0 when the screen is scrollable.  `See if you can make things better.
>>I am going to look into this
Good.  that needs to me fixed
>>oh, I thought we just needed to support back to Gingerbread
That's correct: We need to go only back to Gingerbread, so if it works on Gingerbread that's OK.  does it work there?

EnisM:
OK Hal... I'll give it a try in a little bit... Had a rough 2 weeks with my
Dad.

Refresh for me please... Gingerbread is what API?

halatmit:
Do a web search under android versions and you'll get the entire list.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

E-Hon:
@halatmit 
this feature is supported on Jellybean (API 19) or later

halatmit:
Good.  thanks.  Just comment on it in the source.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

E-Hon:
Added in commit b40f993

halatmit:
LGTM

jisqyv:
Note: This is a component change so will go to the "ucr" branch.

jisqyv:
@E-Hon @halatmit I'm doing my final review and testing on this in preparation for merging.

jisqyv:
@E-Hon @halatmit Note: This will be merged into the "ucr" branch for the upcoming Component Release

jisqyv:
MERGED into the ucr branch

PR #358, opened 2015-04-12
Fixed broken HELPURL for sensors

shrutirij:
The HELPURL for sensor components, accessed from the blocks editor, reached a page that does not exist. This commit corrects the URL in all languages.


halatmit:
LGTM

josmas:
@halatmit @shrutirij where are these helpURLs being used? Are they reachable from anywhere in the interface?

There are a couple of TODOs that seem to need fixing in the only lines I can see using these variables [here](https://github.com/mit-cml/appinventor-sources/blob/master/appinventor/blocklyeditor/src/blocks/components.js#L403) and [here](https://github.com/mit-cml/appinventor-sources/blob/master/appinventor/blocklyeditor/src/blocks/components.js#L646). Any ideas?

shrutirij:
@josmas if the user right-clicks on any block in the Blocks editor, there is a 'Help' option which opens the helpURL webpage. The page that opens should ideally be the reference doc for that component.

The sensor component blocks led to a broken link. Changed it such that it opens the correct reference page.

josmas:
@shrutirij yes, you are totally right, will remove the tag now... still puzzled about those TODO's! :)

shrutirij:
Thanks @josmas!

I'm not sure about those TODOs as well. I only noticed the bug while actually using AI. I'll have a look into what the code that uses helpURL exactly does too.

shrutirij:
@josmas were helpURLs supposed to serve any other purpose? That might explain the TODOs because the 'Help' option in the menu for each block doesn't seem to need any fixes.

jisqyv:
MERGED

PR #360, opened 2015-04-16
Fixed problem with designer-only properties not labeled in the designer ...

lizlooney:
...Properties panel.

For example, Screen1 has designer-only properties AppName, Icon, VersionCode, VersionName.

@jisqyv 

PS - Thanks for doing the work on the automatically creating the translation maps! It makes adding components/properties/methods/events a lot easier.

screen shot 2015-04-15 at 11 01 52 pm


jisqyv:
Good catch, thanks.

jisqyv:
LGTM, Merged.

PR #165, opened 2014-11-16
Add Maths Convert blocks

themadrobot:
This add Math blocks that converts:-

Decimal to Hexdecimal
Hexdecimal to Decimal
Decimal to Binary
Binary to Decimal

It validates the input is correct before converting

PR #365, opened 2015-04-19
Add math convert blocks

halatmit:


halatmit:
This replaces #165
Issue

halatmit:
This does not have the translations of the block names and tooltips.  Someone should add them in this PR, or do it after the merge or release.

jisqyv:
@themadrobot @afmckinney @halatmit I haven't played with this change yet, but I suspect it requires an upgrader for any projects that use the "is a number" block, as its xml representation has changed to include the pulldown.

halatmit:
Gareth,

Thanks,  Please check and let me know.




== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

halatmit:
@themadrobot 
I'm taking this pull request over.   There were some additional issues here, including a bug in Kawa!
I'll submit a new pull request and close this one when I'm ready

halatmit:
Hal notes about fixing this.  There's a bug when the inputs are very large integers that will cause errors to be signaled (e..g,  a bogus divide by 0 error).

(1) The input type on the convert blocks and the is-binary, is-hex, ...  should be text rather than any.  The problem is that the "any" type input might not be sanitized before being handed to the Kawa computation.  (Component method inputs are sanitized, but this is not a component method.)

(2) Kawa's string->number has a bug when the number is a large integer. Recode this in Scheme.

Test to see if this all works and also create unit tests.

halatmit:
This has been subsumed by #423 
Closing

Issue #331, opened 2015-03-23
add the 6.s198 doc on genymotion to the app inventor site.


halatmit:
remember to edit out part about changing adb

halatmit:
#367

halatmit:
#367 was merged

PR #367, opened 2015-04-21
Add documentation for using Genymotion emulator

halatmit:


jisqyv:
@halatmit Did you really intend to remove the index.html file in the doc/reference section and the sample .gitignore?

halatmit:
I think that index.html was a superfluous wrong one.   Am I correct about
that?

Bu you're right,  should have kept the sample-.gitignore.   I think I
created mu .gitignore by renaming sample, when I should have copied.  That
might be wrong in our instructions.




== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
I'm not sure about the index.html page. It is definitely in the old style, but the links on it are valid. I don't know if it is referenced anywhere.

halatmit:
I think that everything on it is duplicated here:

http://appinventor.mit.edu/explore/library.html

in the separate pages linked to at the bottom. Is that correct?



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

halatmit:
issue #331

jisqyv:
@halatmit I've restored the sample-.gitignore but left the index.html file deleted. If it turns out we need it, we can put it back.

jisqyv:
LGTM

jisqyv:
MERGED (master)

Issue #368, opened 2015-04-24
background issue when scrolling a list picker


halatmit:
Please see the following bug report:

https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/mitappinventortest/6068fs9UQlM/1rKmNIPZIpoJ

I verified this and saw the following behavior:
I created a project whose only element is a listpicker.   I set the background color of the listpicker to orange.  I set the elements of the listpicker to the 12 numbers 1,2,3,4,5,6,67,8,9,10,11,12, so that it is too long to fit on the screen and needs to be scrolled.
When I scroll the listpicker, the background turns black while my finger is touching.   When I lift my finger, the background generally  turns back to orange but sometimes remains black.

Note the comment from the user about getting different behavior when compiling under App to Market.

josmas:
@halatmit did you mean to close this issue?

halatmit:
Sorry, my mistake.  I'll fix it.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.
MIT Dept. of Elec. Eng. and Comp. Sci.
MIT Media Lab, Second Class Faculty

jisqyv:
#369 fixes this issue and it has been merged into the "ucr" branch pending the next component release.

PR #369, opened 2015-04-25
background-issue-when-scrolling-a-list-picker-#368

E-Hon:
@halatmit #368 

The issue is due to scrolling cache is enabled in ListPickerActivity. It also occurs if the apk is compiled with AppToMarket.

This change has been tested on emulator with API 10, 15, 18, 19, 21.

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/ListPickerActivity.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/ListPickerActivity.java
@@ -78,6 +78,7 @@ public class ListPickerActivity extends Activity implements AdapterView.OnItemCl
       String items[] = getIntent().getStringArrayExtra(ListPicker.LIST_ACTIVITY_ARG_NAME);
       listView = new ListView(this);
       listView.setOnItemClickListener(this);
+      listView.setScrollingCacheEnabled(false);
 
       itemColor = myIntent.getIntExtra(ListPicker.LIST_ACTIVITY_ITEM_TEXT_COLOR, ListPicker.DEFAULT_ITEM_TEXT_COLOR);
       backgroundColor = myIntent.getIntExtra(ListPicker.LIST_ACTIVITY_BACKGROUND_COLOR, ListPicker.DEFAULT_ITEM_BACKGROUND_COLOR);


halatmit:
I tried this and it's better, but still not completely fixed.   If I set thelistpicker itemBackgrondColor to orange, the background is black while the list is scrolling.  It now reliably changes back to ogange when I lift my finger.   So if you can't make it better, then we can leave it like this and put a comment in the source code.  But I thought you might want to take another look before we do that.

E-Hon:
@halatmit after the change, i can't reproduce what you described. which android version were you testing it on? did you update the companion by "ant installplay"? becasue "ant all" doesn't include installplay

halatmit:
@E-Hon 
I tested this on version 5.1.1 
and I did do ant installplay.
If you cannot reproduce it, then I'll go ahead an give it an LGTM and we can wait to see if other people are bothered by it.   I would not be surprised if another Android update changed the behavior again.

thanks for looking at this so quickly

halatmit:
LGTM

jisqyv:
MERGED INTO the "ucr" branch for the next component release.

halatmit:
Just so I understand:

People working on App Inventor should pull from master as always and not
worry about the ucr branch.  When the pull request moves to Gerrit, that's
when it goes into either master or ucr, depending on whether it requires a
component update.    But only you or Andrew (or I) need to worry about
ucr.  Do I have it right?



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
On Tue, Apr 28, 2015 at 9:38 PM, hal <notifications@github.com> wrote:

> Just so I understand:
>
> People working on App Inventor should pull from master as always and not
> worry about the ucr branch. When the pull request moves to Gerrit, that's
> when it goes into either master or ucr, depending on whether it requires a
> component update. But only you or Andrew (or I) need to worry about
> ucr. Do I have it right?
>

Yep, that's right.

-Jeff

halatmit:
I added some text (which people can ignore) to

https://docs.google.com/document/d/1sAw0QObTxTWqRX7GQRCa2z9TIV2r5AKT9UKMFF1acZI/edit#


== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

Issue #305, opened 2015-03-05
Improve language selection for text to speech in the designer


halatmit:
We could make this a dropdown showing the available languages.
Maybe something similar for countries, but we migth need a 2-D dropdown since there are so many countries.

E-Hon:
Instead of a 2-D dropdown, how about a scrollable dropdown? Does it serve the same purpose?

halatmit:
Good idea to use a scrollable drowdown!

E-Hon:
any default value for the two fields?

halatmit:
I don't know.




== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.
MIT Dept. of Elec. Eng. and Comp. Sci.
MIT Media Lab, Second Class Faculty

E-Hon:
AvailableCountries and AvailableLanguages do not exist in master branch...
But I can see them on the web

halatmit:
???  I see them in the master branch:

https://github.com/mit-cml/appinventor-sources/blob/master/appinventor/components/src/com/google/appinventor/components/runtime/TextToSpeech.java

line 336ff




== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.
MIT Dept. of Elec. Eng. and Comp. Sci.
MIT Media Lab, Second Class Faculty

E-Hon:
Oh, I forgot to pull the latest updates. my bad :p

halatmit:
:)



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.
MIT Dept. of Elec. Eng. and Comp. Sci.
MIT Media Lab, Second Class Faculty

E-Hon:
The 2 fields are device-dependent; only when AI is connected to a device/emulator can it retrieve the possible values. While it works in Blocks view, does the current Designer view support this sort of dynamic update?

halatmit:
The designer doesn't support dynamic update.  

But I dont' understand the issue.  There are lots of fields (e.g., location sensor info) whose value is not know unless a devce is connected.   Is this any different?

E-Hon:
It's that those dynamic values can't be input to the fields in Designer.
The values of a field in Designer have to be known before hand. For
example, emulator supports 5 countries while Galaxy Nexus supports more. It
doesn't know how many and which countries should be put to the fields in
Designer for users to select.

Or did I misunderstand this?

halatmit:
[...]

PR #380, opened 2015-05-04
Improve-language-selection-for-text-to-speech-in-the-designer-#305

E-Hon:
#305 @halatmit 

Add dropdown lists to display available languages and countries

The lists are blank as the values are device-dependent. They are placeholders for future work


halatmit:
@E-Hon 
I'm sorry  -- I was wrong.  I think the best thing is to simply remove these properties from the designer by making them ordinary properties, not designer properties.   Sorry to have misled you.
Can you make that change and mark the issue as ready for review?

E-Hon:
@halatmit 
just submitted a change. does it look good to you?

btw, i think what you want is to make dropdown lists for "Country" and "Language" so that user can pick one from each. am i right?

halatmit:
This isn't correct, from what I can see.    The actual fix is a lot easier:

- Get rid of the Language and Country properties in the designer palette (but keep the blocks).
To do this, all you need to do is remove the two @DesignerProperty annotation lines.

- If we get rid of these, we don't need the associated new property editor type classes that you created.   it's great that you figured out how to do this.  I'm sure we'll need to make new types.  but for now, just remove the code.

- When I start a new project, I get an error message "this project was created with a newer version of App Inventor.   So I guess there's a bug in the project upgrade code you put in.  I did not check on this, however.

I think the simplest thing is to remove all the files in the pull request and submit a simple change to master that removes those two designer property annotations, and bump the component version so that there's a record of what was changed.

PS:  Your'e right about me wanting the dropdown.  But we'll think about a better way to do it.  For now, the app developer can create a lispicker with the availalble languages and countries.

E-Hon:
@halatmit does it look good?

halatmit:
LGTM

jisqyv:
@E-Hon @halatmit I need to test to make sure that a project that had one of these properties set in the designer still loads after this change.

jisqyv:
@E-Hon @halatmit I've added the "on hold" label to just remind myself that I need to do this (likely will do the testing tonight, and if all is well will merge into ucr, otherwise you will hear from me :-) ).

halatmit:
I might have tested this, but it won't hurt to test again.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
@E-Hon @halatmit Indeed, a project that had a non-default value set for Language or Country will now fail to load after this change. You have to update YoungAndroidFromUpgrader to actually remove those properties if they have a non-default value. Take a look at line 768 for some code that is close to what you need. You may also want to put up a pop-up dialog (a window.alert should be sufficient) if the values are being removed to warn the App Inventor programmer that the properties were removed and that they may need to update their project to set them in the blocks (say when Screen1.Initialized is called).

halatmit:
Ick.  Thanks.

More for Ethan to do.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

E-Hon:
@jisqyv i don't quite understand the code around line 786. it makes more sense to me if the "else" is for the nested "if"
i get an error ("no upgrader to upgrade component type TextToSpeech to version 4") when i import an aia which has non-default "Language". but after refreshing and loading the same project, there's no error. is it the same error you mentioned?

jisqyv:
@E-Hon Sorry for the delay.

For the properties you are removing from the designer you should do
something like (I’ll use “Country” as my example):

<code>
   if (componentProperties.contains("Country")) {
   componentProperties.remove("Country"); // Country is no longer a
                                          //   designer property
   /* You might want to put code here to give the end-user an
      alert to warn them that you removed the property */
      }
</code>

The componentProperties array will only contain “Country” if it has a
non-default value.

E-Hon:
@jisqyv thx for the instructions. please have a look to see if it's good enough for this issue.

there's still an error. when importing an aia with a TextToSpeech whose "Language" is set to "en" which is non-default, "Blockly.Versioning.upgrade: no upgrader to upgrade component type TextToSpeech to version 4" is raised. yet when reopening the same project, the error is not shown

halatmit:
@fturbak 
Lyn, could you look at this?  I don't know if it's a bug in the upgrade code, or if Ethan is just missing something from the upgrader declarations.

halatmit:
@E-Hon 
That upgrader should go into versioning.js at around line 1895.
Can you finish this up now?

E-Hon:
Thx @halatmit . Sorry i missed this file

halatmit:
LGTM

jisqyv:
@E-Hon @halatmit So this change boils down to just removing the Language and Country properties from the designer. I am not sure now that this is the right thing to do. Although you cannot know the complete list of supported countries and languages until runtime, I am willing to bet that there is some subset that are supported universally, and these can be set in the Designer (using a pull-down). We should document that other languages may be available at runtime and that you can use the blocks to learn the list.

Comments?

E-Hon:
@jisqyv 
i found this 
http://stackoverflow.com/questions/7973023/what-is-the-list-of-supported-languages-locales-on-android
it seems that the list can be very long. 
also, when i try it on emulator (nexus 4, api 19), only 5 languages are supported. how should i verify the list?

jisqyv:
I would start with just those five languages.

E-Hon:
@halatmit what do you think?

halatmit:
@jisqyv @E-Hon 
i like jeff's suggestion

E-Hon:
@halatmit @jisqyv 
I added 2 new dropdown lists for Country and Language. The choices are from AvailableCountries and AvailableLanguages of TextToSpeech on AppInventor's Android 2.2 emulator.

I set the default country to "USA" and language to "en". these will probably work universally.

halatmit:
@E-Hon @jisqyv 
I think there is something wrong with the versioning here.  When I build this branch and create a brand new project, I immediately get "This project was saved with a newer version of the App Inventor system. We will attempt to load the project, but there may be compatibility issues."  And I don't see the for the new TTS properties.  Maybe I'm just doing something silly.
Ethan, does this work for you?  Jeff?

E-Hon:
@halatmit i just did a "ant clean" and "ant" and created a new project but didn't get what you said. i also imported a project with a TTS from http://ai2.appinventor.mit.edu to localhost and there's no error shown in ODE

halatmit:
Hmm....  I have a feeling that there's something wrong with my repository.
I'll investigate.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

halatmit:
@E-Hon OK.   I built a new repository and I no longer get that versioning issue.
Question:  If I start a new project, with no phone connected, and drag in a TTS component, what should I see in the dropdowns in the designer for the Country and Language properties?   Should they be drop downs, or just textboxes.   I see just text boxes.  Is that what we decided it should be?

E-Hon:
@halatmit for language, you should see a dropdown list with "de, en, es, fr and it"
for country, the list is much longer.
[Imgur](http://i.imgur.com/HVJpf8T.jpg?1)

halatmit:
@E-Hon @jisqyv 
This turned out to be a problem with my repository being corrupted.  The code here is fine.  Thanks.
LGTM

jisqyv:
@E-Hon @halatmit I'll test and merge this tomorrow. Sorry for the delay.

jisqyv:
@E-Hon @halatmit I'm sorry but there still is a problem. In the Designer you have the default set to US and "en". However at runtime you are using the devices default locale, which may be different. I'll leave a line comment where you need to make a change.

jisqyv:
@E-Hon @halatmit OK. Github won't let me place a line comment. On line 125 and 126 of TextToSpeech.java you should change the lines:

    Language("");
    Country("");

to instead be:

    Language("en");
    Country("USA");

To match the defaults in the Designer.

E-Hon:
@jisqyv just added a change.
sorry my tests weren't thorough...

jisqyv:
@E-Hon @halatmit MERGED into the ucr branch for the next component release (which is being put together now)

PR #381, opened 2015-05-06
Deselect projects

KeithGalli:
Added code so that when you select a bunch of projects in the projects view, switch to another view, and then switch back the projects don't remain selected


jisqyv:
@KeithGalli @halatmit LGTM, MERGED

PR #382, opened 2015-05-06
Delete projects in designer

KeithGalli:
Added functionality so that you can delete a project in the designer while you are working on it.

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/TopToolbar.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/TopToolbar.java
@@ -484,9 +484,7 @@ public class TopToolbar extends Composite {
         Project currentProject = ode.getProjectManager().getProject(currentProjectID);
         selectedProjects.add(currentProject);
         if (deleteConfirmation(selectedProjects)) {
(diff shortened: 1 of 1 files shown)


jisqyv:
@KeithGalli You should implement the save project prior to deletion (as we discussed in person) to avoid the race condition where a project is deleted while the workspace is "dirty." This results in an error when the autosaver goes to save the project, which is now deleted on the server.

This is an old bug, but it is exacerbated in the case where you are deleting a project directly from the Designer, because the likelihood of a project being "dirty" is much higher when you are in the Designer. In the traditional approach of first switching to the "Projects View" usually results in a long enough delay that the auto-saver has already triggered by the time the deleteProject request gets to the server.

KeithGalli:
@jisqyv Oops, I forgot to implement this change. I'll work on it later today and hopefully have everything fixed by tomorrow or Monday!

KeithGalli:
@jisqyv I created a new pull request for this change, see #404. Saving the dirty editors before deletion worked just as you thought it would. The one thing I wasn't quite sure about was how I should format this change.

PR #387, opened 2015-05-17
Firebase Component

kasmus:
Creating pull request with completed Firebase Component code from my UAP work.  Hal may want a few revisions made before this is actually pulled in.

My UAP final report with discussion of implementation and design decisions can be viewed here: https://docs.google.com/document/d/14UlxelRqW0GtMjJVZM9q_3hRjTbVNQm_7JZR8H-gmBg/edit?usp=sharing


appinventor/components/src/com/google/appinventor/components/runtime/FirebaseDB.java
```diff
@@ -0,0 +1,347 @@
...
+import com.google.appinventor.components.annotations.UsesLibraries;
+import com.google.appinventor.components.annotations.UsesPermissions;
+import com.google.appinventor.components.common.ComponentCategory;
+import com.google.appinventor.components.common.PropertyTypeConstants;
+import com.google.appinventor.components.common.YaVersion;
+
+import android.app.Activity;
+import android.os.Handler;
+
+// When the component is installed in App Inventor, the Javadoc
```
josmas: some comments (like the ones here or in line 44) are unnecessary.

appinventor/components/src/com/google/appinventor/components/runtime/FirebaseDB.java
```diff
@@ -0,0 +1,347 @@
...
+// three annotations stipulate that Firebase Component will appear in the
+// designer, that it will be an object in the App Inventor language,
+// and say what Android system permissions it requires.
+
+@DesignerComponent(version = YaVersion.FIREBASE_COMPONENT_VERSION,
+    description = "Non-visible component that communicates with Firebase to store and " +
+    "retrieve information.",
+    category = ComponentCategory.STORAGE,
+    nonVisible = true,
+    iconName = "images/tinyWebDB.png") //FIXME
```
josmas: //FIXME is rather explicit.

appinventor/components/src/com/google/appinventor/components/runtime/FirebaseDB.java
```diff
@@ -0,0 +1,347 @@
...
+    super(container.$form());
+    // We use androidUIHandler when we set up operations that run asynchronously 
+    // in a separate thread, but which themselves want to cause actions
+    // back in the UI thread.  They do this by posting those actions
+    // to androidUIHandler.
+    androidUIHandler = new Handler();
+    this.activity = container.$context();
+    Firebase.setAndroidContext(activity);
+    
+    firebaseURL = "https://incandescent-fire-4621.firebaseio.com/";
```
josmas: This cannot be hardcoded. The default value should error out and indicate the developer (the AI developer) that this property is missing and is required.

appinventor/components/src/com/google/appinventor/components/runtime/FirebaseDB.java
```diff
@@ -0,0 +1,347 @@
...
+   */
+  @SimpleProperty(
+      category = PropertyCategory.BEHAVIOR)
+  public String FirebaseURL() {
+    return firebaseURL;
+  }
+
+  /**
+   * Specifies the URL of the firebase.
+   * The default value is currently my private firebase url //TODO: this should be changed
```
josmas: it should be changed indeed; same for the default value in 172.

appinventor/components/src/com/google/appinventor/components/runtime/FirebaseDB.java
```diff
@@ -0,0 +1,347 @@
...
+  public void ProjectPath(String path) {
+    projectPath = path;    
+    resetListener();
+  }
+  
+  private void resetListener() {
+    // remove listeners from the old firebase path
+    myFirebase.removeEventListener(listener);
+    // set listeners for the new firebase path
+    myFirebase = new Firebase(firebaseURL+projectPath);
```
josmas: Is this reseting listeners or throwing away the object and recreating it?

appinventor/components/src/com/google/appinventor/components/runtime/FirebaseDB.java
```diff
@@ -0,0 +1,347 @@
...
+  
+  /**
+   * Asks Firebase to store the given value under the given tag
+   * if no value exists at that location yet.
+   *
+   * @param tag The tag to use
+   * @param valueToStore The value to store. Can be any type of value (e.g.
+   * number, text, boolean or list).
+   */
+  @SimpleFunction
```
josmas: Why two blocks to store values? This can be confusing for new users

appinventor/components/src/com/google/appinventor/components/runtime/FirebaseDB.java
```diff
@@ -0,0 +1,347 @@
...
+  // The @SimpleFunction annotation arranges for this to be a
+  // function (StoreValue)  associated with the component.
+  public void StoreValue(final String tag, final Object valueToStore) {
+    this.myFirebase.child(tag).setValue(valueToStore);
```
josmas: What happens if the write fails? 
I had a quick look at the API and it seems that it does not use Exceptions (which makes sense in an async API) but this method is not using a callback, which seems to be the way the API returns errors, so the user will never know if the value has been saved. There is a setValue(value, callback) version of the method which makes more sense to use in here, even if it's not exposed to AI devs though blocks, but at least to do something with the error if it happens.

appinventor/buildserver/build.xml
```diff
@@ -70,7 +70,8 @@
     <copy toFile="${classes.files.dir}/twitter4j.jar" file="${lib.dir}/twitter/twitter4j-core-3.0.5.jar" />
     <copy toFile="${classes.files.dir}/twitter4jmedia.jar" file="${lib.dir}/twitter/twitter4j-media-support-3.0.5.jar" />
     <copy toFile="${classes.files.dir}/fusiontables.jar" file="${lib.dir}/fusiontables/fusiontables.jar" />
-    <copy toFile="${classes.files.dir}/google-api-client-beta.jar" file="${lib.dir}/oauth/google-api-client-1.10.3-beta.jar" />
+    <copy toFile="${classes.files.dir}/firebase.jar" file="${lib.dir}/firebase/firebase-client-android-2.2.4.jar" />
+  	<copy toFile="${classes.files.dir}/google-api-client-beta.jar" file="${lib.dir}/oauth/google-api-client-1.10.3-beta.jar" />
```
josmas: indentation

appinventor/components/build.xml
```diff
@@ -125,6 +125,7 @@
         <pathelement location="${lib.dir}/twitter/twitter4j-core-3.0.5.jar" />
         <pathelement location="${lib.dir}/twitter/twitter4j-media-support-3.0.5.jar" />
         <pathelement location="${lib.dir}/fusiontables/fusiontables.jar" />
+      	<pathelement location="${lib.dir}/firebase/firebase-client-android-2.2.4.jar" />
```
josmas: indentation, here and in 230

jisqyv:
Superseded  by #660

PR #389, opened 2015-05-24
Upgraded dx.jar to the one from Android SDK Build-tools 22.0.1.

lizlooney:
This fixes the problem I've had with using third party jars containing classes compiled with latest javac in apps.

I have tested this by building HelloPurr and running it on a MotoX with Android 4.4.4 (KitKat), a G1 with Android 1.6 (Donut), and a G1 with Android 1.5 (Cupcake).

I will continue testing by using the MIT AI2 Companion app (built with this new dx.jar) on these devices.

@jisqyv Can you test if this fixes the issue you were seeing with kawa (unbound variables have a strange binding)?


jisqyv:
Cool. I plan to catch up on some reviews today. So I will certainly be
looking at this one.

-Jeff

jisqyv:
@lizlooney @halatmit I'm testing this change, so far so good. However it doesn't solve the problem with an unbound variable (say "x") returning "package $x" instead of throwing an unbound variable error!

Keep in mind this is still with the kawa.jar file compiled (years ago) targeting either Java 5 or 6, but not 7. The next step is to recompile the kawa.jar with a more modern compiler and seeing what happens.

I'm going to merge this change (assuming the other tests I'm running work, which I suspect they will) anyway, as using a newer version is probably a good idea. We should probably also change the build.xml files to target Java 7 instead of 5 or 6 when building jars for the components.

josmas:
@jisqyv does it make sense to merge this jar by itself? Would it be better to upgrade both android.jar and all the tools? There were some workarounds to keep compatibility with classic that are not needed anymore, although the reasons are rather vaguely explained (probably not well understood): https://github.com/mit-cml/appinventor-sources/tree/master/appinventor/lib/android/tools

cc/ @lizlooney @halatmit

jisqyv:
MERGED

jisqyv:
@lizlooney We have upgraded android.jar several times, and are likely do so again (for Bluetooth Low Energy).

Issue #281, opened 2015-02-22
Do-it on a local variable gives nonsense value


halatmit:
If you run do-it on a local varaible, you get a result like value:  "package $x" where x is the name of the variable.   If we can't make doiit work on locals, then at least we should give a sensible result.

EnisM:
Or eliminate the DoIt for the local variables altogether.

marksherman:
I know he's on vacation, but sounds like we need the opinion of @fturbak

halatmit:
@jisqyv @mark-friedman 
Jeff:  I've been looking with this, and I'm mystified.   After playing around, it looks like if you evaluate an unbound variable foo in Kawa lexical environment, the value returned is the list (package foo). 
I'm wondering whether this somehow has to do with the companion.  Maybe we could look at it together.
Mark:  Any idea?

jisqyv:
@halatmit You are correct. If you look at an unbound local variable in Kawa, you get the package object. I don't know if this behavior changes in a compiled environment.

-Jeff

halatmit:
I've been playing more and this does not happen if you type directly at the
kawa-1.11 jar that is in our lib directory.
I wonder where the difference is coming from.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.
MIT Dept. of Elec. Eng. and Comp. Sci.
MIT Media Lab, Second Class Faculty


On Sat, Apr 18, 2015 at 10:06 PM, Jeffrey Schiller <notifications@github.com

jisqyv:
I won't be able to play with this more until tomorrow. But I'm pretty sure
I was able to reproduce this with just the Kawa command line. I'll double
check tomorrow.

-Jeff

halatmit:
Let me know how you did this with the Kawa command line.   I tried, and
Kawa worked fine -- but I hope you're right since that would make a lot
more sense.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.
MIT Dept. of Elec. Eng. and Comp. Sci.
MIT Media Lab, Second Class Faculty


On Sat, Apr 18, 2015 at 11:11 PM, Jeffrey Schiller <notifications@github.com
> wrote:

> I won't be able to play with this more until tomorrow. But I'm pretty sure
> I was able to reproduce this with just the Kawa command line. I'll double
> check tomorrow.
>
> -Jeff
>

jisqyv:
[...]

PR #391, opened 2015-05-25
Signal an error on unbound lexical variables

jisqyv:
Work around an issue in kawa/dalvik where unbound variables were being
displayed as a package instead of being signaled as an error.

Change-Id: Ia4ac53b607a2fb935c8339e7d523b1f8043b4f2d

### DIFF
--- a/appinventor/buildserver/src/com/google/appinventor/buildserver/resources/runtime.scm
+++ b/appinventor/buildserver/src/com/google/appinventor/buildserver/resources/runtime.scm
@@ -248,7 +248,12 @@
 (define-syntax lexical-value
   (syntax-rules ()
     ((_ var-name)
-     var-name)))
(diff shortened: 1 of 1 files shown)


jisqyv:
@halatmit @fturbak This code causes the Companion to report an error on an unbound lexical variable. It needs a bit of testing. I'm going to send out a new Companion to the dev list for people to test with.

halatmit:
LSTM
(looks scary to me) :)
but please add a comment there explaining what is going on.  I've got a premonition that this will screw somebody, so we should record why we did this.

halatmit:
this fixes issue #281

halatmit:
This is issue #281

jisqyv:
Note: The latest dx.jar didn't solve the problem...

afmckinney:
LGTM

PR #402, opened 2015-05-31
Textbox Validation while creating a new AI Project

KeithGalli:
While a user is typing a project name, an error message will pop up and the textbox will turn red if an invalid character has been typed. It will stay there until the problem is fixed.

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/wizards/youngandroid/NewYoungAndroidProjectWizard.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/wizards/youngandroid/NewYoungAndroidProjectWizard.java
@@ -6,8 +6,12 @@
 
 package com.google.appinventor.client.wizards.youngandroid;
 
+import java.util.regex.Pattern;
+
 import com.google.appinventor.client.Ode;
+
 import static com.google.appinventor.client.Ode.MESSAGES;
+
 import com.google.appinventor.client.explorer.project.Project;
 import com.google.appinventor.client.tracking.Tracking;
 import com.google.appinventor.client.widgets.LabeledTextBox;
@@ -50,7 +54,10 @@ public final class NewYoungAndroidProjectWizard extends NewProjectWizard {
     projectNameTextBox = new LabeledTextBox(MESSAGES.projectNameLabel(), new Validator() {
       @Override
       public boolean validate(String value) {
+        //Pattern validText = Pattern.compile("[\\p{L}][\\p{L}0-9_]*", Pattern.UNICODE_CHARACTER_CLASS);
+        //String newString = validText.toString();
         if(!value.matches("[A-Za-z][A-Za-z0-9_]*") && value.length() > 0) {
+          disableOkButton();
           String noWhitespace = "[\\S]+";
           String firstCharacterLetter = "[A-Za-z].*";
           if(!value.matches(noWhitespace)) { //Check to make sure that this project does not contain any whitespace
@@ -62,6 +69,7 @@ public final class NewYoungAndroidProjectWizard extends NewProjectWizard {
           }
           return false;
         } else { //this is valid text, return true
+          enableOkButton();
           errorMessage = "";
           return true;
         }
@@ -145,7 +153,7 @@ public final class NewYoungAndroidProjectWizard extends NewProjectWizard {
     this.center();
 
     setPixelSize(width, height);
-    super.setPagePanelHeight(95);
+    super.setPagePanelHeight(85);
 
     DeferredCommand.addCommand(new Command() {
       public void execute() {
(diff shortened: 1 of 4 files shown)


halatmit:
@KeithGalli 
How do I test this?  I must be missing something, because if  go to create a new project and start typing the project name with a bad first character, I don't see any different behavior than happened previously with App Inventor.

appinventor/appengine/src/com/google/appinventor/client/wizards/youngandroid/NewYoungAndroidProjectWizard.java
```diff
@@ -43,7 +47,32 @@ public NewYoungAndroidProjectWizard() {
     // Initialize the UI
     setStylePrimaryName("ode-DialogBox");
 
-    projectNameTextBox = new LabeledTextBox(MESSAGES.projectNameLabel());
+    projectNameTextBox = new LabeledTextBox(MESSAGES.projectNameLabel(), new Validator() {
+      @Override
+      public boolean validate(String value) {
+        if(!value.matches("[A-Za-z][A-Za-z0-9_]*") && value.length() > 0) {
```
halatmit: The needs a regexp that can also account for international characters.

See

http://stackoverflow.com/questions/14636540/java-regular-expression-with-international-letters

KeithGalli:
@halatmit 

textbox_example

The above is what you should be seeing when there is invalid text in the textbox. The perimeter of the textbox is now red and there is an error message while the text is invalid

halatmit:
It looks like LabeledTextBox.java is not importing Validator.

KeithGalli:
Yea that's what the problem is. The import was working just fine when I originally made the change, but when I switched branches and switched back it stopped working. Is there anything special that I have to do when I add a new file to app inventor? Validator.java is a new class that I created for this change.

halatmit:
A new file is called an "untracked file" in git parlance.
You need to explicitly add it with the git add command.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

KeithGalli:
Hmm, I already explicitly added it with the "git add" command. When I make additional changes to the validator.java file it shows up in my tracked files on git.

jisqyv:
@KeithGalli When you indicate and error, you should also disable the "OK" button.

KeithGalli:
@jisqyv That makes sense, I'll try to add that functionality today.

appinventor/appengine/src/com/google/appinventor/client/wizards/youngandroid/NewYoungAndroidProjectWizard.java
KeithGalli: @halatmit What's the best way to test if it's working correctly with international characters? I added a regex to account for this today, but I want to make sure I'm properly testing it.

KeithGalli: I'm having trouble dealing with international characters. For this change it requires that I match unicode characters. The stack overflow post you linked me to gives two suggestions on how to do this. The first one requires me to import a couple classes from java.util.regex, but when I try to build app inventor after doing this, the build fails. The other suggestion includes the use of an embedded flag expression "(?U)". When I try this an internal error occurs whenever I type a character into the textbox. Any ideas?

KeithGalli: @jisqyv Any ideas on how to get the unicode regex expressions to work with app inventor?

KeithGalli:
@jisqyv I disabled the "OK" button in my most recent commit.

appinventor/appengine/src/com/google/appinventor/client/wizards/youngandroid/NewYoungAndroidProjectWizard.java
halatmit: see

http://stackoverflow.com/questions/10894122/java-regex-for-support-unicode

or do a web search under

java unicode regexp



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

KeithGalli: My problem is that GWT doesn't seem to support the java.util.regex libraries, so I can't use the methods described in that link. GWT uses a package called RegExp, but that doesn't have all the functionality of a pure Java implementation.

http://www.gwtproject.org/javadoc/latest/com/google/gwt/regexp/shared/RegExp.html

This package deals with regexes in the same way as javascript does. Unfortunately it seems like javascript doesn't have any built in libraries that are good for dealing with unicode characters so it is pretty difficult to match international letters. See the second response to this question.

http://stackoverflow.com/questions/150033/regular-expression-to-match-non-english-characters

halatmit: Ugh.   I don't have an idea


== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

halatmit:
@jisqyv @afmckinney 
I started reviewing this and it doesn't seem to be working.  As far as I can see, the validator is not being called.

jisqyv:
@halatmit I'll poke at it.

jisqyv:
@halatmit This code is working fine for me. I'm going to do a more complete review today with the intention if getting this code in tonight's release.

jisqyv:
@KeithGalli @halatmit LGTM

jisqyv:
@KeithGalli @halatmit MERGED

Issue #400, opened 2015-05-29
Image sprite rotation leaves does not fill enclosing rectangle on Android 5.1.1


halatmit:
See thread with attached aia file

https://groups.google.com/forum/#!category-topic/mitappinventortest/specific-programming-issue/EDMn85XAlg0

Momoumar:
Thanks @halatmit, I will look on it.

Momoumar:
Adding  ``` java rotatedDrawable.setAlpha(254) ``` just before ``` java rotatedDrawable.draw(canvas) ``` around **line 140** of the **ImageSprite.java** component seems to "fix the bug". 
I don't really understand why. 

There must be a nicer solution.

halatmit:
@jisqyv 
Momoumar, 
That sounds like the right fix, but I don't see the pull request in our public appinventor/sources repo.
I don't understand  why it's not there, since it's apparently in your fork.

halatmit:
pull request #403

halatmit:
handled in pull reuest #408

PR #403, opened 2015-05-31
Fix the black background that appears after an ImageSprite is rotated.

Momoumar:
… Ref issue #400 

aia for testing https://groups.google.com/group/mitappinventortest/attach/8ef5d5ec178873d4/spriteRotateBroken.aia?part=0.1&authuser=0&view=1

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/ImageSprite.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/ImageSprite.java
@@ -138,6 +138,9 @@ public class ImageSprite extends Sprite {
             // to get the other right and bottom edges
             xinit + w / 2 + rotatedBitmap.getWidth() / 2,
             yinit + h / 2 + rotatedBitmap.getHeight() / 2);
+        // The value of alpha is hardcoded to 254 to avoid the black Background that occurs
+        // after a rotation ref: issue #400
+        rotatedDrawable.setAlpha(254);
         rotatedDrawable.draw(canvas);
       }
     }
--- a/appinventor/components/src/com/google/appinventor/components/runtime/ImageSprite.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/ImageSprite.java
@@ -118,6 +118,7 @@ public class ImageSprite extends Sprite {
               0, 0,
               scaledBitmap.getWidth(), scaledBitmap.getHeight(),
               mat, true);
+          rotatedBitmap.setHasAlpha(true);
           // make a drawable for the rotated image and cache the heading
           rotatedDrawable = new BitmapDrawable(rotatedBitmap);
           cachedRotationHeading = Heading();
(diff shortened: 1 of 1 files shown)


appinventor/components/src/com/google/appinventor/components/runtime/ImageSprite.java
```diff
@@ -138,6 +138,9 @@ public void onDraw(android.graphics.Canvas canvas) {
             // to get the other right and bottom edges
             xinit + w / 2 + rotatedBitmap.getWidth() / 2,
             yinit + h / 2 + rotatedBitmap.getHeight() / 2);
+        // The value of alpha is hardcoded to 254 to avoid the black Background that occurs
+        // after a rotation ref: issue #400
```
halatmit: remove the comment line with the issue reference, since we can't assume that people working on App Inventor in the future will have access to that issue.

appinventor/components/src/com/google/appinventor/components/runtime/ImageSprite.java
```diff
@@ -138,6 +138,9 @@ public void onDraw(android.graphics.Canvas canvas) {
...
+        rotatedDrawable.setAlpha(254);
```
halatmit: A small point, pleases define that 254 as a constant and add a comment on why you used the value.  
Actually I thought Alpha will change to a float in api 11, so you might want to anticipate that (or maybe I am wrong about this).

Momoumar:
Thanks @halatmit.
It turned out that there is a special method for that : 
**public void setHasAlpha (boolean hasAlpha)**
Link : http://developer.android.com/reference/android/graphics/Bitmap.html#setHasAlpha(boolean)
I will use it instead.

halatmit:
Terrific!  It's great you found that.  bravo.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

Momoumar:
Thanks :+1:

halatmit:
LGTM

I'm checking with people on the Android team to try to find out why this fix became necessary, what's here now seems to work.

halatmit:
Merge this, but then look at issue #407 for a (hopefully) better fix.

Momoumar:
Thanks @halatmit, I'm going to have a look at it.

halatmit:
Do you mean issue #407?   If you can fix this now, then it should be
combined merged into pr #403.

I'm going to remove the note on moving #403 to gerrit, and wait for you to
try the method suggested in #407.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

Momoumar:
Ok I'm trying the third method of #407 to see if it fixes the bug we had.

halatmit:
thanks.   even if it brings that bug back, you should be able to fix it
with setAlpha as you did before, and it's a better drawing method (if it
works)


== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

halatmit:
subsumed by #408

halatmit:
closing

PR #404, opened 2015-06-01
Deleting Projects in the Designer

KeithGalli:
Added the ability to delete a project while you are working on it in the designer. To perform a deletion in the designer, click on the projects button in the top toolbar and then select "Delete Project". In order to make this change I had to save the dirty editors prior to deletion to prevent a race condition from causing an error.


KeithGalli:
Also, I can fix the deleting projects too quickly bug for the general case as well, but I figured I would see if the way I made the change for deleting projects in the designer was acceptable before I did that.

jisqyv:
@KeithGalli LGTM

jisqyv:
MERGED

Issue #290, opened 2015-02-28
Fix bug in clock documentation / methods


halatmit:
From Emily

While I was using AI2 with services, I noticed a mistake in the documentation for the Clock sensor component:

In MakeInstant(text from), the documentation says the instant is specified as "MM/DD/YYY hh:mm:ss or MM/DD/YYYY OR hh:mm", when I think it  means to say "MM/DD/YYY hh:mm:ss or MM/DD/YYYY hh:mm” — I looked online, and there are a few users who are having problems using this method because they are confused about the format of the input.

Similarly, I think there are some methods where the output format could be clarified, namely FormatDate and FormatDateTime. Because other methods ask for the date as MM/DD/YYYY, users may think that the FormatDate and FormatDateTime have an output in that format, when that is not the case.

Additionally, there should probably be a method that actually does return the date and time as MM/DD/YYYY hh:mm:ss. This would help users make apps with services because it would be easier to get an instant corresponding to a certain date and time — for example, they could easily make a task launch a notification on a certain date at a certain time.

halatmit:
assign to Grace Ryu

graceRyu:
I would like to focus on this part of the issue: 

“Similarly, I think there are some methods where the output format could be clarified, namely FormatDate and FormatDateTime. Because other methods ask for the date as MM/DD/YYYY, users may think that the FormatDate and FormatDateTime have an output in that format, when that is not the case.”

I am thinking to modify FormatDateTime and FormatDate methods so that they can both output 05/20/2015 and May 20, 2015. This can be done by adding a second parameter(format).  e.g. FormatDateTime(instant, format), FormatDate(instant, format), where the second parameter (format) specifies whether it should be MM/DD/YYYY or May 20, 2015.

halatmit:
I'm closing this because #405  was closed.

PR #405, opened 2015-06-01
Add formats to FormatDate and FormatDateTime

graceRyu:
I modified the FormatDate and FormatDateTime methods in Clock.java and Dates.java to take another argument such that users can specify date and time format e.g. MM-dd-yyyy or MMM d, yyyy.  This implementation will accept any formats accepted by java.SimpleDateFormat (http://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html). The current default format is not changed. 

To run "ant tests", I made some changes in DatesTest.java and ClockTest.java as well. 

Here is a test instance (http://issue290-ai2.appspot.com/)

@jisqyv @halatmit @ram8647 @josmas


jisqyv:
You need to increment the two component versions in YaVersion.java and increment the version of YOUNG_ANDROID_VERSION. You also need to provide an updater for projects that have older blocks. You need to replace any blocks that use the old versions to instead use your version, using an appropriate default pattern.

graceRyu:
I incremented the version number for YaVersion and Clock component. The updater for the old version is also provided. For the blocks that use the older version will get to have a default pattern filled in as a default argument. 

Also, another thing I changed is that the default pattern will be given to users like this so that people would get some ideas about how to utilize them correctly. 
screen shot 2015-06-17 at 3 18 58 pm

Here is a test instance (http://issue290-ai2.appspot.com/) -- same as the previous one

P.S.I was working in my branch called issue-#290 and merged it into review290 branch. That must be why I am getting so many commits for merging. 

@jisqyv @ram8647 @halatmit @fturbak

halatmit:
Jeff,

Could you sort this out so Grace can issue a pull request?



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

graceRyu:
I just pushed another commit to fix a default pattern.

appinventor/components/tests/com/google/appinventor/components/runtime/ClockTest.java
```diff
@@ -41,49 +41,49 @@ public void testGetMillis() throws Exception {
   }
 
   public void testAddYears() throws Exception {
-    assertEquals("Oct 11, 1943 9:30:00 AM", Clock.FormatDateTime(Clock.AddYears(tc, 2)));
+    assertEquals("Oct 11, 1943", Clock.FormatDate(Clock.AddYears(tc, 2), "MMM d, yyyy"));
   }
 
   public void testAddYears2() throws Exception {
-    assertEquals("Oct 11, 1931 9:30:00 AM", Clock.FormatDateTime(Clock.AddYears(tc, -10)));
+    assertEquals("Oct 11, 1931 9:30 AM", Clock.FormatDateTime(Clock.AddYears(tc, -10), "MM/dd/yyyy H:mm a"));
```
halatmit: This test (and others) are failing.   For example, this one gives 10/11/1931 rather than October 11, 1931.
And similar stuff for the other tests.

halatmit:
@graceRyu @jisqyv 
Grace,
This is very nice work!   Nice job on getting the default args into place, and on doing the upgraders.
The only problem I see is that the unit tests are failing.  Please take a look.

graceRyu:
Hi Hal,

I fixed the ClockTest.java I thought BUILD SUCCESSFUL meant passing the test, but apparently it's not. My bad. Now, you should be able to pass all the tests in TestClock.java 

@halatmit @jisqyv @ram8647

halatmit:
@jisqyv Tests pass.
LGTM

jisqyv:
Some comments. You need to update the Component Help String in OdeMessages.java (this is something that we should really fix). It is no longer sufficient to just update the description in the DesignerComponent annotation.

If you include clickable links in the documentation string, you must include a target="_blank" otherwise if someone clicks on the link they loose their App Inventor session.

Source code should not contain tab characters and should be indented properly.

I have taken care of this for you. I expect this change to be merged shortly (final tests running now).

jisqyv:
@graceRyu MERGED into the "ucr" branch for the next Component release.

graceRyu:
Hi Jeff,

Thank you for all your work.!I hope these changes are helpful to some
users. Also, thank you for the comments, I will keep in mind that in the
future.

Best,
Grace

Issue #265, opened 2015-02-15
Renaming components that have collapsed events


halatmit:
rename a component that has a collapsed  event handler , e.g. Button1.Click

this will not rename the collapsed block...
you first have to expand it to get the new name...

josmas:
Taking this over in befhalf of @shrutirij which will start looking into it when GSoC starts.

shrutirij:
Thanks @jos.

-----Original Message-----

shrutirij:
@halatmit @josmas, pull request #409 has been opened for this.

josmas:
This was merged at #409. Closing now.

PR #409, opened 2015-06-03
Fixed renamed component collapsed blocks.

shrutirij:
Issue #265 fix.

Try it at: http://ai2-experiments.appspot.com/

Document explaining the problem and changes [here](https://docs.google.com/document/d/1hPHrTNfzH_EOTxxcKqp7dL4Ht7wRY5GqVCW5qZrOh0I/). Please add comments, as required.

Thanks!

### DIFF
--- a/appinventor/blocklyeditor/src/blocks/components.js
+++ b/appinventor/blocklyeditor/src/blocks/components.js
@@ -170,7 +170,7 @@ Blockly.Blocks.component_event = {
     if (this.instanceName == oldname) {
       this.instanceName = newname;
       this.componentDropDown.setValue(this.instanceName);
-	  renameCollapsed(this);      
+	  Blockly.ComponentBlock.renameCollapsed(this);      
     }
   },
   renameVar: function(oldName, newName) {
@@ -341,7 +341,7 @@ Blockly.Blocks.component_method = {
       //var title = this.inputList[0].titleRow[0];
       //title.setText('call ' + this.instanceName + '.' + this.methodType.name);
       this.componentDropDown.setValue(this.instanceName);
-      renameCollapsed(this);
+      Blockly.ComponentBlock.renameCollapsed(this);
     }
   },
   getMethodTypeObject : function() {
@@ -564,7 +564,7 @@ Blockly.Blocks.component_set_get = {
       //var title = this.inputList[0].titleRow[0];
       //title.setText(this.instanceName + '.');
       this.componentDropDown.setValue(this.instanceName);
-      renameCollapsed(this);
+      Blockly.ComponentBlock.renameCollapsed(this);
     }
   },
   typeblock : function(){
@@ -676,7 +676,7 @@ Blockly.Blocks.component_component_block = {
       //var title = this.inputList[0].titleRow[0];
       //title.setText(this.instanceName);
       this.componentDropDown.setValue(this.instanceName);
-      renameCollapsed(this);
+      Blockly.ComponentBlock.renameCollapsed(this);
     }
   },
 
@@ -713,7 +713,7 @@ Blockly.ComponentBlock.createComponentDropDown = function(block){
   return componentDropDown;
 }
 
-function renameCollapsed(block){
+Blockly.ComponentBlock.renameCollapsed = function(block){
   if (block.isCollapsed()) {
     var COLLAPSED_INPUT_NAME = '_TEMP_COLLAPSED_INPUT';     
     block.removeInput(COLLAPSED_INPUT_NAME);        
--- a/appinventor/blocklyeditor/src/blocks/components.js
+++ b/appinventor/blocklyeditor/src/blocks/components.js
@@ -170,6 +170,13 @@ Blockly.Blocks.component_event = {
     if (this.instanceName == oldname) {
       this.instanceName = newname;
       this.componentDropDown.setValue(this.instanceName);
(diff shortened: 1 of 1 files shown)


appinventor/blocklyeditor/src/blocks/components.js
```diff
@@ -170,6 +170,13 @@ Blockly.Blocks.component_event = {
     if (this.instanceName == oldname) {
       this.instanceName = newname;
       this.componentDropDown.setValue(this.instanceName);
+      if (this.isCollapsed())
+      {
+        var COLLAPSED_INPUT_NAME = '_TEMP_COLLAPSED_INPUT';
```
jisqyv: We tend to use variables IN ALL CAPS as constants. I would recommend you choose different names for this variables.

shrutirij:
Thanks, I'll change that.

I used the same name as in line 1597 of ```/appinventor/lib/blockly/src/core/block.js``` - I thought it might be suitable as it's referring to the same string.

appinventor/blocklyeditor/src/blocks/components.js
shrutirij: Thanks, I'll change that.

I used the same name as in line 1597 of ```/appinventor/lib/blockly/src/core/block.js``` - I thought it might be suitable as it's referring to the same string.

shrutirij: @jisqyv, what variable name should be changed? 

```COLLAPSED_INPUT_NAME``` in this case is used as a constant string.

Is there a better way to write this code?

Thanks.

Comment (jisqyv) [inline appinventor/blocklyeditor/src/blocks/components.js]: If it is used as a constant, then that is fine (sorry if I missed that when
I first looked at the code).

-Jeff

Comment (shrutirij) [inline appinventor/blocklyeditor/src/blocks/components.js]: Is there anything else I should change?

There's an appspot instance with this feature: http://ai2-experiments.appspot.com/

Thanks!

File: appinventor/blocklyeditor/src/blocks/components.js [hunk 2]
```diff
@@ -340,6 +347,13 @@ Blockly.Blocks.component_method = {
       //var title = this.inputList[0].titleRow[0];
       //title.setText('call ' + this.instanceName + '.' + this.methodType.name);
       this.componentDropDown.setValue(this.instanceName);
+      if (this.isCollapsed())
+      {
```
josmas: Don't use C# curly brackets style here (and all if statements). Place them in the same line, as the rest of the code.

shrutirij: Thanks, @josmas. Just fixed this.

josmas:
@shrutirij are the 4 blocks of code exactly the same? is refactoring to a function possible?

shrutirij:
@josmas, I didn't do that because the ```rename``` function was implemented four times, although it's doing the same thing.

It's implemented individually for ```component_method```, ```component_set_get```, ```component_event``` and ```component_component_block```. Functions are implemented individually for each of these.

halatmit:
First, I agree with José:  Write that code only once, and call it as a method.

But second, This does not work for me.   download and build this as a branch.  Then I created a project with a single component named Button1.   I went to the blocks editor and pulled out the When Button1.Click block.    Then I collapsed the block.   Then I went back to the designer and renamed Button1 to Renamed.   Then I went the the Blocks editor.   The collapsed block still said When Button1.Click.   Expanding the block fixed things and changed the text to When Renamed.Click.

halatmit:
LGTM
nice job!

I sort of agree with José about not duplicating code.  But this is an edge case, so it's OK.

shrutirij:
Thanks, @halatmit!

Before I made the changes, the file had four functions called `rename`, with the exact same code (but different scopes) - so I added a few lines of code to each of these.

I assumed that it is convention, but is it more appropriate for such functions to be refactored into a global function (there are a couple of more examples in the same file)?

halatmit:
It's almost always a good idea to refactor so as not to duplicate code.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

shrutirij:
Thanks for the feedback, @halatmit @josmas 

I refactored my code into a function. Does it look better now?

Try it here - http://ai2-experiments2.appspot.com

halatmit:
LGTM
By the way, the general software principle here is called DRY (don't repeat yourself)

jisqyv:
@shrutirij @halatmit Actually, there is a problem. You declared your renameCollapsed function in the global scope instead of of in the local scope. Instead of

    function renameCollapsed(block)...

You should instead do:

  Blockly.ComponentBlock.renameCollapsed = function(block) ...

And then call it as Blockly.ComponentBlock.renameCollapsed.

shrutirij:
Thanks, @jisqyv. I've made the change.

jisqyv:
@shrutirij @halatmit LGTM, moving to Gerrit. Note: This doesn't handle non-component block renames. For example if you have a procedure block and a "call procedure" block. If the call block is collapsed, it doesn't get renamed when the procedure block is renamed.

Expect this pull request to be merged shortly. Note: If you want to work on fixing the procedure case, submit a new pull request (based on the soon to be updated master) rather then continuing to work on this one, which I will close when the merge is complete.

jisqyv:
MERGED

PR #413, opened 2015-06-07
Do not require GCS Bucket if not using GCS

jisqyv:
Change-Id: I4502c3e9d193fac79c1878d10a84aabb34846ae6


jisqyv:
@lizlooney Please review, this fixes the problem you pointed out where if you set use.gcs to false but don't setup a GCS bucket, things break (at least on older instances where a GCS bucket doesn't exist by default).

Issue #396, opened 2015-05-28
Rewite the image picker


halatmit:
The current image picker saves an extra copy of the picked image, which is unnecessary and overly complex.

There is a much easier implementation of the component which is illustrated by this use of the Activity Starter:
https://puravidaapps.com/snippets.php#2imagepicker

Rewrite the ImagePick using this method.

halatmit:
I'm abandoning this issue.  Making the copy seems to be the most straightforward way to handle things.   One could instead have the ImagePicker return a content URL.  But that would not work with the sharing component.

PR #415, opened 2015-06-08
ImagePicker component improvement

Momoumar:
Modify the ImagePicker Component so that the component doesn't have to save an extra copy of the picked image.
Ref issue #396  
An aia for testing: https://drive.google.com/open?id=0Bw0fmqX0oC9Yfk00Vnk4YWUzT1U2UVVHcVBNRW04TEJXcHFjUGlobnV2VWI5MkIyV3lQZ3c&authuser=0

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/ImagePicker.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/ImagePicker.java
@@ -27,8 +27,7 @@ import com.google.appinventor.components.runtime.util.MediaUtil;
 @DesignerComponent(version = YaVersion.IMAGEPICKER_COMPONENT_VERSION,
     description = "A special-purpose button. When the user taps an image picker, the " +
           "device's image gallery appears, and the user can choose an image. After an image is " +
-          "picked,the <code>Selected</code> property will be the path to the file containing the image that was selected" +
-          "the image that was selected",
+          "picked,the <code>Selected</code> property will be the path to the file containing the image that was selected",
     category = ComponentCategory.MEDIA)
 
 @SimpleObject
--- a/appinventor/components/src/com/google/appinventor/components/runtime/ImagePicker.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/ImagePicker.java
@@ -6,23 +6,8 @@
 
 package com.google.appinventor.components.runtime;
 
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


appinventor/components/src/com/google/appinventor/components/runtime/ImagePicker.java
```diff
@@ -102,114 +73,10 @@ protected Intent getIntent() {
...
-      // get the file type extension from the intent data Uri
-      ContentResolver cR = container.$context().getContentResolver();
-      MimeTypeMap mime = MimeTypeMap.getSingleton();
-      String extension = "." + mime.getExtensionFromMimeType(cR.getType(selectedImage));
-      Log.i(LOG_TAG, "extension = " + extension);
-
-      // save the image to a temp file in external storage, using a name
-      // that includes the extension
-      saveSelectedImageToExternalStorage(extension);
+      selectedImage = data.getDataString();
```
halatmit: This code code does not work for me.   Did you test it?    I wrote a simple program where AfterPicking sets the screen  background to the result returned by the picker and it does not work.    The issue here is that data.getDataString() is a content URL and you need to return an actual file path.   The code in the version of resultReturned your removed was doing that conversion by writing the result to a file in external storage.   That seems like a waste, but I don't know a better way to do it, and I didn't see a better method on stackoverflow.   Maybe you can find it.  I might have missed something.

Let me know if you want to just give up on fixing this.  But first look arond

halatmit:
I've been playing with this some more.  Another way to do this is to let imagePicker return the content URL.   But then you have to teach image and screenbackground and Canvas background to deal with content URLs.  This can be done.  But then I'm wondering about people who want to take the result of image picker and mail it somewhere. ..  this seems to be getting complicated.

halatmit:
Ignore previous two comments.   I need to look at this a bit more.

Momoumar:
@halatmit,
Yes, I tested it using the Image component (.aia file above) and it worked.
However, it seems not to be working with a screen background.

halatmit:
@Momoumar  @jisqyv 
It turns out that what you did was correct.  The screen background problem was because a screen background won't appear unless the background color is "none".   We should fix that and I'll file a separate issue.

But it turns out that this issue I posted is more complex than I thought.   I

Briefly, the ImagePicker should return a path to a real file rather than a content URL.  The content URL is fien for setting images and screen backgrounds and the like.   But it's not good for everything.  For example, the sharing component actually needs a file.   As far as I know, the most straightforward way to deal with this is to copy the picked image to a real file, which is what happens now.

We could work on avoiding that, but it starts to get complicated.  So let's hold off.   I'm sorry to have assigned this to you -- I didn't appreciated how complicated it would become. 

Closing now, but we might come back to this some day.

Thanks for working on this.

Issue #251, opened 2015-02-08
Fix the activity starter and documentation


halatmit:
The documentation in  "How to Use the Activity Starter" is obsolete.  And many of the examples are now wrong,  See

https://developer.android.com/guide/components/intents-common.html

Upgrade the activity starter so that is can handle most (all?) of the examples on the common-intents page, and maybe more.   This should  be.   App Inventor users should be able to provide a lot of keys rather than just the fixed set of keys used now.  Probably this should be done with a list of pairs ((key1 data1) (key2 data2) ....).
One issue to consider is how to deal with data types in the case where the values as no strings.

E-Hon:
@halatmit the tasks are to 
1. upgrade activity starter to support the common-intents listed on the site
2. allow users to use more than one key-value pair for extra data
3. update the obsolete doc
is it correct?

halatmit:
Yes, this is correct.   Note that some activities might use various keys for returning data.  
Thanks.

E-Hon:
@halatmit 
for 2, can we simply allow users to enter comma-separated values into ExtraKey and ExtraValue such that the first key corresponds to the first value, etc?

halatmit:
You can have a mode where users enter in CSV format, but also provide a way to specify the keys,values as  actual list of pairs.  It's similar to how ListPicker has both Elements (a list) and ElementsFromString (a comma-separated string).   You might want to restrict the CSV form to where the values a strings without spaces so you don't get into issues about quoting the input elements, or create an adhoc input notation.  In the long run, we need to provide dicsionaries and JSON input format, but I'd like to not get into that yet. 

On second thought, it might be better to not include the CSV format, so as to avoid these issues.   The people who would be using the activity starter would probably prefer the list-of-pairs form anyway, since most of those inputs would be computed and/or the actual call to activityStarter would be from another procedure.

E-Hon:
see #424

halatmit:
#424 was merged.  Closing this issue

PR #112, opened 2014-09-11
Adds code to mark blocks as bad when deprecated:

josmas:
- Modification of the annotation processor to include deprecated (userVisible
    = false) methods and events in simple_components.json with a deprecated
    property.
  - Modification of the creation of components in mainWorkspace to mark
    deprecated blocks as bad blocks.
  - Creation of a badBlock function to mark blocks as bad (highlight them in red
    and disable them).
  - Modification of the loading of blocks in Drawer.js to skip deprecated blocks
    in flyouts.

Review please: @jisqyv @afmckinney @halatmit @fturbak

PR #424, opened 2015-06-22
Upgrade activity starter. (Partially addresses #251)

E-Hon:
see #251 
@halatmit 
I added an Extras simple property to accept a list of key-value pairs to put to the intent.
I didn't use csv for the reasons you had stated.
Do you think we should remove the existing ExtraKey, ExtraValue so that the way to put extras is unified? 

after this, i will add support for those common intents


josmas:
Is the documentation supposed to be included here?

E-Hon:
Hi, @josmas thx for pointing it out. please see if it is clear to you

halatmit:
I haven't reviwed this yet.   When I do, I might split off the
documentation as a new issue.
I should probably write the documentation.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

E-Hon:
@josmas oh do you mean this doc?
https://docs.google.com/document/d/1nEflZ2OQc4tZ-6jGuVuqd66GFitCA-zuDLp3uKAqFwM/edit

halatmit:
Yes, that's the doc.   Do you want to do that update?   If not, I can do it
in a separate pull request.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

E-Hon:
@halatmit yea, i think it would be easier to manage if there's a separate pull request for doc so that this one is just for new property

halatmit:
I agree.   I'll file the new issue.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

halatmit:
@E-Hon 
New issue is filed.   I've marked this for review.  Jeff or I will take a look.

E-Hon:
@halatmit 
Do you think we should remove the existing ExtraKey, ExtraValue?

halatmit:
No, let's keep these so as not to break old projects.
Or maybe there's a way to arrange that new projects won't use them.
I"m not sure what the upgrader can do here to provide that functionality.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

appinventor/components/src/com/google/appinventor/components/runtime/ActivityStarter.java
```diff
@@ -359,6 +362,30 @@ public String ResultUri() {
     return "";
   }
 
+  /**
+   * Specifies the list of key-value pairs that will be passed as extra data to the activity.
+   */
+  @SimpleProperty
+  public void Extras(YailList pairs) {
```
halatmit: I don't think you've initialized Extras.   You should set it to an empty (Yail) list when the Activity Starter component is constructed.

E-Hon: just did. i thought it was unnecessary. but users may get the value before
setting it. the old code will return null to them

appinventor/components/src/com/google/appinventor/components/runtime/ActivityStarter.java
```diff
@@ -425,6 +453,15 @@ private Intent buildActivityIntent() {
       intent.putExtra(extraKey, extraValue);
     }
 
+    for (Object extra : extras.toArray()) {
+      YailList castExtra = (YailList) extra;
+      String key = castExtra.getString(0);
```
halatmit: Does getString work on all Yail data types?   If not, you'll need a way to ensure that these elements are strings, and we'll have to think about how to do that.

halatmit:
What happens in the case where one of the extras is DataUri?   
Also, can you provide a test aia for me to try?  I'm having trouble making this work, so I'd like to see a specific example of an action and extras that works for you.

halatmit:
@E-Hon Did you see my comment about the need to ensure that the elements in the list of extras are strings?   do you need help with that?

E-Hon:
@halatmit sorry for the delay. yes, i saw it. i am close to finishing my current task for cdk. i will continue soon as i have done that.

halatmit:
No problem. The CDK is more important to get done so we can move ahead on
it.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

appinventor/components/src/com/google/appinventor/components/runtime/ActivityStarter.java
E-Hon: yes, it works. I tried to log the key and value variable by adding 
`Log.d("ActivityStarter", "key: " + key + ", value: " + value);`
in the for-loop and they were printed correctly

E-Hon:
>What happens in the case where one of the extras is DataUri?

it is treated as string so it doesn't matter what it is

>Also, can you provide a test aia for me to try? 

hmm... i don't think you can use it in the client because it comes from a newer code base
>Or at least, we'd like to give good error messages to end users who specify extras as something other than strings.

"something other than strings" may include numbers and yaillist but is it a problem? both have toString() that gives them some meaning. and they can use the getter of Extras to see what they have put

btw, i think we can change YoungAndroidFormUpgrader to remove properties. i think it's better to have a consistent way to put extras. we currently have two, one from designer (single pair) and one from blocks (list of pairs)

halatmit:
I agree these all have toString(), but it's very unlikely that that these will be what the user intended, so it's better to give good error messages than leave things to the point where the user has to notice that calling the activity didn't work.

I'm not sure about removing the properties, unless you can do it with minimal pain to people who have projects thtat use the existing properties.  But give it a try.

E-Hon:
@halatmit you are right. simply using strings may not work properly as desired. For example, [alarm intent](https://developer.android.com/guide/components/intents-common.html#Clock) needs Integer for some extra value. using string won't work.

i propose the following changes
* support all value types in [common intents](https://developer.android.com/guide/components/intents-common.html) (i suppose ActivityStarter is for common intents only)
 * use new properties like ExtraUri, ExtraString, etc for each type. each of them accepts a key-value pair
* remove the current ExtraKey, ExtraValue, Extras

Consequently, users have to use the setter multiple times, just like in native android code.

i think this task is less trivial than i expected : D

halatmit:
Comments on your note:

>>>>support all value types in common intents
<https://developer.android.com/guide/components/intents-common.html> (i
suppose ActivityStarter is for common intents only)

Activity starter should work with all activities, including new ones that
have not been created yet.   So it needs to be general.  One use would be
as an extension mechanism, something like the CDK.    Note that you can
also use it to start other App Inventor Apps.

>>>   use new properties like ExtraUri, ExtraString, etc for each type.
each of them accepts a key-value pair

That's OK, but we also need to have the general property( Extras?) so that
list of properties does not get too large, and we can accommodate new keys.

>>> Remove the current ExtraKey, ExtraValue, Extras

OK to remove ExtraKey and ExtraValue but we need to keep Extras.   Also
need to create a smooth upgrade path for old projects that used ExtraKey
and ExtraValue.

>>>@halatmit <https://github.com/halatmit> you are right. simply using
strings may not work properly as desired. For example, alarm intent
<https://developer.android.com/guide/components/intents-common.html#Clock>
needs Integer for some extra value. using string won't work.

I'd like to get out something simple quickly, even if it is restricted
So I'd suggest that as a first version, all the values should be strings.

Later, we can extend this, so that the key,value pair can include
additional optional information such as data type. Maybe we could  extend
the (key value) notation to include an optional type so that an extra for
the ACTION_SET_ALARM might be (EXTRA_MINUTES, 20, int).   This needs some
thought.

I suggest you get it working for strings now, and then we can decide on how
to extend it to types other than strings.

kkashi01:
Just a note that some EXTRAS may require special eprmissions. For example, Hal's example (EXTRA_MINUTES) will require SET_ALARM permission

E-Hon:
@halatmit 
i had trouble removing ExtraKey and ExtraValue. Also, Liz stated in the [comment](https://github.com/mit-cml/appinventor-sources/blob/master/appinventor/components/src/com/google/appinventor/components/runtime/ActivityStarter.java#L158) that removing them has compatibility issues. So, should we go back to what you suggested, simply add Extras which accepts a list of pair of strings and keep ExtraKey and ExtraValue?

the test aia is [here](https://drive.google.com/file/d/0B22N7pfciq0CSUJrR25NcnRqVVU/view?usp=sharing) which supports the change up to the last commit

halatmit:
Removing ExtraKey and ExtraValue boills down to writing a suitable
upgrader.   That could be tricky to do automatically in this case.  And
alternative is to mark those blocks as "bad" and put up an alert to the
develop that those the use of those blocks needs to be replaced by Extras.
As an analogous example, look at the
upgrade of BluetoothClient from version 2 to version 3, where the
BluetoothError block was discontinued.

So you can remove those blocks.

For now, I'd restrict to where all the values are strings.   We can think
about how to handle non-string types later.   Maybe that method I suggested
abotu extending the pair with type info would work, but it will need some
thought.  So get the string version working first.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

E-Hon:
@halatmit the BluetoothClient example is in BlockSaveFile.java which is obsolete now and i can't find an example of removing property editors. 

how about i remove the designer property annotation and add `userVisible = false` and warn users if their projects have non-default values for ExtraKey or ExtraValue

halatmit:
@E-Hon 
I'm not sure that removing the designer property annotation and adding user visible is exactly the right thing, but you may be correct.  Try it and test it.  Make sure to test it with projects created with the current release,  to verify that upgrading works.

E-Hon:
@halatmit, i just added some changes. they involved removing the designer properties of ExtraKey and ExtraValue and adding `userVisible = false`
turns out when importing a project with ActivityStarter with ExtraKey or ExtraValue, 
`[WARNING] java.lang.IllegalStateException: no such property: ExtraKey`
 is raised and it can't reach the code in `YoungAndroidFormUpgrader` to warn the user

halatmit:
@E-Hon 
I don't see any changes to versioning.js
I think what you want there is to mark the blocks as bad, and show an alert to the user saying to use extras instead.    There are examples in versioning.js that you can use as models.   Does this solve the problem?

E-Hon:
@halatmit the change of versioning.js is in the very first commit (cfabdc7)

i can't find an example of marking blocks bad. the ones in versioning.js are obsolete; they are from BlockSaveFile.java

halatmit:
Hmm.   You're right.

Lyn,
We need to do an upgrade where we mark a block as bad.  How can that be
done?




== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

halatmit:
OK, here's the story on updates involving marking bad blocks.

Back in March of this year, José merged in some code
<https://github.com/mit-cml/appinventor-sources/pull/112/files#diff-6> so
that component events and methods could be marked bad simply by adding
userVisible
= false in their annotations. For example, here's such an annotation for
the doQuery method in
components/src/com/google/appinventor/components/runtime/FusiontablesControl.java:

      @SimpleFunction(userVisible = false,
          description = "DEPRECATED. This block is deprecated as of the end
of 2012.  Use SendQuery.")
      public void DoQuery() {...}

Nothing needs to be added to the upgrader in versioning.js in these cases,
because the annotation processor sets a deprecated flag on these events and
methods, and each of the corresponding blocks invokes the .badBlock()
method, which causes it to render as bad (i.e., with a red outline), and
invokes the .setDisabled(true) method, which disables the block. (Such
blocks cannot be re-enabled because many context menu options are turned
off for blocks marked bad.)

However, if you wanted to mark blocks other than component events and
methods as bad, you would need a helper function in versioning.js that
would find all the blocks in question and call .badBlock() and
.setDisabled(true) on each such block. I just wrote and tested such a
helper function, which appears below. Should I submit a pull request for
this?

/**
 * @param blockType: type of block to be marked bad. E.g.
'lists_pick_random_item'
 * @returns a function that maps a blocksRep (An XML DOM or workspace)
 *   to a modified workspace in which all blocks of this type have
been marked bad
 *   (i.e., they will be disabled and given a red highlight).
 *
 * @author fturbak@wellesley.edu (Lyn Turbak)
 *
 * Note that this method should *not* be used to mark component events
or methods as bad.
 * Component events and method are marked bad by adding
 *
 *     userVisible = false
 *
 * to the annotation for the event or method. A mechanism outside of
the upgrading in this
 * file takes care of marking the event and method blocks as bad. As
an example of this,
 * see the doQuery method in
components/src/com/google/appinventor/components/runtime/FusiontablesControl.java
 *
 */

Blockly.Versioning.markBlockBad = function(blockType) {
  return function (blocksRep) {
    var mainWorkspace = Blockly.Versioning.ensureWorkspace(blocksRep);
    var blocks = mainWorkspace.getAllBlocks();
    for (var i = 0; i < blocks.length; i++) {
      var block = blocks[i];
      if (block.type === blockType) {
        block.badBlock(); // Give block red highlight indicating it's bad
        block.setDisabled(true);  // Disable the block; blocks marked
bad cannot be enabled.
      }
    }
    return mainWorkspace; // Return the modified workspace, as
required by the upgrading structure.
  }
}

Here's how you would use the helper function in an upgrade to the language
version:

// Special upgrading map for language version.
// All dictionary keys except this one are component names.
"Language": {

  //... skipping the first 19 upgrades ...

  // THIS IS NOT A REAL UPGRADE, JUST A TEST OF markBlockBad
  20: Blockly.Versioning.markBlockBad('lists_pick_random_item')

}, // End Language upgraders


- lyn -

halatmit:
Ethan,

Can you try this method of marking the block bad by changing the annotation
and see it if works?

Lyn,

Thanks!! you might as well go ahead and submit that pull request, since
will probably be marking non-component blocks bad in the future.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

halatmit:
Hi Lyn,

Thank you for your explanation. It is very clear and helpful.

The method you created is just for non-component-blocks, right? because it
uses "type" as the argument.

What should I do with component-blocks, like setters and getters? Should i
do the same with methods and events that changes the corresponding
annotation processor to flag a block as deprecated?

Ethan

halatmit:
Ethan,

I think that Lyn is saying that for component blocks, you simply have to
mark uservisible = false, an change the annotation on the methods
(simplefunction) and properties.

The markBlockBad function is for the non-component blocks.

In your case, you only need to remove two properties from the
ActivityStarter, so you use the change-annotation method.   See if that
works.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

halatmit:
As Ethan correctly surmised, the code that I provided in my previous email
works only for non-component blocks.

From what I see in José's code,  the userVisible = false specification
works only for component events and methods, not properties. So I think
that would need to be extended to mark component property getters and
setters as bad.

- lyn -

halatmit:
This is messier that I thought -- I thought it was easy to mark a property bad.  I guess not.

Ethan, I suggest that you handle this by removing ExtraKey and ExtraValue from the properties panel and keep the properties blocks, but change the tooltips on them to say that they are deprecated and people should use Extras instead.    All you should need to do is to delete the @DesignerProperty annotations from these and add a description in @SimpleProperty.  I think this might not quite upgrade old projects exactly right, but it should be close enough.

E-Hon:
@halatmit 
i just added the description. Removing `@DesignerProperty` was done in f7c7ad8

halatmit:
Thanks.  Did you test this ?  Is the behavior ok when you load an old
project that uses those properties?  I did a little testing myself and it
seemed OK.   the old project still worked,



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

E-Hon:
Yes, i tested it. 

but in the debug panel, there's 
`[WARNING] java.lang.IllegalStateException: no such property: ExtraKey`
if an old project has non-default ExtraKey or ExtraValue in designer

halatmit:
Hmmm ... Can you take a little time to track that down, just to make sure
there's no real problem?



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

E-Hon:
@halatmit 
a small problem is that when an old project is imported, if ExtraKey or ExtraValue was set to non-default in designer, its value is wiped and can't be retrieved in Blocks 

i think this is unavoidable as we don't have a mechanism for removing property editors.

also, i didn't find any problem related to that exception

halatmit:
I think we can live with these flaws.   Are you ready for me to review ?




== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

E-Hon:
yes, please.

halatmit:
@E-Hon 
check line 453-454 in ActivityStarter.java    It requires the extra key and extra value to be initialized, or else getting the length will cause an error.  
I did a little testing and this was the only problem I found.

E-Hon:
@halatmit sorry i forgot to add back the code i once removed. just added it back

halatmit:
LGTM
@jisqyv @afmckinney Let me know when this is merged, so I can update the "Using the Activity Starter" documentation.

jisqyv:
@halatmit Doing some final tests. In particular I am concerned about what happens when we encounter a project that has set the "Extras" property in the Designer. Now that it is deprecated and removed from the Designer, I need to make sure that a project that set it in the Designer still loads correctly.

halatmit:
I think I tested that.  The project still loads but the initial values in
the designer are lost .  is that what you see?

jisqyv:
@halatmit @E-Hon Indeed the project loads. However although you no longer see the "ExtraKey" and "ExtraValue" properties in the designer, they are still present in the Screen.scm file after upgrading and saving with this change. I don't believe this is good behavior!

I suggest one of two solutions:

1. Have YoungAndroidFormUpgrader remove the two properties if they are present, and pop up a dialog indicating that they have been removed.  It should point to the documentation explaining what a project author needs to do.

   or

2. (My Preferred) Leave them alone as Designer Properties (so they don't disappear) and have YoungAndroidFormUpgrader pop up a dialog if they are present indicating that the projects uses these two deprecated properties and point to appropriate documentation explaining what a project author needs to do.

halatmit:
I prefer (2), but I'm not sure how to do it with the current upgrader.
Ethan, see if you can make it work.

By the way, if you set ExtraKey or ExtraValue and also include these in
Extras, the value in Extras will overwrite the others.   I think that's
fine.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.


On Fri, Jul 17, 2015 at 11:10 AM, Jeffrey Schiller <notifications@github.com

jisqyv:
> I prefer (2), but I'm not sure how to do it with the current upgrader.
> Ethan, see if you can make it work.

It should be able to be done with the current upgrader. Just see if the properties are present (there should be code already in the upgrader for other components that needs to see if a property is present) and call Window.alert(message [using OdeMessages]).

> By the way, if you set ExtraKey or ExtraValue and also include these in
> Extras, the value in Extras will overwrite the others.   I think that's
> fine.

And that is consistent to what happens in general when blocks set a property that had been previous set in the designer.

E-Hon:
@halatmit @jisqyv 
i added back the 2 designer properties and also a warning for upgrading a project that uses them

halatmit:
@jisqyv 
Does this look OK now?

halatmit:
@E-Hon @jisqyv 
Ethan: Can you resolve the conflicts and let us know if it's ready to go?

E-Hon:
@halatmit just merged the code to the master branch

halatmit:
@jisqv Jeff, please check and push to Gerrit if it's OK.

jisqyv:
@halatmit @E-Hon Checking...

jisqyv:
@E-Hon @halatmit I'm almost ready to merge this change. Note: I made some fixes, including making sure to only send an alert once when upgrading. I also added some logging to facilitate testing.

jisqyv:
@E-Hon @halatmit  LGTM

Preparing to merge into the ucr branch for the next component release

jisqyv:
@E-Hon MERGED into the ucr branch. You will want to abandon your topic branch because it is out of date to what was finally merged. See your commit on the ucr branch for the final version.

PR #426, opened 2015-06-22
Fixed Rendering Order of Background Color and Background Image

thequixotic:
Let me know for any changes.
Please do thorough testing :)

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/Form.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/Form.java
@@ -25,6 +25,8 @@ import android.content.ActivityNotFoundException;
 import android.content.Intent;
 import android.content.pm.ActivityInfo;
 import android.content.res.Configuration;
+import android.graphics.PorterDuff;
+import android.graphics.drawable.ColorDrawable;
 import android.graphics.drawable.Drawable;
 import android.os.Bundle;
 import android.os.Handler;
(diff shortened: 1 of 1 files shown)


appinventor/components/src/com/google/appinventor/components/runtime/Form.java
```diff
@@ -681,16 +681,9 @@ public int BackgroundColor() {
       defaultValue = Component.DEFAULT_VALUE_COLOR_WHITE)
   @SimpleProperty
   public void BackgroundColor(int argb) {
-    backgroundColor = argb;
-    if (argb != Component.COLOR_DEFAULT) {
-      viewLayout.getLayoutManager().setBackgroundColor(argb);
```
thequixotic: setBackgroundColor(argb) is same as setBackgroundDrawable(new ColorDrawable(argb))
Android does not seem to provide a way to setbackgroundcolor and setbackgroundimage simultaneously.
It treats a color as an Image (Drawable) . 
It does not differentiate between a color and an image both are drawables to android.
Which is why setting color would override the background image.
Its like assigning two values to the same variable, the last one assigned would prevail.
x = image;
x = color; 
color wins here. unless we blend both values when setting color...

jisqyv:
LGTM

jisqyv:
@thequixotic I slightly reformated the setBackground function just a little bit.

jisqyv:
@thequixotic MERGED into the ucr branch for the next Component Release.

PR #428, opened 2015-06-24
Watch, Icons 

emeryotopalik:
@halatmit @fturbak 

* I added Watch and separate icons/textboxes for Comment, Do It, Generate Yail, and Watch

Change-Id: Ib378eff98e5b67a6da660a3f8aa2790c88c8dd7d


halatmit:
@emeryotopalik Thanks for submitting this.
NB: documentation is at
https://docs.google.com/document/d/1MTiMXDUm5Z1Nx-5ghNqEPlCh9ZGw-Ke4K5UmnpMBshI/edit
and instance is running at
http://emery-johanna-watch.appspot.com/

fturbak:
It would be nice to include a link to the instance in the documentation.
Also, Emery, you need to include a way to download the modified companion,
because you cannot test the instance without using the modified companion.

- lyn -

emeryotopalik:
Okay, here is the companion: 

https://github.com/emeryotopalik/MIT_AI2_Companion/blob/master/MIT%20AI2%20Companion.apk?raw=true

jisqyv:
@emeryotopalik @fturbak @halatmit I'm concerned about the amount of changes to the blockly core. Can this code be re-factored to avoid that. It makes life very difficult when we go to upgrade the version of Blockly that we are using.

PR #434, opened 2015-06-26
Update Splash Screen Management

jisqyv:
Move the configuration of the Splash Screen into the Data Store. The
contents and size of the splash screen is fetched from the datastore
at runtime. The content can either be local, or a reference to an
iframe on another site.

Add an internal version number to the splash screen.  Provide a "Do
Not Show Again" Checkbox so that end-users can dismiss the splash
screen until the version is incremented.

Provide a python script "splashmanager.py" (in
appinventor/misc/splashscreen) to permit site administrators to update
the splash configuration. It requires the App Engine python SDK to be
installed.

Change-Id: I8b162fac35a1b3f2d0d78f1be6069f2d46c0f717

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/Ode.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/Ode.java
@@ -70,6 +70,7 @@ import com.google.appinventor.shared.rpc.project.GalleryService;
 import com.google.appinventor.shared.rpc.project.GalleryServiceAsync;
 import com.google.appinventor.shared.rpc.project.youngandroid.YoungAndroidSourceNode;
 import com.google.appinventor.shared.rpc.user.Config;
+import com.google.appinventor.shared.rpc.user.SplashConfig;
 import com.google.appinventor.shared.rpc.user.User;
 import com.google.appinventor.shared.rpc.user.UserInfoService;
 import com.google.appinventor.shared.rpc.user.UserInfoServiceAsync;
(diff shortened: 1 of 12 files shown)


jisqyv:
@afmckinney @halatmit Here is the splash screen update code.

appinventor/appengine/src/com/google/appinventor/client/Ode.java
```diff
@@ -1361,49 +1367,90 @@ public DialogBox createNoProjectsDialog(boolean showDialog) {
...
+      openProjectsTab();
+      return;
+    }
     // Create the UI elements of the DialogBox
     final DialogBox dialogBox = new DialogBox(false, true); // DialogBox(autohide, modal)
     dialogBox.setStylePrimaryName("ode-DialogBox");
     dialogBox.setText(MESSAGES.createWelcomeDialogText());
-    dialogBox.setHeight("400px");
-    dialogBox.setWidth("400px");
+    dialogBox.setHeight(splashConfig.height + "px");
```
halatmit: should these be dp rather than px?

appinventor/misc/splashscreen/splashmanager.py
```diff
@@ -0,0 +1,202 @@
+#!/usr/bin/env python
+#
```
halatmit: Very slick!

halatmit:
LGTM, although I did not actually run this.

appinventor/appengine/src/com/google/appinventor/client/Ode.java
jisqyv: No, these are on the Web interface not the phone.

jisqyv:
MERGED

PR #437, opened 2015-06-28
Add ability to disable an account

jisqyv:
Accounts are disabled by setting a URL in their general settings. This
URL is then displayed in an iframe when they login and App Inventor is
not available to them.

Change-Id: Iebf1c268358608d29a8e000935d0b9e4458f77f3

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/Ode.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/Ode.java
@@ -1812,6 +1812,33 @@ public class Ode implements EntryPoint {
   }
 
   /**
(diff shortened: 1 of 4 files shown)


jisqyv:
@halatmit @wolberd Here is the account deactivation code...

halatmit:
@jisqyv @wolberd 
How do we actually disable and reenable an account?

jisqyv:
@halatmit I didn't write the tool to do that yet (I tested it by directly editing the UserData object in the data store)

halatmit:
LGTM

PR #440, opened 2015-06-29
JSON Format Option

emeryotopalik:
@halatmit @fturbak 

I have created the option to use JSON Format for strings, lists, etc.

Documentation: https://docs.google.com/document/d/1L1UqXmMCCVEsjUYuq1tewPI9zprszc4nAwJpBm0E5OI/edit
Instance: http://emery-switching-json.appspot.com/
Companion: https://github.com/emeryotopalik/MIT_AI2_Companion/blob/master/json%20MIT%20AI2%20Companion.apk?raw=true


halatmit:
@emeryotopalik Go ahead and remove the option and just put in JSON formatting in all those places, and we can reviiew it.
Thanks

afmckinney:
I think we should uniformly change to json string through out but we need to check with Lyn.

fturbak:
I think all printed representations of all App Inventor values should be
consistent with JSON format.

There's also the issue of reading JSON into App Inventor values.  Because
App Inventor does not currently have dictionaries (it should, but that's a
separate issue), there's the question of how to turn JSON dictionaries into
App Inventor. I believe this is currently done as association lists, which
is fine for the time being, but would need to be changed if/when
dictionaries are added to App Inventor.

- lyn -

halatmit:
Lyn say to go ahead an make everything print in JSON format.

halatmit:
Hal is now working on this.

halatmit:
This is being handled in a different PR

PR #443, opened 2015-07-01
Add ContactURI property to Contact picker

halatmit:
@jisqyv


halatmit:
@jisqv  This pull request needs to be fixed because it was not built on master, so YaVersion and YaFormUpgrader are broken (at least).  Jeff, can you fix this, or is there something else I should do?
When this is finished, we should close #240

jisqyv:
This change is now handled in #445

PR #449, opened 2015-07-03
TimePicker and DatePicker: Instant getter, Date and Time setter with Instant

graceRyu:
TimePicker, DatePicker, and Clock components are all related to date and time. However, while most of Clock component methods take "instant" as arguments, there is no way to get an instant from DatePicker and TimePicker unless users manually create MM/dd/YYYY hh:mm:ss format by concatenating DatePicker.Month, DatePicker.Day, DatePicker.Year and so on. 

Thus, I added instant as a property in DatePicker and TimePicker. The initial date for DatePicker instant is set with current date, month, and year, and the initial time for TimePicker is set as current time. Since TimePicker does not have specified month, date, and year, the default date of the instant is set as Jan 01, 1970, following the Gregorian Calendar. (These are the same behavior as the way Clock component makes an instant.)

I also added two more methods, SetDateToDisplayFromInstant, SetTimeToDisplayFromInstant, which allow users to set Date and Time from the instant. 

Here are the instance and documentation. 

Instance:  http://grace-ai2.appspot.com/
Proposal: https://docs.google.com/document/d/1DmBi4iTSevmtREw5OD-x9kIWguPxyDt_NSZUfJN5Nmc/edit?usp=sharing

@ram8647 @halatmit @jisqyv

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
@@ -2493,6 +2493,10 @@ public interface OdeMessages extends Messages {
   @DefaultMessage("Icon")
   @Description("")
   String IconProperties();
+  
+  @DefaultMessage("Instant")
+  @Description("")
+  String InstantProperties();
 
   @DefaultMessage("IgnoreSslErrors")
   @Description("")
@@ -4297,6 +4301,10 @@ public interface OdeMessages extends Messages {
   @Description("")
   String HourMethods();
 
+//  @DefaultMessage("Instant")
+  //@Description("")
+  //String InstantMethods();
+  
   @DefaultMessage("MakeInstant")
   @Description("")
   String MakeInstantMethods();
@@ -5430,6 +5438,10 @@ public interface OdeMessages extends Messages {
   @Description("")
   String SetDateToDisplayMethods();
 
+  @DefaultMessage("SetDateToDisplayFromInstant")
+  @Description("")
+  String SetDateToDisplayFromInstantMethods();
+
   @DefaultMessage("IncomingCallAnswered")
   @Description("")
   String IncomingCallAnsweredEvents();
@@ -5461,6 +5473,10 @@ public interface OdeMessages extends Messages {
   @DefaultMessage("SetTimeToDisplay")
   @Description("")
   String SetTimeToDisplayMethods();
+
+  @DefaultMessage("SetTimeToDisplayFromInstant")
+  @Description("")
+  String SetTimeToDisplayFromInstantMethods();
 
   @DefaultMessage("XMLTextDecode")
   @Description("")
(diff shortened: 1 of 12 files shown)


halatmit:
@afmckinney @jisqyv @graceRyu @ram8647 
- I notice that one cannot get the instant from the time picker and use that to set the instant for the date picker, and vice versa.   why is that?  They look pretty similar.  How do either of them relate to the "instant" parameter for the clock setter methods.   That's just a number, right?               ?

- So it's confusing if clock, timePicker, and datePicker, all have a thing called "instant" and they are different things.   So (assuming it's not a bug that they are different) I suggest you give then different names.

In any case, this needs to be explained in the "working with times and dates" documentation that I suggested  you write.

- If these are not the same data type, then the "instant" parameter name on the set time/date from block, should be changed to "timeInstant" and "dateInstant", and you should make sure these parameter names are internationalized.  And maybe the clock "instant" should also get a different name.  There's also an "instantInTime" parameter in the code, but I don't think this appears any more at the user level.  (Although it is in the local language files, and it did appear in AI Classic.)

- See if you can add translations for instant (already internationalized) and timeInstant and dateInstant (after you internationalize them) to the varlous language translation lists.

See How to add a component:
 https://docs.google.com/document/d/1xk9dMfczvjbbwD-wMsr-ffqkTlE3ga0ocCE1KOb2wvw/edit#heading=h.sajmtj7l1183

halatmit:
That sounds ok, except let's not name them all "instant".    How about
clockInstant for clock, then DateInstant, and TimeInstant?

And why can't we get a clock instant from the time picker and data picker.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

graceRyu:
Hi Hal 

Thank you for noticing that bad behavior. It was a bug. You could set the TimePicker with the instant from DatePicker, but not the other way. I just fixed the bug and updated the instance. Could you try one more time? 

Instance: http://grace-ai2.appspot.com/

screen shot 2015-07-03 at 1 16 19 pm

This is what I tried and they worked fine. 

Like you said, what I intended to happen was to get a clock instant from DatePicker and TimePicker. For that purpose, I think the instant in Clock, DatePicker and TimePicker should stay as "Instant" (or "InstantInTime") How do you think?

halatmit:
@graceRyu @ram8647 @jisqyv 
Nice!
One nit to fix while your here is to add the local strings for the name "instant" for the instant property.
You can get those names from the translated name of of "instant" parameter.

graceRyu:
I just added "instantProperties" for four other languages. I took the Instant string from the InstantParams.

ram8647:
LGTM.  I think Grace has addressed the above requests.

jisqyv:
@graceRyu @ram8647 Doing final testing and integration. This code will be merged onto the "ucr" branch for the next component release.

jisqyv:
@graceRyu @ram8647 MERGED into the ucr branch

PR #458, opened 2015-07-09
Backpack: copy and paste blocks to different screens / projects 

graceRyu:
Backpack allows users to carry around blocks throughout their project repositories. The contents of the Backpack persist throughout an App Inventor session. 

Here is the link to the Google Docs for the user documentation. It includes a short video demo:
https://docs.google.com/document/d/1If7Lv0K8gWLEp63Jan1MZ1fewLwgINVCdz4VzfUvgt0/edit?usp=sharing

Here is the link to my instance: http://ai2-backpack.appspot.com/

Here is the link to the original proposal: https://docs.google.com/document/d/1Su5kC2whM4PIsXgzPTmVTCRaSS4TBxeh_bKSbGlSAd4/edit?usp=sharing 

@halatmit @ram8647


ram8647:
I've been reviewing and testing this feature as it has been developed by
Grace.  It looks good to me, but I won't do an official review until, say
Monday.  That will give others  a chance to have a look at this.

-- ralph

paulmedwal:
I have a higher level question. Why don't we implement copy and paste across screens instead of the backpack. When the user hits ctrl-c, you can save the xml of the block inside of a javascript object, and when they hit ctrl-v, then it pastes the block. Instead of storing the xml in the Blockly object (as I assume copy and pasting does right now), if it is stored in a global object (similar to Blocklies) then it would work across screens.

I think this implementation of the backpack is problematic on a number of levels. If you are going to allow users to have multiple items in the backpack, then it's important that users can inspect those items.

Currently, users have no way of knowing the contents of the backpack. They have no way of removing individual objects from the backpack (you either add something or delete everything). You also have no control over where the blocks from the backpack get pasted.

These problems are not as bad when there is only one block that you are dealing with. If you copy and paste one block, users could right click on the workspace, select paste, and have the block appear at that location. This behavior doesn't make sense when the backpack could have more than one stack of blocks inside of it.

I don't want to de-rail anyone's plans, but I really think that implementing copy and paste across screens is more intuitive, simpler, takes less screen real estate, and accomplishes almost all of the same goals as the backpack in its current implementation.

ram8647:
Backpack takes care of both copy/paste across screens and projects.  Teachers and students that we've been working with want both capabilities.  We used the drag-and-drop version of the backpack for an entire year at Trinity when I was running my own instance.  Teachers have been asking where it's gone?

The flyout for the Backpack works pretty much the same way as the flyout for the Blocks palette/toolbox on the left of the workspace.  It's a pretty simple feature that users will have little trouble learning to use.  I think we should go ahead with it and deal with user issues if and when they arise.

One potential problem with having the Backpack flyout open automatically when the user puts a block into it is that if the user then accidentally clicks on one of the blocks it will be pasted back into the workspace?

marksherman:
How long do these blocks stay in the backpack? 
Can you clear the backpack, removing what's in it?

ram8647:
The blocks remain during the entire AI2 session.  The background is
initialized when you login.    Yes, you can clear the blocks by right
clicking on the workspace and selecting  "Empty the backpack".

graceRyu:
DONE:
Add missing component alerts (warning flag)
Update flyout when a block is added while the flyout is opened
Rename “paste blocks from Backpack” → “paste all blocks from Backpack”
Add “copy all blocks to Backpack” option

graceRyu:
Hi everyone,

Thank you for comments and suggestions about the Backpack. It was a great
chance for me to see how to make Backpack better and more useful. I
apologize that I could not reply to each of the comments.

I was trying to get what people think must be done before the first release
and leave others to the enhancement list in addition to fixing some bugs.

DONE:

   1.

   Add missing component alerts (warning flag)
   2.

   Update flyout when a block is added while the flyout is opened
   3.

   Rename “paste blocks from Backpack” → “paste all blocks from Backpack”
   4.

   Add “copy all blocks to Backpack” option


ENHANCEMENT:

   1.

   Disable Duplicates to be added to Backpack:

As many people have said, it’s unnecessary to have duplicates in Backpack.
However, it will not keep people from using the main function of the
Backpack - copy and paste. I think we can do this for the second version.


   1.

   Fix the width of the flyout and have a horizontal scrollbar:

When we have a huge block that exceeds the size of main workspace, we might
zoom out the screen so that we can see the entire block. The same principle
is applied to the Backpack. When you have a large size of block in the
backpack and it takes over the entire workspace, you can easily zoom out
the screen and see all the blocks.

                            [ 100% ]



[image: Screen Shot 2015-07-15 at 5.56.55 PM.png]

          [75%]


[image: Screen Shot 2015-07-15 at 5.57.10 PM.png]

However, I think it’s good idea to modify the flyout so that it can be more
user-friendly and have less hassle. My plan is to fix the width of the
flyout and have a horizontal scrollbar instead of changing the width of the
flyout according to the largest block in the Backpack. In this way, users
will be able to see the blocks and still have the work space available no
matter how big the blocks in the backpack are.


   1.

   Make more user- friendly missing component alert

I added a warning handler for missing components. As of now, the warning
flag will be raised when there is a missing component, and the flag will
disappear once the component is added. The ideal scenario would be to ask
users if they want to add such missing components and we add those for
them.


   1.

   Improve UI


   -

   Leave some margin on top of the flyout so that users will not accidently
   paste blocks while closing the Backpack or better way to close the Backpack.
   -

   Add the title and total number of blocks with the backpack icon.


Unfortunately, this week is my last week of summer program, and I know it's
bad to leave some works behind and leave like this. But I am wondering if
we should wait until someone else to take over and finish up or if we can
merge it as it is and improve it when I come back.


Here is an instance: http://ai2-backpack.appspot.com

Source codes in GitHub:
https://github.com/mit-cml/appinventor-sources/pull/458


Thanks,

Grace Ryu

ram8647:
LGTM but Jeff or someone else must approve. 

Grace, Lyn, and I met on Thursday to discuss the last remaining issue on Lyn's list of must-fix items -- i.e., the issue of the flyout  width being too wide when the backpack contains extra wide blocks.  Grace has now fixed that issue.   

Lyn's review listed several issues.  These have all now been addressed:

1. Add missing component alerts (warning flag)
2. Update flyout when a block is added while the flyout is opened
3. Rename “paste blocks from Backpack” → “paste all blocks from Backpack”
4. Add “copy all blocks to Backpack” option
5. Fix the width of the flyout to see the left-side blocks

The following issues are being left as possible enhancements.

* Not allowing duplicate blocks in the backpack.
* Improvements to the UI
* Extending the backpack across multiple sessions.

The second of these would benefit from hearing what users say.  

Grace has done a nice job on this.

halatmit:
@afmckinney 
@jisqyv 
After this is approved, let's start by releasing it to PowerUsers and seeing if there are any problems.
@graceRyu 
@ram8647 

Grace, great job!  Can you send a pointer to the Google doc with documentation?  We'll need to post that when the backpack is released.  

Ralph, should we also do a blog post?  Can you draft it?

graceRyu:
Here is the updated video and the documentation. 

Video: https://www.youtube.com/watch?v=diQ8wJCYb6o

Documentation: https://docs.google.com/document/d/1If7Lv0K8gWLEp63Jan1MZ1fewLwgINVCdz4VzfUvgt0/edit#

halatmit:
@ram8647 @graceRyu @afmckinney 
Andrew, please take a final look at this and merge it if it's ready, and check with Josh  about where to put the documentation and how to release to powerusers

halatmit:
@afmckinney Andrew, please add a comment in this thread on the status of the review.

appinventor/blocklyeditor/src/blocklyeditor.js
```diff
@@ -156,6 +157,7 @@ Blockly.BlocklyEditor.render = function() {
   var timeDiff = stop - start;
   Blockly.Instrument.stats.totalTime = timeDiff;
   Blockly.Instrument.displayStats("Blockly.BlocklyEditor.render");
+  Blockly.WarningHandler.checkAllBlocksForWarningsAndErrors();
```
jisqyv: Why is this change here? This adds significant computation to an inner loop function. This can significantly slow down projects with a lot of blocks.

appinventor/blocklyeditor/src/backpack.js
```diff
@@ -0,0 +1,548 @@
...
+ */
+'use strict';
+
+goog.provide('Blockly.Backpack');
+
+/**
+ * Class for a backpack.
+ * @param {!Blockly.Workspace} workspace The Workspace to sit it.
+ * @constructor
+ */
```
jisqyv: Is this comment still accurate?

appinventor/lib/blockly/src/core/block.js
```diff
@@ -826,6 +832,20 @@ Blockly.Block.prototype.showContextMenu_ = function(e) {
       options.push(disableOption);
     }
 
+    // Option to copy to backpack.
```
jisqyv: We really want to minimize changes to blockly core. This can be done in the customMenuContext function in blocklyeditor.js (which is our code).

appinventor/lib/blockly/src/core/blockly.js
```diff
@@ -341,8 +348,7 @@ Blockly.onMouseDown_ = function(e) {
   if (e.target == Blockly.svg && Blockly.isRightButton(e)) {
     // Right-click.
     Blockly.showContextMenu_(e);
-  } else if ((Blockly.readOnly || isTargetSvg) &&
-             Blockly.mainWorkspace.scrollbar) {
+  } else if ((Blockly.readOnly || isTargetSvg) && Blockly.mainWorkspace.scrollbar) {
```
jisqyv: Did the code here change, or is this just a reformatting (which if it is, should be undone, as it makes merging new versions of blockly that much harder).

appinventor/lib/blockly/src/core/css.js
```diff
@@ -381,7 +381,10 @@ Blockly.Css.CONTENT = [
   '  fill: #ddd;',
   '  fill-opacity: .8;',
   '}',
-
+  '.blocklybackpackFlyoutBackground {',
```
jisqyv: Can this go in our localized css instead of core?

jisqyv:
@ram8647 @graceRyu I've done a first pass. Generally the code looks good. There are some whitespace issues, but GitHub makes them hard to find. Don't worry about them, I can clean them up when I squash the commits to merge them.

However one of the things we are really really trying to avoid is changes to blockly core. Merging in new versions of blockly are painful enough as it is with the changes we have. There are several places in the code (I pointed our a few) where changes were made to blockly core that can be made in our blocklyeditor tree. Someone needs to go over all of the changes to blockly core and see if they can be done via changes to our code instead.

appinventor/blocklyeditor/src/blocklyeditor.js
graceRyu: Hi Jeff,

This line was added to update the warning when users add a component for the block that its component is missing. In other words, this is how it would work. 

- When users paste a block from Backpack of which component does not exist, the red flag will be raised ' the component does not exist'. 
- Once the users add the missing component, then the red flag will be disappear (the line 160 checks errors and warnings and remove unnecessary flags)

Without the line, still the warning will be updated when users click the block or anywhere in workspace. This may not be as intuitive as updating the flags right away, but it does the job.

Let me know if it is okay without the line 160, or there is a better way to do it. 

http://grace-ai2.appspot.com/
This is the Backpack instance and I just commented out the line160. 

Best,
Grace 

@jisqyv @ram8647

jisqyv:
@graceRyu @ram8647 Can you hook the location where the pull down on the block selects the appropriate component?

jisqyv:
@graceRyu @ram8647 I found a better place to do the warning check that isn't in an inner loop. In particular I make sure it gets called when a new component is added. I'll post the location where you can see the code shortly...

jisqyv:
@graceRyu @ram8647 Here are the changes I have made:

https://github.com/jisqyv/appinventor-sources/commit/ece68ba75b43332859d35295f9d46d50076c3d00

This code will be going out on http://ai2-test.appinventor.mit.edu momentarily and we plan to release TONIGHT!!! :-)

jisqyv:
@graceRyu LGTM

NOTE: I expect this code to be merged within minutes (after Jenkins runs the unit tests). I will then close this pull request. The code has been "squashed" down into once commit which will be on the master branch when this PR is closed.

Please do future work on a fresh branch starting from master. I have made significant whitespace changes that will result in merge conflicts if you continue to work on this branch.

jisqyv:
@graceRyu @ram8647 MERGED

PR #459, opened 2015-07-10
Use BlobKey directly to read blobs from the blobstore!

jisqyv:
No longer need the files API to read files stored in blobs. Note: This
code assumes that the MapReduce job to compute the blobkeys for blobs
has been run. DO NOT USE THIS COMMIT until that has been run! We also
assume that we are not longer writing new objects to the blob
store (which requires the files API).

Change-Id: Icef104de53df0233a8876e2c7a3f26407c5b7560


jisqyv:
@halatmit @afmckinney Note: The necessary MapReduce job is almost finished running on ai2...

afmckinney:
Jeff LGTM.

Issue #456, opened 2015-07-07
`&` in AppName causes apk build failure


fmntf:
If Screen1.AppName contains `&`, the apk build fails with:

`WARNING: YAIL compiler - AAPT execution failed.`

Should we prevent users to put characters like `&` in AppName? Are there other characters not allowed?

halatmit:
Could you send a specific example of how to create this error?   App Inventor already has a check to remove ampersands in project names, but I guess we need more checks.  I'd like to see where they are missing.

fmntf:
Create a new project. In Designer, type (or copy/paste) in Screen1.AppName "foo & bar".
The ampersand is not removed.

weihuali0509:
I am looking into that.

weihuali0509:
I can confirm the symptom - build failed when the app name contains the *&*, ampersand. 

After several intense debugging session, I found out there is no problem in the implementation, but rather a bug in the android *aapt*.  An ampersand in the name string breaks AAPT and other places in Android. 

This bug is also described in this Java file, line 300
https://github.com/mit-cml/appinventor-sources/blob/master/appinventor/buildserver/src/com/google/appinventor/buildserver/Compiler.java

I am going to follow the same style to fix this problem - by replacing the *&* with the word *and*.

@halatmit

kkashi01:
You can use &amp; also

halatmit:
#464 was closed

PR #464, opened 2015-07-16
Fixed the Ampersand problem in the app name

weihuali0509:
This patches around a bug in AAPT (and other placed in Android) where an ampersand in the app name (android:label) string breaks AAPT.

This pull request addresses the problem in this issue - 
https://github.com/mit-cml/appinventor-sources/issues/456


halatmit:
@afmckinney Please review for Wei

afmckinney:
LGTM

kkashi01:
What happens if user enters &amp; The result will not be what user is expecting

halatmit:
Handles issue #456

Issue #468, opened 2015-07-18
Align button in arrangement problem


liyucun:
This problem is originally reported from issue#410, and there are some related discussion in developer group: https://groups.google.com/forum/#!topic/app-inventor-dev/wcw6iKvAJrQ

*if I have a horizontal arrangement, with only 2 buttons in it, who have fixed and equal height, they should be aligned so that their bottom and top are equal. But that is not the case, if they contain text with multiple lines, and not the same number of lines, as my picture shows. And I should not need to set a vertical alignment!*  -- from Ghica's description for the problem

Other than the problem in our app inventor, I tried to make the same view setting in Android, it turns out that Android also have this "problem". If I have a linearlayout that is horizontal orientation, with 2 buttons in it, who have fixed and equal height, it will turns out to be the same view like Ghica's description. Shows as below:

![image](https://cloud.githubusercontent.com/assets/4750970/8763498/05dcb154-2d6d-11e5-931b-098598852208.png)

halatmit:
Yucun,

Thank you for verifying that!  That sounds like a bug in linear layout.  I
wonder if it's been reported on the web. In any case, if that's a bug
there, we won't try to patch around it in App Inventor.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.
MIT Dept. of Elec. Eng. and Comp. Sci.
MIT Media Lab, Second Class Faculty

liyucun:
Hi @halatmit ,

It turns out that this problem is related to baselineAligned attribute of LinearLayout (http://developer.android.com/reference/android/widget/LinearLayout.html#attr_android:baselineAligned).
 
As below shows:
![image](https://cloud.githubusercontent.com/assets/4750970/8763820/a6f12222-2d7d-11e5-9f80-355af59172b3.png)
These two button above are aligned according to their text instead of button layout. By default, linearlayout's baselineAligned attr is TRUE, and hence we have this kind of alignment problem.

One possible and simple solution would be adding baselineAligned property in our App Inventor for LinearLayout.

I put a reference link here in case of my explanation not clear enough:
https://possiblemobile.com/2013/10/shifty-baseline-alignment/

halatmit:
Would it work to just set BaseLineAligned to false?   It would be good if
App Inentor users didn't have to deal with this.  Or is there a reason why
App Inventor users might  sometimes was it set to true?   If so then you
should go ahead and add the property -- but be sure to document it in a way
[...]

Issue #410, opened 2015-06-03
Alignment in arrangements is may have unnecessary restrictions


halatmit:
See this thread in the forum:

https://groups.google.com/forum/#!category-topic/mitappinventortest/JthYE6ZdxWU

Note that the problem is fixed by changing the HorizontalArrangement  to give is a fixed height and making the VerticalAlignment centered.  But should that be necessary?

Also, in general, if you have a horizontal arrangement whose height is Automatic, then you cannot change the vertical alignment.  (similarly for width/Horizontal).   Hal put this restriction in, but he doesn't remember why, or whether it is necessary.

liyucun:
The possible explanation i can find is that this feature might be used for guiding user to set layout alignment. That can avoid a user wondering why there is no changes after setting AlignVertical to centre when layout height is Automatic.

halatmit:
Adding ghica's comment:

Try the attached .aia
The height of the 2 buttons will become not equal, if they have different amount of text lines in them, also in the test version given here.
The workaround now is: Go to the design view, set the height of the Horizontal Arrangement to something fixed, then set AlignVertical to center. Then you can set the height back to automatic if you wish. After this, the buttons will look OK.
But the real fix should be something more logical I think. 
Cheers, Ghica.

halatmit:
Yucun,
Here is a picture.
![mail google com](https://cloud.githubusercontent.com/assets/789431/8686917/ba2aca64-2a58-11e5-9e46-5508fbebaee9.png)


It shows the screen after the top "try" button is pressed, and it shows that the buttons are not properly aligned.

​I also uploaded a new, modified .aia, to make it (I hope) easier to see what is wrong.
Try the "try" buttons from top to bottom. The first one will show the problem.
The last one will set Align Vertical to center. And that fixes the problem.
You cannot do this from the design view, because "center" is greyed out.

One other remark, I do not know whether you can fix that too: The documentation does not tell how to set the AlignVertical attribute. It does not accept "center", therefore I tried 2 and it works.
Cheers, Ghica.

Met vriendelijke groeten,
Ghica van Emde Boas
email: emdeboas@bronstee.com

halatmit:
adding Mark's comment:

So, it seems that there are some inconsistencies between App Inventor behavior and the documentation, as well as the UI inconsistency I pointed out earlier.

[...]

PR #474, opened 2015-07-23
remove align restriction, and set linearlayout baselineAlign to false…

liyucun:
This commit is for solving issue #410 and #468 .

The part for #410 is to remove restriction code for alignVertical/alignHorizontal property.

The part for #468  is to add method for setting baselineAlign property, and set it to false by default when initialize arrangement.

Other than this, there are some code to be done on designer screen side. But I create this pull request for testing for this change first, designer won't cause any problem if this one can be integrated to App Inventor.


halatmit:
@liyucun 
I looked at this quickly and it looks OK, but I'd rather no to a review until it's also fixed in the designer:  we don't want to merge something that's only half done.

liyucun:
Hi @halatmit 

I have fixed this in the designer now. The app inventor instance can be found at: spinner-issue.appspot.com

This enable designer screen show related changed when alignVertical property is Top/Center/Bottom in a automatic height/weight Arrangement.

halatmit:
LGTM
Really nice work!

jisqyv:
LGTM, Pushed to Gerrit for the UCR branch

jisqyv:
@liyucun @halatmit MERGED into UCR

PR #475, opened 2015-07-24
Fixed collapsed block text on renaming

shrutirij:
This fix creates a utility function for block types that can be renamed by the user (components, procedures and variables).

It recursively checks for collapsed parent blocks as these might also contain the old name of the renamed block. There is a constant maximum number of recursive calls. 

Examples and further details [here](https://docs.google.com/document/d/1FPpiiGsVax3muclKTaGGTWiKgGLRrMMptiI1fHMopqs/edit?usp=sharing).

Try it - http://ai2-collapsed-fix.appspot.com/

CC: @josmas

### DIFF
--- a/appinventor/blocklyeditor/src/blocks/utilities.js
+++ b/appinventor/blocklyeditor/src/blocks/utilities.js
@@ -75,6 +75,9 @@ Blockly.Blocks.Utilities.wrapSentence = function(str, len) {
   }
 };
 
+// Change the text of collapsed blocks on rename
+// Recurse to fix collapsed parents
+
 Blockly.Blocks.Utilities.MAX_COLLAPSE = 4;
 
 Blockly.Blocks.Utilities.renameCollapsed = function(block, n) {
--- a/appinventor/blocklyeditor/src/blocks/utilities.js
+++ b/appinventor/blocklyeditor/src/blocks/utilities.js
@@ -75,6 +75,26 @@ Blockly.Blocks.Utilities.wrapSentence = function(str, len) {
   }
 };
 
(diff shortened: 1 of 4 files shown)


halatmit:
@fturbak @josmas 
Lyn (or José) could you review this?

fturbak:
Yes, I can review this.

- lyn -

appinventor/blocklyeditor/src/blocks/utilities.js
```diff
@@ -75,6 +75,29 @@ Blockly.Blocks.Utilities.wrapSentence = function(str, len) {
   }
 };
 
+// Change the text of collapsed blocks on rename
+// Recurse to fix collapsed parents
+
+Blockly.Blocks.Utilities.MAX_COLLAPSE = 4;
```
jisqyv: @shrutirij Why the max of 4 and what happens when the recursion depth is greater?

shrutirij: @jisqyv 

The recursive check is used when parents of "renamed blocks" are collapsed. For example, 

capture2

The "if" block is collapsed, but uses the "get name" in its collapsed string (`_TEMP_COLLAPSED_INPUT`).

Since `_TEMP_COLLAPSED_INPUT` already has a fixed string length (30, I think) - I couldn't find instances where more than 2 recursive parents would contain the "renamed block". 4 seemed to be a safe number of parents to check.

jisqyv:
LGTM

jisqyv:
Squashed and MERGED

PR #477, opened 2015-07-25
Fix NPF in Player

jisqyv:
If a Player is used only for vibration and no source sound is set,
then an NPF occurs on application exit.

Change-Id: I3de012ec71a1a085b0e13bb818708150f05efbea


jisqyv:
@halatmit @afmckinney Tripped on this one while testing something else... Easy and obvious fix. Please review.

halatmit:
LGTM

jisqyv:
MERGED into ucr

PR #480, opened 2015-07-26
Fix error message to specify that input must be an integer, not just …

halatmit:
…a number


afmckinney:
squashed and pushed to gerrit to be merged. LGTM

PR #484, opened 2015-07-31
Update appengine/build.xml

lizlooney:
Delete old blockly-<guid>.js files before creating new one.

I noticed that everytime I do "ant" or "ant noplay", I end up with an additional blockly-<guid>.js file in the appengine/build/war directory. These files are 3.9M each.

### DIFF
--- a/appinventor/appengine/build.xml
+++ b/appinventor/appengine/build.xml
@@ -96,6 +96,9 @@
         <filter token="GUID" value="${guid1}"/>
       </filterset>
     </copy>
+    <delete>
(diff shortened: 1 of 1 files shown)


lizlooney:
@halatmit @jisqyv

jisqyv:
@lizlooney There was another place to get (which I took care of) so "ant Blockly" also gets rid of the old blocklies.

jisqyv:
LGTM

jisqyv:
@lizlooney MERGED.

Issue #481, opened 2015-07-29
"Export all projects" disabled in My projects view


halatmit:
I am exporting my students' work from earlier this month.

I've noticed that if I go to the projects listing from the top-level "My Projects" button, then the menu option to "Export all projects" is not available.

OTOH, if I go there by choosing Projects > My projects, then it's available.

I think this is a bug.

be1be1:
Dear Prof. Hal, has this issue be fixed already? I am trying to do it and I have one question. As a normal user, should I be allowed to export all projects when I am in the designer/block view?

halatmit:
@bxie 
Yes, I think you should be able to do it from the designer view.
I'll assign the issue to you.

PR #490, opened 2015-08-03
Enable "Export all Projects" in My projects view issue #481

be1be1:
Previously, "export all projects" item will be disabled if you go to the project lists from the top-level “My Projects” button. It is fixed now. Issue #481


afmckinney:
Pushed to gerrit and merged.

PR #491, opened 2015-08-03
Exportselectedprojects cleaned up

attraylor:
This branch adds a "Export selected projects" menu item to the Projects dropdown in the menu at the top. It downloads all projects the user selects (using the check boxes next to the projects) as a zip file.

I closed my last pull request and opened a new one because my last pull request a) had a bunch of extra additions, and b) was not squashed into one commit.

@afmckinney


afmckinney:
Works but needs to be changed slightly so that if there is one project selected in my projects page the menu items are properly deselected and selected. I would also propose that the menu item for downloading a project from the designer or blockseditor be rename to download current project and on My Projects page even if there is one project selected it should not give you and error when you select download selected projects as a zip file.

PR #508, opened 2015-08-15
Added supportsScreens to manifest

halatmit:
Added keyboard for ConfigChanges so app won't stop when attaching an external keyboard.


colinwirt:
Did ConfigChanges end up having keyboard added or was the change abandoned?

PR #523, opened 2015-08-20
Fix NPF in Player

jisqyv:
When playing MIDI files the player will fail to prepare after a
stop (Android Bug). This leaves the player pointer null, so we should
not dereference it.

Change-Id: Iff739b307eac325e608c729223651329ea1a9784


jisqyv:
@halatmit After this fix, you do get the error from prepare reflected to the app.

halatmit:
LGTM

halatmit:
Yes


== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

PR #525, opened 2015-08-21
Arrange for a “Send Error Report” in the Companion

jisqyv:
If ACRA is active, add a menu item, only on the Companion, offering to
send an error report. Once the report is sent, a two word report ID is
displayed to the user. We will ask people who submit reports to send
us an e-email (or update the forum) with the two word ID so we can
find their report in our database.

Change-Id: I9898b0077ce163bdbc6af9aae546fd833d9a8a6c


jisqyv:
@halatmit As promised. Here is the PR for adding the report sending...

jisqyv:
@halatmit If you simply compile App Inventor, ACRA is not active.  To activate it. You need to do an "ant clean" and then build the companion as:
 
  "ant -Dacra.uri=<URI> PlayApp"

Where <URI> is the url for the backend reporting. If you set it to http://acra.appinventor.mit.edu you should see the menu choice (though that backend may be down at the moment, but it won't matter, you can still attempt to submit a report and you will see the dialog with the ID). ACRA queues reports that it cannot send immediately.

halatmit:
>   Where is the url for the backend reporting

You mean

"ant -Dacra.uri= <uri for backend reporting>"  ???



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
@halatmit Yes, when I first entered the comment github swallowed the angle brackets. You probably want to use the "PlayApp" target so you don't build the client side just to build a companion.

For testing you can use http://acra.appinventor.mit.edu/acra/

That is the old acra reporting location we used with Lyn, it should still be up, but it is discarding information it receives.

appinventor/components/src/com/google/appinventor/components/runtime/ReplForm.java
```diff
@@ -145,6 +147,23 @@ public boolean onMenuItemClick(MenuItem item) {
...
+    if (!ReplApplication.isAcraActive()) { // If ACRA isn't active
+      return;                              // we don't show the button
+    }
+    MenuItem showSettingsItem = menu.add(Menu.NONE, Menu.NONE, 4,
+      "Send Error Report").setOnMenuItemClickListener(new OnMenuItemClickListener() {
+          @Override
+          public boolean onMenuItemClick(MenuItem item) {
+            String reportId = genReportId();
+            ReplApplication.reportError(null, reportId);
+            Notifier.oneButtonAlert(activeForm, "Your Report Id is:\n" + reportId, "Error Report Id", "OK");
```
halatmit: add another line:   use this id when reporting your error

appinventor/components/src/com/google/appinventor/components/runtime/ReplForm.java
```diff
@@ -145,6 +147,23 @@ public boolean onMenuItemClick(MenuItem item) {
...
+            return true;
+          }
+        });
+    showSettingsItem.setIcon(android.R.drawable.sym_def_app_icon);
```
halatmit: how about
android.R.drawable.stat_sys_warning

halatmit:
two tiny comments.
Otherwise LGTM, so push it.

jisqyv:
@halatmit I'll make the changes!

jisqyv:
MERGED

PR #526, opened 2015-08-22
Make sure the Web Component has the permissions it needs

jisqyv:
The Web Component needs READ_EXTERNAL_STORAGE and
WRITE_EXTERNAL_STORAGE. When we created packages with minSdk of 3, we
didn’t need these permissions because they were not introduced until
API 4, so if you declared you minSdk to be 3 (or less) then their
granting was implied. When we updated our minSdk to 4, we lost this
grandfathered implied permission.

Change-Id: I01bedb15bc82e58231bf1ae3226af788b7d86bb5


halatmit:
LGTM

Issue #406, opened 2015-06-03
Correct value of selection index with filtered list pickers


halatmit:
Below sample shows how filtering works. Select 1st a and then last a. On each selection, the selected text, index will display.

Next, toggle filter and click on same selections. Notice the selectionIndex. Taifun's recommendation is to keep / show the original unfiltered index. I can see the benefit in it, but not sure on how to accomplish it.

Try test app here: http://3nportal.com/temp/TestListView.aia

Momoumar:
@halatmit,
I was not able to reproduce this bug on a One Plus One running Android 5.0.2.

halatmit:
Are you saying that the selection index now refers to the position in the
unfiltered list?
If so, we can close the issue.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.
MIT Dept. of Elec. Eng. and Comp. Sci.
MIT Media Lab, Second Class Faculty

Momoumar:
Yes that is correct. The index are fine whether or not filter is toggled.

Momoumar:
Oh I think I've missedsomething. 
The bug is that the filtered text should show the index of the unfiltered one ?

halatmit:
The correct behavior is that the selection index should be the index in the
unfiltered list, not the index in the filtered list.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.
MIT Dept. of Elec. Eng. and Comp. Sci.
MIT Media Lab, Second Class Faculty

Momoumar:
Noted.

jisqyv:
Fixed in #695 now merged into UCR

PR #528, opened 2015-08-22
 Correct value of selection index with filtered list view. ref issue …

Momoumar:
…#406
@halatmit I didn't find another way to solve this.


halatmit:
I'm going to close this for now, until someone makes progress on the issue

PR #532, opened 2015-08-23
Fix MediaUtil for devices with density of 1.0

jisqyv:
Turns out that calling scaling with a density of 1.0 causes Android to
just return the original bitmap. But we called recycle() on the
original bitmap, which results in an error when the scaled
bitmap (which is “eq” to the original bitmap) is used.

Remove the call to recycle() as it likely isn’t needed. Don’t bother
attempting to scale a bitmap when the density is 1.0

Change-Id: Ib9b52baa29c99b2ccd77fbef299522446552ab60


jisqyv:
@halatmit Please review. I'll likely update the commit message and the comments in the code to be more explanatory. But this is the gist of the actual change.

halatmit:
LGTM

PR #534, opened 2015-08-24
Make sure the SoundRecorder has the permissions it needs

jisqyv:
It needs READ and WRITE External Storage now that minSdk is 4.

Change-Id: I1e3bd74449f154020b531b4815b9f3e8c8dfbcfa


jisqyv:
@halatmit Here is the SoundRecorder fix (tested).

-Jeff

halatmit:
LGTM

PR #539, opened 2015-08-27
Bugfixes in VideoPlayer

jisqyv:
Make sure we can use Percentage settings in the Video Player. Also
have to manually apply screen density as well.

Change-Id: I8ae8299f39af3321a1a6ffc57505fcb14fa14aeb


jisqyv:
@afmckinney @halatmit This code should fix the video player, both percentages and scaling.

afmckinney:
LTGM - Pushed to Gerrit

Issue #285, opened 2015-02-25
change HideKeyboard to be a single block under control


halatmit:
.. rather than a method of textboxes.

jisqyv:
@halatmit So we should deprecate the TextBox.HideKeyboard method (and convert the blocks to the new generic block on project upgrade?).

halatmit:
Yes, that's what we should do.




== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.
MIT Dept. of Elec. Eng. and Comp. Sci.
MIT Media Lab, Second Class Faculty


On Tue, Feb 24, 2015 at 11:38 PM, Jeffrey Schiller <notifications@github.com

EnisM:
Thanks Hal...

halatmit:
I think that we should do this in two stages:

(1) Add HideKeyboard as a screen method.
(2) Deprecate TextBox.HideKeyboard.

Let's do only (1) for now.   It doesn't hurt to keep the TextBox method around.

kkashi01:
@halatmit If you like, you can assign this to me

marksherman:
@halatmit, you've said it two different ways, and I'd like to clarify. Do you want the new "hide keyboard" block to be under the Screen menu, or under the built-in Control menu?

(question derived from talking with @farxinu)

halatmit:
@marksherman I think we should make it s method of Screen, not a new builtin under Control.

halatmit:
@kkashi01 Thanks for volunteering, but I'd like to assign this to Farzeen, since it's a good starter project.   Anything else you'd be interested in taking.

kkashi01:
@halatmit NP. I'll take a look at list of issues

josmas:
Can this be closed? I can see the block in `Screen` now. Is the deprecation of the TextBox block needed?

ewpatton:
@josmas Agreed. I don't think the deprecation is necessary. There may be many ways to solve the same problem. Some people might be inclined to look at TextBox as a way of hiding the keyboard, whereas others might look to the Screen. I think both ways are valid. (I'd also be willing to throw a block in control, but it's less clear to me that it should be there than something to do with screen management.)

PR #541, opened 2015-08-28
Moved HideKeyboard from TextBox to Form (Bug #285)

farxinu:
The method block has been moved from TextBox.java to Form.java, which increments the versions of YoungAndroid, Form, and TextBox.


farxinu:
@halatmit @marksherman

halatmit:
@afmckinney @farxinu @marksherman 
Andrew please review this.  Pne issue is that removing the HideKeyboard method from TextBox break existing projects.   It's better to keep the old method there (and also have it in Form), but arrange the upgrader so that the TextBox method is show as deprecated, in a tooltip or something.

afmckinney:
I am testing this now.

jisqyv:
@farxinu Sorry for the slow response, we have been busy on other issues.

A change like this has to provide a an upgrader to upgrade existing projects that are using the HideKeyboard function in TextBlock to use the block from Form (Screen1). As currently written any project that using the HideKeyboard block will break and become unusable.

jisqyv:
@farxinu Btw. If you leave the version in place in TextBox, then you don't have to do an upgrader. We believe that is a reasonable approach.

farxinu:
So I should re-add HideKeyboard as a TextBox method and also have it as a Form method?

farxinu:
@jisqyv sorry, forgot to tag you.

jisqyv:
@farxinu Yes. That will avoid the need for an upgrader.

farxinu:
@jisqyv Should be fixed now.

jisqyv:
@farxinu Note: When you add a new error number, you also have to add the message that goes with it. In this case I have done that for you:

     errorMessages.put(ERROR_SCREEN_INVALID_ANIMATION,
         "Bad value for screen open/close animation: %s");
+    errorMessages.put(ERROR_NO_FOCUSABLE_VIEW_FOUND,
+        "No Focusable View Found");
     // Canvas errors

(for future reference).

Otherwise: LGTM. Tested and on Gerrit for final unit tests.

halatmit:
What's the actual error message being shown by default as the notofication?




== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
@halatmit "No Focusable View Found" (Yeah, I know... I just merged this change, but we can add another change to change the message).

jisqyv:
MERGED

halatmit:
How about "No keyboard is showing".

Does that adequately describe the error condition?


== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.


On Sun, Oct 18, 2015 at 10:01 PM, Jeffrey Schiller <notifications@github.com

jisqyv:
Not really. The keyboard may well be showing...

halatmit:
Hmm.

OK.  Let's leave it as is and someday figure out something better to say.




== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.


On Sun, Oct 18, 2015 at 10:32 PM, Jeffrey Schiller <notifications@github.com

farxinu:
Thank you! Sorry, I didn't realize that.

PR #542, opened 2015-08-29
Give Canvas the ability to include additional gesture detectors for u…

halatmit:
…se with extensions.


halatmit:
@jisqyv @afmckinney 
This addition is for use by extension components, but let's merge it into the main branch now (and also merge it into the cdk branch).  There's no user-visible change, but it doe require an update to the Companion.

jisqyv:
LGTM

jisqyv:
MERGED

PR #548, opened 2015-09-01
Report Button

toropeza:
The report button for apps in the gallery should not be as big as the description and the share tabs. Reporting an app is not something that is as often used as looking at the description and sharing the app. I moved the report button to the bottom left of the UI across from the back to gallery page for a better UI. Also, the back to gallery button has some awkward padding and styling that I fixed to be the same as the report button.
You can take a look at the new report button on http://1.aigallerythomas.appspot.com/

Screenshots
--------------------------------------------------------------------------------
screenshot from 2015-09-01 12 29 49
--------------------------------------------------------------------------------
screenshot from 2015-09-01 12 30 40


josmas:
Hi Thomas, thanks for your pull request but it seems to be kinda messed up. You have committed your build folder and some other stuff there. There are 4,447 files changed in your request.
My gut feeling is that you did not copy the sample-gitignore file into .gitignore?

Have you discussed this change with anyone in the forums or directly?

I am going to close the pull request for now, only because of the gitignore issues. Feel free to open a new one when you sort out your local repo, but I would suggest discussing the changes in the forum if you haven't done so already and the change was agreed upon. Thanks!

cc @halatmit

toropeza:
Hey,

Yes that is exactly what happened. Just fixed and created a new pull request.

toropeza:
thanks!

PR #555, opened 2015-09-09
Work around a bug in screen resizing invoking the keyboard

jisqyv:
There is a bug where on some devices when the soft keyboard is closed
and we are in “Fixed” mode the screen is resized to be too tall,
cutting off the bottom of the App’s screen. This change does not fix
this problem, but provides a work around where we detect that the
keyboard is closed and recompute the layout of the app.

Change-Id: I31bb58e4fba874c9ea7c5b0a2bb739ebb9817989


jisqyv:
@halatmit @afmckinney Tested on your (Hal's) tablet and my phone. Changing orientation works as well.

afmckinney:
LGTM - Needs to be moved to Gerrit

jisqyv:
Removed the smart quotes. Jenkins doesn't like them...

PR #557, opened 2015-09-11
Set scaleFrameLayout scale to 1.0 when keyboard open

jisqyv:
Set scaling to 1.0 when the keyboard is open. This effectively puts us
in Responsive mode when the keyboard is open. Set it back to the
appropriate value (when in Fixed mode) when the keyboard is
closed. This works around the issue where the screen gets improperly
scaled when the keyboard is open. We suspect there is a bad
interaction between our scaling in Fixed mode and the scaling that
Android provides (at least in Android 5.1.1) when the keybaord is
opened. So we remove ours.

Change-Id: I1cec18ea57ead5527885d96c8b395ab90c240f5d

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/Form.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/Form.java
@@ -342,18 +342,19 @@ public class Form extends Activity
       Log.d(LOG_TAG, "keyboard hidden!");
       if (keyboardShown) {
         keyboardShown = false;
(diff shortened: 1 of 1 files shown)


jisqyv:
@afmckinney @halatmit Please review. This code fixes the scaling problem even when the keyboard is open (previous commit fixed it when the keyboard was closed, but the problem remained while the keyboard was open).

If happy, just mark with LGTM. I'll move it to gerrit and merge etc.

halatmit:
awesome
LGTM

PR #558, opened 2015-09-19
Support for options menu, ordered components on palette, renaming method/property blocks

lizlooney:
These are changes that I made for FTC, but I think would be useful on the master branch.

@jisqyv @halatmit


halatmit:
@lizlooney 
Liz,
Please add something -- maybe an extended comment -- that explains how the component ordering mechanism is intended to be used?     thanks

lizlooney:
I added another file: NxtPaletteHelper.java to show how to use the component ordering mechanism.

Please let me know if I'm re-inventing the wheel here. I could not find an existing way to order the components in the palette, but it's hard to believe we never needed/wanted that before.

halatmit:
This is not reinventing the wheel.  It's very useful.   Can you add an
extended comment to show developers how to use this when adding new
components?






== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

afmckinney:
Hi Liz, 
Don't worry about the comment I will add a section in our "How build a component" doc how to use this.

Thanks 

Andrew

afmckinney:
Looks good to me. Now merged into gerrit.

PR #569, opened 2015-10-01
Back Pack Bugfix: Fix Warning Handler

jisqyv:
Make sure generic component blocks are not flagged as errors.

Change-Id: Ieab969a2ca3d14bb6761935526eddf899a1276f2


jisqyv:
@halatmit @afmckinney Please Review. This fixes one of the problem we have found with the Back Pack. If reviewed (I just need an LGTM) we can put this out tonight or tomorrow morning.

jisqyv:
LGTM from Hal in #570 I've updated the comment

PR #571, opened 2015-10-02
Optimize eventparam computation

jisqyv:
Alter getEventParam to effectively cache results

Change-Id: I8ad7647b54a2b37b3d437157a45572badea365ef


jisqyv:
@halatmit I'm not sure this will solve the problems we are seeing, but it does remove significant overhead (which I added :-( ). If nothing else it doesn't hurt. I have verified that event parameters with spaces in their names in non-English languages still work properly with this change.

halatmit:
LGTM
I don't understand what's going on here, or why it should relate to the comapnion being connected,  But as far as I can see, it can't hurt to merge this.

PR #577, opened 2015-10-09
Rightclick

ram8647:
Adds a context menu to the backpack that is brought up by right-clicking the backpack.  The context menu has a single item "Backpack documentation" that when selected brings up an alert window with a short description of the Backpack and a link to its online documentation. 

@jisqyv @halatmit

### DIFF
--- a/appinventor/blocklyeditor/src/backpack.js
+++ b/appinventor/blocklyeditor/src/backpack.js
@@ -348,6 +348,29 @@ Blockly.Backpack.prototype.position_ = function() {
 };
 
 /**
+ * On right click, open alert and show documentation
+ */
+Blockly.Backpack.prototype.openBackpackDoc = function(e) {
+  var options = [];
+  var backpackDoc = {enabled : true};
+  backpackDoc.text = "Backpack documentation";
+  backpackDoc.callback = function() {
+  var str = "The Backpack is a copy/paste feature. It allows you to copy blocks from one project or screen " +
+   " and paste them into another project or screen. " +
+   " To copy, you can drag-and-drop blocks into the Backpack. To paste, click on the Backpack icon and " +
+   " drag-and-drop blocks into the workspace." +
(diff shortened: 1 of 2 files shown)


halatmit:
@jisqyv @ram8647 
Jeff, When you modify this, see if there's a way to keep the change out of core.

ram8647:
In this case, there's only 1 line added to core in blockly.js in the
onMouseDown_ function.   I think this is unavoidable because the core is
the only place where mouse clicks are handled.    It seems to me that
changes like this to the core could be avoided if it were possible to
override onMouseDown_ and similar functions outside of the core?

However, I think there's quite a bit of context-menu code in blockly.js
that could be moved outside the core.  I'm thinking of the
Blockly.showContextMenu_() function.    Was that originally originally a
Blockly function or was it added for App Inventor?

-- ralph

halatmit:
I don't know specifically about showContextMenu.    But the idea is to
start structuring our changes so that it will be easier to update App
Inventor to new blockly releases.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

ram8647:
What would be helpful is if Blockly core contained functions like
showContextMenu that could be overridden outside the core.   As it stands
now, even we moved all the showContextMenu() code out of the core, it would
still have to be called from the core.

-- ralph

halatmit:
We need to talk more with Neil about ideas like this.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

afmckinney:
This is refactored in Neils latest Blockly code. It is on my list of things I'm working on to get Blockly in sync with ours.

jisqyv:
@afmckinney Note: I am working on converting Ralph's Alerts into more friendly dialogs

jisqyv:
#580 Now includes this code. Closing this PR

Issue #562, opened 2015-09-22
Procedure call missing sockets in type blocking


halatmit:
I noticed that when I use type-blocking and I choose a procedure that I have defined, the call-procedure block that appears sometimes does not have the sockets for the arguments.
Here's some steps to reproduce:
1. From the Built-in Procedures drawer, drag out a to-procedure-do block. Change the name to proc_with_two_args and add two arguments: x and x2.
2. From the Built-in Procedures drawer, drag out a to-procedure-do block. Change the name to proc_with_no_args.
3. From the Screen1 drawer, drag out the when-Screen1.Initialize-do block.
4. With the selection still on the when-Screen1.Initialize-do block, type "proc_with_n" and enter. It should correctly place a call-proc_with_no_args block.
5. With the selection still on the call-proc_with_no_args block, type "proc_with_t" and enter. It places a call-proc_with_two_args block, but it has no sockets for the arguments.

![foo](https://cloud.githubusercontent.com/assets/789431/10018786/700aa05e-6107-11e5-85b2-8d999fb9ada3.png)

colinwirt:
I have also struggled with this problem for some time, but am normally very busy when I come across it and just work around it.

I'll take a look myself if I get some time.

halatmit:
Thanks.

Let us know if you have time.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.
MIT Dept. of Elec. Eng. and Comp. Sci.
MIT Media Lab, Second Class Faculty

josmas:
Please review #587

jisqyv:
Fixed in #587 about to be merged.

PR #591, opened 2015-10-22
fix for #562 procedures missing sockets in typeblocks

lizlooney:
Change-Id: Ib1799dc882b8be48d33d3a09c1ca263599afa93c


lizlooney:
@halatmit @jisqyv

PR #592, opened 2015-10-22
fix for #562 procedures missing sockets in typeblocks

lizlooney:
Change-Id: Ib1799dc882b8be48d33d3a09c1ca263599afa93c

@halatmit @jisqyv


josmas:
A bit puzzled about this PR. The code is on master already, if you rebase ftc it should just be cleanly merged in? @lizlooney @halatmit

PR #595, opened 2015-10-26
Several block changes to FTC components.

lizlooney:
Several block changes to  FTC components.

* Added blocks
    + FtcLinearOpMode
        + Time  
          The number of seconds this op mode had been running when the RunOpMode event was triggered.
    + FtcOpMode
        + Time  
          The number of seconds this op mode had been running when the Loop event was triggered.

* Renamed blocks
    + FtcDcMotor
        + <s>ChannelMode</s> Mode

@jisqyv 
@halatmit


jisqyv:
@lizlooney LGTM, Merged

PR #597, opened 2015-10-28
Added GyroscopeSensor component.

lizlooney:
Added GyroscopeSensor component.

Since appinventor/blocklyeditor/src/msg/ko_kr/_messages.js has not been merged into the ucr branch yet, we'll have to add LANG_COMPONENT_BLOCK_GYROSCOPESENSOR_HELPURL, LANG_COMPONENT_BLOCK_GYROSCOPESENSOR_PROPERTIES_HELPURL, LANG_COMPONENT_BLOCK_GYROSCOPESENSOR_EVENTS_HELPURL, and LANG_COMPONENT_BLOCK_GYROSCOPESENSOR_METHODS_HELPURL to it later.

@halatmit @jisqyv @josmas


josmas:
Hi @lizlooney this works nicely. What's the timestamp supposed to be used for? Is it to allow timed readings (not to get too many readings)? Could anything higher level than that be added instead?
Other than that question, LGTM.

lizlooney:
Thanks for the review.

The timestamp parameter is to be used to calculate the change in angles over time. For example, this is useful for balancing a segway-type robot.

If you look at the documentation for Android's SensorEvent values field, for TYPE_GYROSCOPE (go here http://developer.android.com/reference/android/hardware/SensorEvent.html#values and then scroll down to TYPE_GYROSCOPE), it says "Typically the output of the gyroscope is integrated over time to calculate a rotation describing the change of angles over the timestep" and then gives a code example.

I included the timestamp parameter in the event because:
+ Although the user could use a Clock component, it only has millisecond precision, where the timestamp has nanosecond precision.
+ Using the Clock component would not give the time that the gyroscope data was recorded.
+ The user doesn't have to use the timestamp parameter. They can just ignore it.

jisqyv:
@lizlooney MERGED onto the ucr branch

PR #598, opened 2015-10-29
JavaScript version of project merger tool

aesmaili:
@afmckinney


josmas:
Hi there, a few questions and comments, if you don't mind:

there seems to be a mix of licenses, but not the one used by the project currently; is this by design or am I looking at it wrong?

Indentation seems to vary greatly from file to file. App Inventor uses 2 spaces for new lines, and 4 spaces for continuation.

The paths for many of the files look a bit weird, for instance the part in bold seems like a temp file?: /ai2merger_javascript/src/**Stuk-jszip-3a39a71**/Gruntfile.js

It also looks like that is a library (jszip?) but both that and jquery seem to be in the same folder as the rest of the source, which makes reviewing the code really difficult (one has to keep checking which code is yours and which code is not).

Digging a bit deeper, it seems that it's just two JS files linked from the html file (other that libraries) but the review has 126 changed files.

It would be great if the code in the files was separated into smaller units, and some kind of building system was used instead. Something like grunt, bower, and so forth, so you can also pull in the dependencies instead of committing them with the code.

aesmaili:
Hi!

I will fix the indentation. 

Stuk-jszip-3a39a71 is a folder with the JSZip library files (most of the 126 files are from in there). I kept the original package to preserve the content incase there’s a reference I’ve used or for future development. What I will do is put all of the files not written by me into a folder called “lib” to make it clear which code is mine. 

As for the license, I’ve used JSZip and FileSaver which each have their own, but I’m not sure if that answers your question or not...

jisqyv:
Superceded by #607

PR #599, opened 2015-10-29
Add Support for Multiple Dex Files

jisqyv:
Add support for multiple classes.dex files in a packaged app. This
overcomes the limit of 65536 methods that otherwise would come into
play.

All normal App Inventor component code and up to 12 additional support
libraries are bundled into the primary classes.dex file. Additional
libraries are placed in classes2.dex. Android API 20 and higher
automatically utilizes multiple Dex files. Code from Google is
included here to support manually adding classes2.dex in order
versions (back to API 4, Tested to API 8).

The Companion is special cased, because it includes every possible
library, it takes a long time to start the first time it is run after
installation while the MultiDex code parses and optimizes the
classes2.dex file. In the Companion we do this optimization in a
background thread and display a Progress Dialog.

Change-Id: I5f05841f09cd03e96c421685f65098464aba1053

### DIFF
--- a/appinventor/buildserver/src/com/google/appinventor/buildserver/Compiler.java
+++ b/appinventor/buildserver/src/com/google/appinventor/buildserver/Compiler.java
@@ -180,6 +180,7 @@ public final class Compiler {
   private Set<String> assetsNeeded; // Set of component assets
   private File libsDir; // The directory that will contain any native libraries for packaging
   private String dexCacheDir;
+  private boolean hasSecondDex = false; // True if classes2.dex should be added to the APK
 
   /*
    * Generate the set of Android permissions needed by this project.
@@ -381,6 +382,8 @@ public final class Compiler {
       out.write("android:icon=\"@drawable/ya\" ");
       if (isForCompanion) {              // This is to hook into ACRA
         out.write("android:name=\"com.google.appinventor.components.runtime.ReplApplication\" ");
+      } else {
+        out.write("android:name=\"com.google.appinventor.components.runtime.multidex.MultiDexApplication\" ");
       }
       out.write(">\n");
 
@@ -596,8 +599,8 @@ public final class Compiler {
     // are copied into temporary storage) and processed via a hacked up version of
     // Android SDK's Dex Ant task
     File tmpDir = createDirectory(buildDir, "tmp");
-    String dexedClasses = tmpDir.getAbsolutePath() + File.separator + "classes.dex";
-    if (!compiler.runDx(classesDir, dexedClasses)) {
+    String dexedClassesDir = tmpDir.getAbsolutePath();
+    if (!compiler.runDx(classesDir, dexedClassesDir, false)) {
       return false;
     }
     setProgress(85);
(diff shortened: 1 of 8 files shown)


jisqyv:
@afmckinney @halatmit Ready to review. I believe it has been run on enough devices...

Note: I am putting this in for master instead of ucr because we are not updating YaVersion. Of course the benefit of this code will not be seen until we next update the Companion and buildservers.

appinventor/buildserver/src/com/google/appinventor/buildserver/Compiler.java
```diff
@@ -985,22 +993,43 @@ private boolean prepareApplicationIcon(File outputPngFile) {
...
+    // worked to put all libraries into the first classes.dex file when we
+    // had 16 libraries and broke at 17. So this is a conservative number
+    // to try.
+    if (!secondTry) {           // First time through, try base + 12 libraries
+      if (offset > 12)
+        offset = 12;
+    } else {
+      offset = 0;               // Add NO libraries the second time through!
+    }
+    for (int i = 0; i < offset; i++) {
```
halatmit: I don't understand what's happening on this second try.  Is it that the first try is handling 12 libraries and if that fails, we  allow more?  Could you add a comment to explain this better?

appinventor/buildserver/src/com/google/appinventor/buildserver/Compiler.java
```diff
@@ -1016,7 +1045,24 @@ private boolean runDx(File classesDir, String dexedClasses) {
...
-      setProgress(75);
+      if (dxSuccess && (class2List.size() > 0)) {
+        setProgress(60);
+        dexTask.setOutput(dexedClassesDir + File.separator + "classes2.dex");
+        inputList = new ArrayList<File>();
+        dxSuccess = dexTask.execute(class2List);
+        setProgress(75);
+        hasSecondDex = true;
+      } else if (!dxSuccess) {  // The initial dx blew out, try more conservative
+        LOG.info("DX execution failed, trying with fewer libraries.");
```
halatmit: I don't get what "fewer libraries" means here.   I guess it's related to my comment above.

appinventor/components/src/com/google/appinventor/components/runtime/Form.java
```diff
@@ -225,6 +247,51 @@ public void onCreate(Bundle icicle) {
     viewLayout = new LinearLayout(this, ComponentConstants.LAYOUT_ORIENTATION_VERTICAL);
     alignmentSetter = new AlignmentUtil(viewLayout);
 
+    progress = null;
+    if (!_initialized && formName.equals("Screen1")) {
```
halatmit: Are we setting ourselves up to lose in the future, if we try to generalize things so that the initial screen doesn't need to be named Screen1?   Can we add a warning comment in the code somewhere ?

halatmit:
@jisqyv @afmckinney 
no changes, but a couple of requests for comments

jisqyv:
@halatmit I believe I have addressed the comments.

-Jeff

halatmit:
LGTM

jisqyv:
MERGED

PR #600, opened 2015-10-30
Fix Memory Leaks

jisqyv:
Fix memory leaks in the Companion

Change-Id: I254b475abdc412c0df1213dc886dd771a20db41e


jisqyv:
@afmckinney @halatmit Here is the code that fixes the largest memory leak we have. I believe there are still some smaller leaks, which I will pursue when I have time.

If you want to test it out, Here is a link to a Companion built with this code + the multidex code. You can use it against ai2.appinventor.mit.edu.

https://www.dropbox.com/s/kumig6i8c5ljh5u/MITAI2Companion.apk?dl=0

halatmit:
LGTM

jisqyv:
MERGED

Issue #507, opened 2015-08-15
Companion fails to "connect" to a YA app with error!


thequixotic:
Connecting Companion to a Project with errors in it will cause a crippled connection.
The problem is only there during connecting. The companion will be connected (technically) to the AI client.
But the companion will fail to load anything.

For example try connecting to a project that has blocks as shown below
![start_companion_error](https://cloud.githubusercontent.com/assets/4526473/9289872/047af4e4-439b-11e5-8a82-07c750a9b887.png)

It will fail to load the Companion. Even after you try to correct the error above, it still won't show any updates.You have to try reconnecting the Companion after correcting the error.
However once connected, you can have errors later(it just wont show an app with error ). You can see changes immediately after correcting all error.

No, I do NOT have an elegant solution to fix this issue. 
However I presume, that the ReplForm fails to get loaded by runtime.scm due to an entire fail to the first scheme.eval .. 

@halatmit, is this bug important to fix.

halatmit:
@jisqv PTAL

jisqyv:
Ack. I'll look at this tonight or tomorrow.

jisqyv:
@thequixotic @halatmit So I figured this one out. The problem is that during the initial load of the app there is a (begin expr ...) where each expr is one of the scheme expressions that sets up a component. An error in an early expr stop execution so later components do not get setup.

The work around is to fix the problem in the project and then make a change in the designer (any change). A change in the designer triggers an App reload which will reload all components. If there is not error this time, all will proceed normally.

I am working on a fix, which involves modifying (do-after-form-creation ...) in runtime.scm. The idea being to wrap a try-catch around each expr being evaluated and continue after an error *and* signal the error back to the blocks editor. It may or may not be possible to indicate the component with the error (still looking at that) but at least the user will be alerted to the problem.

I'll hack on this either tomorrow or next week, depending on what else is demanding my attention :-)

jisqyv:
@thequixotic @halatmit P.S. I may want to defer the final fix until after we merge the changes for "watch" which also hacks a lot of the code in that neighborhood.

jisqyv:
This was fixed with #604

PR #604, opened 2015-11-06
Properly Report Errors during initial form load

jisqyv:
When we load the initial form, we tend to issue many Yail evaluation
requests in a row. Low level code in replmgr.js then consolidates
these changes into one XHR call instead of sending each one in its own
request. This improves form loading performance significantly.

However if an error occurs during code that is consolidated, we cannot
determine where the error occurred. Also, because the code is
consolidated into one (begin ...) block, an error in the middle causes
later s-expressions to not be evaluated. This leaves the Companion in
an undefined and usually broken state.

This code change detects an error when s-expressions are consolidated
and arranges to resend the form (reset the screen) without the
optimization in place. This results in the older “slow” code being
used, but better handles errors in the consolidated code. After 20
successful calls sending s-expressions over, the optimized code is
re-enabled.

Change-Id: Ic29ecdca7d3842f91829169842802419dd1a9fcc

### DIFF
--- a/appinventor/blocklyeditor/src/replmgr.js
+++ b/appinventor/blocklyeditor/src/replmgr.js
@@ -29,7 +29,21 @@ goog.require('goog.crypt.Sha1');
 goog.require('goog.crypt.Hmac');
 goog.require('goog.crypt.base64');
 
-top.loadAll = true;             // Global for debugging!
+top.loadAll = true;             // Use "Chunked" loading for initial form load
+                                // or any case where we have more then one "chunk"
+                                // of code queued for the device. We put this variable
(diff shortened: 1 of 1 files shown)


jisqyv:
@halatmit @afmckinney @thequixotic Please Review.

This change addresses #507

halatmit:
@jisqyv This looks pretty awesome?  Is there a good way to test it?

jisqyv:
@halatmit I'll give you a test case in a few. I'm actually going to update this PR in a few as well. I missed a case...

jisqyv:
@halatmit Look at the example in the issue (#507). Try it both before and after this change.

PR #605, opened 2015-11-08
Make blocks marked as bad persistent

jisqyv:
Fix a bug where a reload of a project with bad blocks would “unbad”
those blocks. Make sure we store whether or not a block is bad in the
blocks file.

Also, some minor code cleanup (deleting redundant code).

Change-Id: I0300554ec9a1c8805e48aa579c8c24f3960f08ce

### DIFF
--- a/appinventor/lib/blockly/src/core/xml.js
+++ b/appinventor/lib/blockly/src/core/xml.js
@@ -124,6 +124,9 @@ Blockly.Xml.blockToDom_ = function(block) {
   if (block.isCollapsed()) {
     element.setAttribute('collapsed', true);
   }
+  if (block.isBadBlock()) {
+    element.setAttribute('badblock', true);
+  }
   if (block.disabled) {
     element.setAttribute('disabled', true);
   }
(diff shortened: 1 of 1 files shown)


jisqyv:
@halatmit @afmckinney Please review

Yes, this is a change to core. The whole bad block implementation is in Blockly core. This fixes a bug where the fact that a block is marked bad was not saved in the project's blocks file, so when you loaded the project any bad blocks were no longer marked as bad.

appinventor/lib/blockly/src/core/xml.js
```diff
@@ -347,27 +350,6 @@ Blockly.Xml.domToBlockInner = function(workspace, xmlBlock, opt_reuseBlock) {
     block.initSvg();
   }
 
-  var inline = xmlBlock.getAttribute('inline');
```
halatmit: Is this code in Neil's blockly core?

jisqyv: @halatmit This code was in the original blockly (1752) that we are using (it may have since changed, but I haven't checked). Lyn moved it further down (see his comment on/near line 459). However he didn't delete it from its original position, which is an error.

jisqyv: @halatmit I just checked the latest blockly, and all of this code is refactored. Not sure if that should change our action here. We *have* bad blocks, blockly doesn't. So we will have to re-integrate them into the latest blockly when we merge it back. The Good News is it shouldn't be hard.

halatmit:
OK.   
I wasn't asking about the bad blocks as much and the other blockly attributes like inline.  But if the duplicated code was our errror, it's good do remove it.
LGTM

PR #610, opened 2015-11-14
updated Korean strings. (on 151114)

pillarlee16:
There were some mistakes at initial patch for the Korean translation.

1) French strings were included at the patch.
When I made a properties file from po file, I used a template based on French to convert it.
some missing strings were remained as French, so it was released on the running server.
I changed all French strings as Korean.

2) Added translations for missing strings, and fixed weird translations.
Now, I'm working with a publishing company about IT books.
They are reviewing strings based on the running server.
I would keep doing this until this year to improve the quality of Korean version.


jisqyv:
LGTM

jisqyv:
@pillarlee16 MERGED

PR #613, opened 2015-11-18
[ko_KR] fixed mistyping. 3bytes -> 4bytes

pillarlee16:
There was a tiny mistake.

The original string in English is '4Byte'.
But it was mistyped as '3바이트(Byte)'

So... it is fixed.


halatmit:
LGTM

jisqyv:
@pillarlee16 MERGED

PR #615, opened 2015-11-21
Fix the English for the ‘min’ and ‘max’ block tooltips

jisqyv:
Change-Id: I3eeb96722712d5e45b9ac4f352d839a6fccf6a30

### DIFF
--- a/appinventor/blocklyeditor/src/msg/en/_messages.js
+++ b/appinventor/blocklyeditor/src/msg/en/_messages.js
@@ -453,8 +453,6 @@ Blockly.Msg.en.switch_language_to_english = {
      Blockly.Msg.LANG_MATH_ONLIST_OPERATOR_STD_DEV = 'standard deviation';
      Blockly.Msg.LANG_MATH_ONLIST_OPERATOR_RANDOM = 'random item';
      Blockly.Msg.LANG_MATH_ONLIST_TOOLTIP_SUM = 'Return the sum of all the numbers in the list.';
-     Blockly.Msg.LANG_MATH_ONLIST_TOOLTIP_MIN = 'Return the smallest of its arguments..';
-     Blockly.Msg.LANG_MATH_ONLIST_TOOLTIP_MAX = 'Return the largest of its arguments..';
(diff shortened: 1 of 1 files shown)


jisqyv:
@halatmit @afmckinney Please review...

halatmit:
LGTM but I don't understand why was wrong or why that fixed it.

jisqyv:
@halatmit If you look at the whole you will see that there are a block of messages commented out. That block included these strings. I just moved them out of the commented out block. I did a spot check of the other strings that are commented out and they appear not to be used.

I'll do a more thorough check when I'm more awake.

halatmit:
But why Russian?



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
@halatmit It's a side effect of how the strings are initialized. There is a "switch_language_to_..." for each language. When the javascript files defining them are read in, we apparently switch to each language in turn. Once this is complete, we switch to the actual language we will use. So apparently Russian is switched to last *and* has a definition for the MIN and MAX blocks. When we switched to English at the end, because it didn't have a definition, the Russian one prevailed. This is a bit broken because it means that if any language doesn't define a given string, the string will be left defined to the last language loaded that did define it.

I'll see if we can come up with a better approach!

-Jeff

halatmit:
Thanks.   Who would have dreamed.




== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

josmas:
the fix LGTM but the more I hear/learn about the internationalisation code, the scariest it sounds :(
The TODO at the top of the commented block is also quite scary. I can do a bit of clean up after you merge this file (I can see with Intellij if the variables are *orphan*).

halatmit:
That would be great if you could do that.  I think Jeff has some ideas on
fixing things,


== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

josmas:
Please look at #617 if this hasn't been merged yet

jisqyv:
Superceded by #617

PR #620, opened 2015-11-25
Only Attempt to set the AppName property of MockForm

jisqyv:
This eliminates the annoying IllegalArgumentExceptions that litter the
OdeLog :-)

Change-Id: Ib7a0ab5f0dce578cf13d5020ee4840a9cdc26bfe


jisqyv:
@halatmit @afmckinney Please Review

We've been seeing these warnings in the logs since the AppName property was added. This fixes it.

PR #625, opened 2015-12-03
Fixed the search to allow for pagination through all the search results

toropeza:
A working version of this code can be found at:
http://searchfix.aigallerythomas.appspot.com/
note: In order to test the search on this site, search for the keyword "search". It is a keyword that is in most of the app's descriptions.

Changes made to the code
------------------------------------
*Reset the search counter when a new search is performed. This is done in GalleryList.refreshApps() if the refreshable variable is true, meaning that a new search is performed.

*Cursor is added to the query in order to page through the database. A new Cursor is created if the search is starting at index 0 or it is null.

*The total result number accuracy on a search is <=100. Meaning that unless the total number of results from the search is <= 100, it is just an estimation. This is because no one would be counting the accuracy of the results passed 100. 

*Set the limit on the search to 10. The reason for this is because the cursor points to the app after the last one returned from the search and the search originally queried 20 apps (by default since there was no limitation) but only 10 were shown (because this was the constant number of apps to show), then the next search (if using a cursor) would point to the 21st app skipping the previous 10.

*No longer skipping apps with a counter variable because the cursor will return the right results that allows the SearchIndex to just iterate over them

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/GalleryList.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/GalleryList.java
@@ -69,7 +69,6 @@ public class GalleryList extends Composite implements GalleryRequestListener {
   private GalleryAppTab appSearchTab;
   private GalleryAppTab appTutorialTab;
 
-
   public static final int REQUEST_FEATURED = 1;
   public static final int REQUEST_RECENT = 2;
   public static final int REQUEST_SEARCH = 3;
@@ -87,6 +86,7 @@ public class GalleryList extends Composite implements GalleryRequestListener {
   private int appPopularCounter = 0;
   private int appSearchCounter = 0;
   private int appTutorialCounter = 0;
+
   private boolean appRecentExhausted = false;
   private boolean appFeaturedExhausted = false;
   private boolean appPopularExhausted = false;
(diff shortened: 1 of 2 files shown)


toropeza:
@wolberd @jisqyv

appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/GalleryList.java
```diff
@@ -427,6 +427,8 @@ private void refreshApps(GalleryAppListResult appsResult, int requestId, boolean
         }else{
           appSearchTab.getNoResultsFound().setVisible(false);
         }
+        if (refreshable){appSearchCounter = 0;}
```
jisqyv: @toropeza 

Can you reformat this as

if (refreshable) {
   appSearchCounter = 0;
}

(using the correct indentation level)

appinventor/appengine/src/com/google/appinventor/server/GallerySearchIndex.java
```diff
@@ -5,30 +5,17 @@
...
-import com.google.appengine.api.search.IndexSpec;
-import com.google.appengine.api.search.PutException;
-import com.google.appengine.api.search.Query;
-import com.google.appengine.api.search.QueryOptions;
-import com.google.appengine.api.search.Results;
-import com.google.appengine.api.search.ScoredDocument;
-import com.google.appengine.api.search.SearchException;
-import com.google.appengine.api.search.SearchServiceFactory;
-import com.google.appengine.api.search.StatusCode;
+import com.google.appengine.api.search.*;
```
jisqyv: @toropeza 

Bad: We don't like imports of the form .*; I suspect your IDE did this. Please revert this change and alter the settings of your IDE to not do this.

toropeza:
@jisqyv @wolberd Requested changes are made

jisqyv:
LGTM

jisqyv:
@toropeza MERGED

PR #627, opened 2015-12-04
Permit isGCS in FileData to be null

jisqyv:
Change the definition of FileData.isGCS from “boolean” (lower case
boolean) to a wrapped “Boolean” (upper case) because the later can
take on a null value. Older projects may not have a value stored for
isGCS, which means its value is effectively null. A null value cannot
be loaded into a FileData object is isGCS is a boolean because
booleans cannot be null. This results in an error when loading older
projects.

Change-Id: I653fef7f3141c0febbec837929d28e4a00b256ff


jisqyv:
@halatmit @afmckinney Please review. I am still testing this myself, but so far so good. This problem is causing old projects and some modified during February of 2015 to fail to load.

jisqyv:
I'm going to open a fresh one...

PR #629, opened 2015-12-07
Reduce the number of Datastore reads

jisqyv:
Datastore reads are a significant cost of running MIT App
Inventor. To reduce the number of datastore reads we need to remove
calls that are of dubious value as well as use caching as much as
possible. The Objectify package which we use to interact with the
datastore implements both caching and transactions. However when we
use transactions, caching is disabled.

This change removes a very expensive access check and replaces it with
a much lighter weight version. It also reduces the use of transactions
where they add no value, which enables caching, which is of much
value.

Change-Id: Id215b15cd39579fea7854beaefd7c3594c161305


jisqyv:
@halatmit @afmckinney I believe this is ready to go...

afmckinney:
Tested ... LGTM ready to be moved to gerrit and merged.

PR #631, opened 2015-12-08
Fix Gallery "Likes" Feature

jisqyv:
This change involves two commits (which will *not* be merged together
as they represent two distinct, although related, changes).

1. Fix Gallery Like Data Management

Make sure a person can only like an App once. Enforce this on the
server side. Also add code to the Like Salvager to fix things when a
person has managed to like something more then once.

2. Remove the “salvageAll” feature from the Gallery

The “salvageAll” feature, available only to Gallery moderators was
useful back when the Gallery only had a few apps. However there are
too many apps today for it to be of much use. It would just use a lot
of resources and then fail. Because the salvageAll link is very near
the “salvage” and other related links, there is a risk of accidentally
invoking it. So we remove it here.

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/GalleryPage.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/GalleryPage.java
@@ -601,8 +601,6 @@ panel
     initTutorialSection(container);
     // Adds dynamic salvage
     initSalvageSection(container);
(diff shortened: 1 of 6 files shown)


jisqyv:
@halatmit @afmckinney @wolberd Please Review. Dave: You should be able to do this review.

wolberd:
Hi Jeff. Sorry I didn't get to reviewing it, looks like you merged which
his awesome.

Dave

PR #635, opened 2015-12-15
Don't export yail -- other changes

jisqyv:
This PR contains two commits (which will be kept separate). The first is to inhibit the exporting of project Yail except when doing a build or when the user is an admin. If necessary we can add a new menu item under the "Project" Menu to export a project along with its yail. That isn't included here and I'm not quite sure what to call it!

These commits are a pre-requisite for the FirebaseDB component addition, which I'll submit after one more minor change (and after these changes are merged). We are getting very close on FirebaseDB.

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/widgets/properties/EditableProperty.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/widgets/properties/EditableProperty.java
@@ -49,7 +49,7 @@ public final class EditableProperty extends Property {
   private final EditableProperties properties;
 
   // Type of property
-  private final int type;
+  private int type;
 
   // Property editor for use in properties panel
   private final PropertyEditor editor;
@@ -157,4 +157,13 @@ public final class EditableProperty extends Property {
   final void orphan() {
     editor.orphan();
   }
+
+  public int getType() {
+    return type;
+  }
+
+  public void setType(int aType) {
+    this.type = aType;
+  }
+
 }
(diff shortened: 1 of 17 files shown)


jisqyv:
OK, I lied. Its really three commits.

jisqyv:
@afmckinney @halatmit Please review

appinventor/appengine/src/com/google/appinventor/client/widgets/properties/EditableProperty.java
```diff
@@ -33,11 +33,23 @@
    */
   public static final int TYPE_INVISIBLE = 0x02;
 
+  /**
+   * Properties of this type will be included in a yail generation
+   * even if they are not persisted. This is used for computed properties
+   * which aren't stored, but which must be passed to a component both
+   * in the Companion and went built. Example: The FirebaseToken and
```
halatmit: went -> when

halatmit:
Tiny typo in a comment.
I looked over the code but I did not test this in hairy cases because I don't know what the hairy cases are.  We should put this out for testing, and be alert for things that seem funny.
LGTM

jisqyv:
@halatmit I'll fix the typo. This code is now on firebasetest.appinventor.mit.edu. I'm going to make another change to the code there tomorrow and then put out a call to the power users for another round of testing.

I'm pretty confident in this particular change. I did find some scary stuff in our code while testing this change. I didn't fix/modify the scary stuff because the fix would require a *lot* of testing, much more then this change needs (and the problem isn't serious, just ugly).

halatmit:
LGTM

jisqyv:
MERGED

PR #643, opened 2015-12-28
Added ms descriptor to the MinimumInterval property

ajcolter:


ajcolter:
Accidentally opened PR...

Issue #606, opened 2015-11-09
Location Sensor Location Changed Reports 0,0


ram8647:
This is an issue that's been reported by at least a dozen people in the edX course.  Apparently, on some devices, the LocationSensor.LocationChanged block reports the lat,long as 0,0.   I haven't noticed this happening on any of my devices but it shouldn't be possible.  

Looking at the Texting source code, latitude and longitude are initialized to **UNKNOWN_VALUE**, which is defined as 0.   The **onLocationChanged()** method (below) sets longitude and latitude without checking for UNKNOWN_VALUE. Apparently, the Android *onLocationChanged()* method can be called before the device has gotten a fix on its latitude and longitude.   Android documentation is not clear about this. 

And note that *onLocationChanged()* (below) also sets **hasLocationData** to true.  This variable is used elsewhere in the code to set the  **HasLongitudeLatitude** property to true.  If lat,long is 0,0, the sensor should not be reporting that it HasLongitudeLatitude and should not be firing the LocationChanged block.    

```javascript
public void onLocationChanged(Location location) {
      lastLocation = location;
      longitude = location.getLongitude();
      latitude = location.getLatitude();
      // If the current location doesn't have altitude information, the prior
      // altitude reading is retained.
      if (location.hasAltitude()) {
        hasAltitude = true;
        altitude = location.getAltitude();
      }
      hasLocationData = true;
      LocationChanged(latitude, longitude, altitude);
    }
```
I think this bug can be fixed by adding the following lines to *onLocationChanged*:

```javascript
      if (longitude != UNKNOWN_VALUE && latitude != UNKNOWN_VALUE) {
         hasLocationData = true;
         LocationChanged(latitude, longitude, altitude);
      }
```
In other words, **LocationChanged** will not be fired when lat,long = 0,0.  If we want to allow 0,0 as a legitimate location -- it is a valid lat,long -- then we should use some some value other than 0 as UNKNOWN_VALUE (e.g., something outside the range 0..360).

ram8647:
Doing some additional research on this issue.  

Apparently, by default, [Location.latitude and Location.longitude are set to 0,0](http://developer.android.com/reference/android/location/Location.html#Location(java.lang.String)) when a Location object is created.   

And according to this thread on [Stackoverflow](http://stackoverflow.com/questions/32650872/latitude-and-longitude-values-are-giving-null-values-0-0),  

[...]

PR #646, opened 2015-12-29
Fixes issue #606 -- prevents LocationChanged from returning lat,long …

ram8647:
…of 0,0.

Change-Id: Id143d8572972b6cb45ef08dd0e011ff247e5032c

An instance containing this change is available at https://trinity-appinventor.appspot.com.   A test app and source code are available here:  http://appinventor.cs.trincoll.edu/csp/gpstest/LocationSensorTester.apk and http://appinventor.cs.trincoll.edu/csp/gpstest/LocationSensorTester.aia.

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/LocationSensor.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/LocationSensor.java
@@ -87,7 +87,7 @@ public class LocationSensor extends AndroidNonvisibleComponent
 
       // By default Location.latitude == Location.longitude == 0.
       // So we want to ignore that case rather than generating a changed event.
-      if (longitude != UNKNOWN_VALUE && latitude != UNKNOWN_VALUE) {
+      if (longitude != UNKNOWN_VALUE || latitude != UNKNOWN_VALUE) {
 	hasLocationData = true;
 	LocationChanged(latitude, longitude, altitude);
       }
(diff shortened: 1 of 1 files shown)


appinventor/components/src/com/google/appinventor/components/runtime/LocationSensor.java
```diff
@@ -84,8 +84,13 @@ public void onLocationChanged(Location location) {
         hasAltitude = true;
         altitude = location.getAltitude();
       }
-      hasLocationData = true;
-      LocationChanged(latitude, longitude, altitude);
+
+      // By default Location.latitude == Location.longitude == 0.
+      // So we want to ignore that case rather than generating a changed event.
+      if (longitude != UNKNOWN_VALUE && latitude != UNKNOWN_VALUE) {
```
halatmit: should this be OR, not AND?
what happens if the latitude is 0 and the longitude is non-zero?

ram8647: Yes, I think you are right -- i.e., as long as either lat or long is
non-zero it's a valid location.  Good catch.    I'll revise and resubmit.

Thanks,
-- ralph

ram8647:
Hal's right.  An App Inventor location update event should not be triggered with both lat and long are zero.   But if either one is non-zero, that is a valid location and should trigger an event.   So the logic should be:

if (lat != 0 || long != 0)
  trigger a locationchanged event.

halatmit:
LGTM

jisqyv:
On Gerrit for merging into UCR

jisqyv:
MERGED into UCR

PR #647, opened 2015-12-29
Fix the spelling of "obsfucate" to obfuscate

halatmit:


jisqyv:
@halatmit MERGED onto UCR

Issue #645, opened 2015-12-29
add a way to rotate images


halatmit:
Currently, image sprites can be rotated, but not images.   add a way to rotate images.

Plan is to do this by adding a new image property called RotationAngle.

halatmit:
PR #648

halatmit:
PR #659 was closed

PR #648, opened 2015-12-30
Rotate image

halatmit:


halatmit:
Issue #645

jisqyv:
@halatmit ImageView.setRotation requires API 11 (Honeycomb). I was able to build a Companion with your code and it loaded in the Emulator (API 8), but generated an error at runtime. I'm not sure why it didn't get a linkage error when loading though. That said, a conservative approach might be to move the call to setRotation to HoneycombUtil and call it there from Image.java based on an appropriate SdkLevel test.

halatmit:
Weird.  I thought I'd tested that on Gingerbread but I guess I blew it.

I agree with your plan.   Do you want to do it, or should I?




== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
If you do it I can review it. Or I can do it and you can review it. The
interesting question is what to do on older devices.
-Jeff

halatmit:
Can you do it?  I'm in Hong Kong jail.

I suggest on older devices to just give a notice that it cannot be done.
I think full screen on video shows a similar notice.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
@halatmit I'll take care of it then...

halatmit:
Thanks.


== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
Superceded by #659

PR #660, opened 2016-01-10
Adds the “Firebase” Component

jisqyv:
The Firebase component provides a mechanism similar to
TinyWebDB. However it is hosted using a Firebase database
(firebase.com) and supports updates to variables from different copies
of an application. If multiple copies of an application are running on
different devices which are connected to the network. When one device
updates a variable the other will get a “DataChanged” event with the
new value. This permits a form of “cloud variable.”

The use of Firebase normally requires an account at firebase.com, but
by default an account sponsored by MIT is configured (via “branding”
commits which are not public) for the MIT sponsored service.

Author: Kristin Asmus <kna934@gmail.com>
Author: William S. Byrne <will2596@gmail.com>
Author: Jeffrey I. Schiller <jis@mit.edu>

Change-Id: I197445bd4ddd7eafb188c43569f902e7f5924c4d


jisqyv:
@halatmit @afmckinney @William-Byrne Here is the Firebase Component (one commit, all squashed down). It's ready! I'm working on updating firebasetest.appinventor.mit.edu to run this code.

Please review!

jisqyv:
@halatmit @afmckinney @William-Byrne 

  http://firebasetest.appinventor.mit.edu is now running this code (and the rest of the master branch)

[It is still an instance of the stand alone code as well]

halatmit:
I did a quick test, but did no try anything fancy with accounts or buckets.   The question is, should we release a version that does only the simple stuff; or should we keep the full suite of commands in (including unauthenticate)  and say something in the documentation about how these might change their behavior?

jisqyv:
Are you proposing removing the ability for people to configure their own account?

I would be in favor of leaving things are they are, and tell people that setting the account to something other then DEFAULT is an expert option.

I'm not sure what to do about Unauthenticate. Without that block, the only way to get Firebase to reset is to uninstall the Companion and re-install.

halatmit:
Let's leave things as they are and handle it in the documentation.




== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
@halatmit @afmckinney I just marked this PR as on hold. I'm going to implement a suggestion from Mark Sherman to *not* re-initialize the firebase connection if the project bucket is set to the same value (probably something like a three line change). I'll add it here as a separate commit in this PR when it is ready.

halatmit:
Good idea.

Thanks.  And thanks, Mark.




== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.



On Sun, Jan 10, 2016 at 10:56 AM, Jeffrey Schiller <notifications@github.com

jisqyv:
@halatmit @afmckinney I have added the necessary change (so this review is "live" again). I have tested the change locally, but have not yet put the updated code on firebasetest.appinventor.mit.edu

William-Byrne:
I built a few test apps last night, and everything seems to be working well. Great work getting this finished Jeff!

jisqyv:
@halatmit @afmckinney I've updated firebasetest.appinventor.mit.edu

halatmit:
LGTM

jisqyv:
Last commit was a squash of the two previous commits

PR #671, opened 2016-01-13
added project name validation to save project as, validation code mov…

bxie:
…ed to TextValidators.java

Current Implementation:
When creating a new project and input invalid project name, help text appears to warn you. When you "save project as", this help text does not appear until after you click "ok".

My fix:
Added validation to SaveAs. Had to move validation code to TextValidation for reuse.

To test:
In created project, rename project (Projects > Save Project As). Attempt to input invalid project name. Ensure appropriate error message appears (prior to clicking "ok")

NB:
-for SaveAs, could not disable OK button when invalid project name (I say this is ok because validation after clicking "ok" will catch this so invalid project name still rejected)
-validation needs to be added to RemixedYoungAndroidProjectWizard.java (should look exactly like NewYoungAndroidProjectWizard.java). I don't know how to test this (need gallery) so I did not do this.

@afmckinney

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/explorer/commands/CopyYoungAndroidProjectCommand.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/explorer/commands/CopyYoungAndroidProjectCommand.java
@@ -145,10 +145,7 @@ public final class CopyYoungAndroidProjectCommand extends ChainableCommand {
         @Override
         public boolean validate(String value) {
           errorMessage = TextValidators.getErrorMessage(value);
-          if (errorMessage.length()>0){
-            return false;
-          }
-            return true;
+          return !(errorMessage.length()>0);
         }
 
         @Override
--- a/appinventor/appengine/src/com/google/appinventor/client/explorer/commands/CopyYoungAndroidProjectCommand.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/explorer/commands/CopyYoungAndroidProjectCommand.java
@@ -16,6 +16,7 @@ import com.google.appinventor.client.youngandroid.TextValidators;
 import com.google.appinventor.shared.rpc.project.ProjectNode;
 import com.google.appinventor.shared.rpc.project.ProjectRootNode;
 import com.google.appinventor.shared.rpc.project.UserProject;
+import com.google.appinventor.client.widgets.Validator;
 import com.google.gwt.event.dom.client.ClickEvent;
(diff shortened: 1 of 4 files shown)


appinventor/appengine/src/com/google/appinventor/client/youngandroid/TextValidators.java
```diff
@@ -101,4 +101,25 @@ public static boolean isValidCharFilename(String filename){
...
+   * @param filename The filename (not path) of uploaded file
+   * @return String representing error message, empty string if no error
+   */
+  public static String getErrorMessage(String filename){
+    String errorMessage = "";
+    String noWhitespace = "[\\S]+";
+    String firstCharacterLetter = "[A-Za-z].*";
+    if(!filename.matches("[A-Za-z][A-Za-z0-9_]*") && filename.length() > 0) {
+      if(!filename.matches(noWhitespace)) { //Check to make sure that this project does not contain any whitespace
+        errorMessage = "Project names cannot contain spaces";
```
josmas: these messages should be internationalised.

appinventor/appengine/src/com/google/appinventor/client/explorer/commands/CopyYoungAndroidProjectCommand.java
```diff
@@ -140,7 +141,21 @@ public CopyProjectDialog(final ProjectRootNode oldProjectNode) {
       }
 
       newNameTextBox = new LabeledTextBox(checkpoint ? MESSAGES.checkpointNameLabel()
-          : MESSAGES.newNameLabel());
+          : MESSAGES.newNameLabel(), new Validator(){
+        @Override
+        public boolean validate(String value) {
+          errorMessage = TextValidators.getErrorMessage(value);
+          if (errorMessage.length()>0){
```
josmas: you could basically `return !errorMessage.length() > 0;`
or `return errorMessage.length < = 0;` instead of using the negation.

bxie:
ok @josmas made those changes. thanks!
how do we know what messages have not been internationalized? (since I just added three messages to OdeMessages.java)

josmas:
@bxie I don't think there's an automated way of knowing what's missing for I18n. All strings in interface should be, so whenever you see one, just change it.

It would be good to also keep the other languages file in sync, even if the English sentence is used (at least we can keep track that they need to be translated).

afmckinney:
squashed commits rebased on master and got rid of white space. LGTM moved to Gerrit.

PR #674, opened 2016-01-14
Backpack

a-suarez:
Added designs for new backpack icon. Updated view and its behavior to use new icons.

@afmckinney @jisqyv


jisqyv:
@a-suarez Hmmm. When I empty the back pack, it still shows that there are blocks in it. It shrinks, but the blocks (not the real blocks, but the icon with blocks showing) are still there.

a-suarez:
@jisqyv I updated code to fix the 'empty backpack' behavior

jisqyv:
@a-suarez LGTM, moving to Gerrit

jisqyv:
MERGED

PR #686, opened 2016-01-26
Issue 682: Welcome popup, splash screen, and viewing the Gallery

ajcolter:
Added an additional check to hide the welcome and the splash screens if the current view is the Gallery. Unfortunately, I haven't been able to test this because I need a real Gallery link, which I cannot do locally.


marksherman:
I think you can enable the gallery in local by turning the flag on in appengine-web.inf:

```diff
diff --git a/appinventor/appengine/war/WEB-INF/appengine-web.xml b/appinventor/appengine/war/WEB-INF/appengine-web.xml
index b0f6de2..74ff0d9 100644
--- a/appinventor/appengine/war/WEB-INF/appengine-web.xml
+++ b/appinventor/appengine/war/WEB-INF/appengine-web.xml
@@ -56,9 +56,9 @@
     <property name="use.gcs" value="true" />
 
     <!-- Name of gallery bucket in Google Cloud Store -->
-    <property name="gallery.bucket" value="" />
+    <property name="gallery.bucket" value="my-gallery-bucket" />
     <!-- Whether or not to use the gallery -->
-    <property name="use.gallery" value="false" />
+    <property name="use.gallery" value="true" />
```

ajcolter:
I did that, but it was an empty gallery. And it was all on localhost, no
unique address, so I couldn't follow the steps you gave me to replicate the
bug.

marksherman:
Yes you can. Just use two accounts.
Log in with an account, to your localhost server, make and share a project
to the gallery. The share button gives you a local url:
`localhost:8888/?galleryId=4785074604081152` which then works to trigger
the bug when you log in as a second account.

appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/ProjectToolbar.java
```diff
@@ -157,7 +157,8 @@ public void onSuccess(Void result) {
               Ode.getInstance().getProjectManager().removeProject(projectId);
               // Show a welcome dialog in case there are no
               // projects saved.
-              if (Ode.getInstance().getProjectManager().getProjects().size() == 0) {
+              if (Ode.getInstance().getProjectManager().getProjects().size() == 0
+                  && Ode.getInstance().getCurrentView() != 2) { //2 is the int for Gallery
```
josmas: What is the 'int for the Gallery'? Can this be a **constant with a proper name** instead of a **magic number**?

ajcolter: I agree, it's a silly thing. In Ode.java (starting on line 181) it has this code:
```
  // Remembers the current View
  static final int DESIGNER = 0;
  static final int PROJECTS = 1;
  private static final int GALLERY = 2;
  private static final int GALLERYAPP = 3;
  private static final int USERPROFILE = 4;
  private static final int PRIVATEUSERPROFILE = 5;
  private static final int MODERATIONPAGE = 6;
  private static int currentView = DESIGNER;
```
I'm not sure how to extract this to a constant that exists outside of Ode. Where do those live?

josmas: Can't recall where things go, but isn't Ode.java like the main entry point to the Designer? You should be able to use `Ode.GALLERY` from the other files?

appinventor/appengine/src/com/google/appinventor/client/Ode.java
```diff
@@ -1439,7 +1439,8 @@ private void openProjectsTab() {
     getProjectService().getProjects(new AsyncCallback<long[]>() {
         @Override
           public void onSuccess(long [] projectIds) {
-          if (projectIds.length == 0 && !templateLoadingFlag) {
+          if (projectIds.length == 0 && !templateLoadingFlag
+              && Ode.getInstance().getCurrentView() != 2) { //2 is the int for the Gallery
```
josmas: You use `GALLERY` in line 1465; would be a good idea to do the same here.

ajcolter: Good catch. :)

appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/ProjectToolbar.java
ajcolter: I've just tried Ode.GALLERY and Ode.getInstance().GALLERY and they don't work. Both say it's a private field. Should I write a getter method?

ajcolter:
Uploaded new code. Jose helped me fix the random constant. However, I realized that the Welcome box should really only ever be displayed IF we are on the projects view (right?). Tried to change the condition in the If statements to that. However, now the welcome box is never displayed, which makes me think that maybe the currentView isn't updating properly? I can't find where to see the OdeLog message I wrote in, but if I could see it, then maybe I could figure out what is going wrong.

josmas:
@ajcolter if you are logged in as an Admin, you should see an `Admin` tab containing something like `switch to debug panel` (or similar). Logging ends up there at times. If the Log statement you used does not appear, `grep` the appengine client folder for any string that does appear in the debug view, and do the logging that way. Does that make sense?

ajcolter:
Makes sense. I found it. It's saying the current view is int 0, which is the designer. But I don't know why it's saying that when I log in for the first time... But that explains why, with my code the way it is, the welcome box is no longer displayed.

ajcolter:
Jeff wrote something so the welcome box won't show when looking at a gallery page. He'll be creating the PR.
Also, we decided to stop showing the splash screen, except for major updates (like needing a new companion).

PR #690, opened 2016-01-30
Provide a Menu Item to Update the Companion

jisqyv:
Provide a Menu Item under the “Help” Menu to trigger a manual update
of the Companion. Also fix a bug in the Companion itself so that the
correct CORS headers are returned during the update (also put a work
around in replmgr.js to deal with the non-fixed
Companion). Automatically disconnect the Companion after the update so
the user doesn’t have to use the “Reset Connection” item before
connecting to the newly installed companion.

Change-Id: I31215863163974cf8f23bb68b03200cad8229ae5

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
@@ -563,6 +563,14 @@ public interface OdeMessages extends Messages {
   @Description("Information about the Companion")
   String companionInformation();
 
+  @DefaultMessage("Update the Companion")
+  @Description("Menu item to update the Companion to the latest version")
+  String companionUpdate();
+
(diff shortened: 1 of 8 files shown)


jisqyv:
@halatmit @afmckinney Please Review

halatmit:
looks OK, but I did not test.   so I guess LGTM

jisqyv:
I'm assuming you meant to put the "OK to move to Gerrit" label... not "reply to review" because I don't see a question...

halatmit:
LGTM

halatmit:
I mean that  I didn't test it.   But I just changed this to LGTM



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
I tested it though...

PR #692, opened 2016-02-02
Addressing image sprites are not scaled correctly in the designer #653

### LINKED ISSUE #653: Image sprites are not scaled correctly in the designer
See Ghica's note:

https://groups.google.com/forum/#!category-topic/mitappinventortest/FDEEx9PYyF0

The problem is both with images and image sprites, although they are slightly different.
In all cases, the phone works correctly.  It's the designer that's not scaling the image.   And there are 
different results for images depending on whether you select Scale Picture to Fit.

E-Hon:
see #653 
The code replaces ScalePictureToFit with Scaling property, a drop list with "Scale proportionally" and "Scale to fit". I failed to code "Do not scale" on the andriod side. Note that in the current version, when unchecking ScalePictureToFit, "Scale proportionally" is the behavior on android while not scaling the picture is that in ode.

It doesn't actually fix #653 which occurs when ScalePictureToFit is unchecked but it is solvable. #653 won't occur with the new property.

This also fixes the position of the icon, fixes the case when refreshed, the image restores to its original dimensions regardless of ScalePictureToFit and other minor things.

The new property is not added to ImageSprite because ImageSprite is very different from Image. ImageSprite uses BitmapDrawable which doesn't have function to scale easily.

~~The code for verioning.js to upgrade blocks is not finished because i don't know how to remove property blocks~~


halatmit:
Ethan,

To deprecate a property, see the section recently added to

How to add a property to a component
<https://docs.google.com/document/d/1ukpZXcd0fIn1fV0cNp4J00wrfwwA1daswU1yY6dqqfs/edit#heading=h.vsfxd7ls6pw7>

For an example, look at how the UseFront property has been deprecated in
Camera   (that deprecation is right now in ai2-test and master, but not in
ai2 until the new release)

Let us know when the pull request is ready for review.   Also, it would
help if you could send a good test case we can try, with a note on what
behavior to expect.  Thanks so much for doing this.

== Hal




== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

E-Hon:
Thx @halatmit. I wasn't aware of the update. It's fixed now.

## Test case 1:
- Automatic Height + Fill parent Width, 40% Height + Automatic Width, Fill parent Height + 100 pixels Width
- small image using small image
- large image using larger image
- switch between "Scale proportionally" and "Scale to fit"
- setting the dimensions first or the Picture property first shouldn't matter

## Test case 2:
- refreshing the page should retain the changes

## Test case 3:
- changing the Picture property to none should switch back to the icon which will be center-aligned, regardless of the Scaling choice

## Test case 4:
- After importing an old project with an Image whose ScalePictureToFit is checked, the new scm file should not have ScalePictureToFit but have Scaling set to "1"

appinventor/components/src/com/google/appinventor/components/runtime/Image.java
```diff
@@ -173,4 +173,28 @@ public void ScalePictureToFit(boolean scale) {
   public void Animation(String animation) {
     AnimationUtil.ApplyAnimation(view, animation);
   }
+
+  @DesignerProperty(editorType = PropertyTypeConstants.PROPERTY_TYPE_SCALING,
+      defaultValue = Component.SCALING_SCALE_PROPORTIONALLY + "")
+  @SimpleProperty(description = "This allows you to scale the picture in " +
```
halatmit: We should internationalize new tooltips since we'll be going back later an internationalizing them all.   See the note on internationalization in How to Add a Component.

Also change the text to make it more clear (add there are now only two choices, not three).

This property determines how the picture scales according to the Height or Width of the Image.   Scale proportionally (0) preserves the picture aspect ratio.  Scale to fit (1) matches the Image area, even if the 
aspect ratio changes.

Don't forget to add this to the web documentation as described in How to Regenerate the App Inventor Documentation.
Also feel free to add the Simplified or Traditional Chinse localizations for the new strings your' added.

halatmit:
@E-Hon 
Nice job.  Thanks!  Just a couple of things on the documentation.

E-Hon:
@halatmit 
Just put some changes as per your comment. 

We can't currently internationalize description of properties, right? I couldn't find any translated help strings for properties.

halatmit:
@E-Hon 
Leave these not internationalized.  I looked at the code and there are some inconsistencies about how this is being done that we should clean up first.

E-Hon:
OK, then all the changes are there

halatmit:
LGTM

@E-Hon Thanks!!!

jisqyv:
@halatmit I would like to "undeprecate" the ScalePictureToFit property. So both properties are present for this coming release. This will let us keep our promise that the current companion will work with the new release. Once a Companion that supports the "Scaling" property is deployed (aka the new Companion we are about to release [assuming we merge this change]) we can deprecate ScalePictureToFit in a future release

halatmit:
@jisqyv I agree with the "undeprecate" for now, and we can deprecate it later, as you say.

E-Hon:
@jisqyv @halatmit 
So should I change the code to support both properties or just wait until the new companion is released?

halatmit:
@E-Hon @jisqyv 
I think Jeff is going to finish this up and add it to the delease.

jisqyv:
@E-Hon @halatmit Yep, I've finished the necessary changes. I'm testing them now. I'll post them to Gerrit directly (Ethan: You can visit our Gerrit Review site at https://gerrit.appinventor.mit.edu and login with a Google Account).

jisqyv:
@E-Hon @halatmit The change is now on gerrit at: https://gerrit.appinventor.mit.edu/#/c/1888/

Patch Set 1 is Ethan's original code and Patch Set 2 is my update to it.

With this modified change, the fix to the Designer is still present, but it is implemented with the "ScalePictureToFit" property. The Scaling property is introduced, so it is included in the Companion (version 2.36 when we release) but is hidden in the UI. When we do the next Component/Companion release (to say version 2.37) we can update the code to expose the Scaling property and eliminate the ScalePictureToFit.

E-Hon:
@jisqyv thx for the update. Please let me know if there's anything I can help

jisqyv:
MERGED

PR #699, opened 2016-02-12
Remove comment about supporting SDK level 3 in Compiler.java

halatmit:

PR #701, opened 2016-02-12
clean up broadcast receiver permission

josmas:
Here's the change I mentioned in #699 (have rebased, so should be clean to merge when reviewed).

I am working on an annotation for broadcast receivers, using Texting as a test, and I noticed that weird permission in a place that is not even [allowed in the xml schema](http://developer.android.com/guide/topics/manifest/action-element.html). Kinda surprised it does not crash.

@halatmit @jisqyv


jisqyv:
@josmas Closing, this change is made obsolete by #703

PR #703, opened 2016-02-14
Adds an annotation for Broadcast Receivers

josmas:
Adds an annotation to mark a class as a _@SimpleBroadcastReceiver_ with attributes _class name_ and _actions_ (to be added as intent filters).

The **_Texting_** component is changed to use the new annotation (it's the only example in the system right now).

The main purpose of this change is that _Extension_ developers can create extensions that use receivers without having to modify Compiler.java directly.

Please review @halatmit @jisqyv

### DIFF
--- a/appinventor/buildserver/src/com/google/appinventor/buildserver/Compiler.java
+++ b/appinventor/buildserver/src/com/google/appinventor/buildserver/Compiler.java
@@ -1,6 +1,6 @@
 // -*- mode: java; c-basic-offset: 2; -*-
 // Copyright 2009-2011 Google, All Rights reserved
-// Copyright 2011-2012 MIT, All rights reserved
+// Copyright 2011-2016 MIT, All rights reserved
 // Released under the Apache License, Version 2.0
 // http://www.apache.org/licenses/LICENSE-2.0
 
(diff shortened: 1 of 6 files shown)


kkashi01:
Niceeeee

josmas:
thanks @kkashi01 

In two minds about adding the annotation to the Texting class or to the BR class itself (SmsBroadcastReceiver class in this case). Any thoughts on that? It works exactly the same, and it would be rare that two components decide to reuse the same Broadcaster, so not sure which way to go.

kkashi01:
@josmas I was referring to use of annotation and making compiler code abstract for avoiding hardcoded manifest stuff; a great idea

josmas:
@kkashi01 yeah, thanks; permissions and libraries were already there, so nothing new :)

In any case, my comment is about where to use the annotation; in the receiver itself, or in the class (component class) that uses the receiver. They both work; it's just a matter of thinking which one makes more sense.

kkashi01:
@josmas Not sure, but I think at component level (where you have it) and since the two classes are tightly copuled, pulling in the receiver into component as inner class

jisqyv:
@josmas MERGED onto the ucr branch

josmas:
Thanks, @jisqyv

PR #704, opened 2016-02-14
Change horizontal alignment so that 2 is centered and 3 is right, in

halatmit:
accord with the documentation.
Note that this is an incompatible change that will break current projects.

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/common/ComponentConstants.java
+++ b/appinventor/components/src/com/google/appinventor/components/common/ComponentConstants.java
@@ -48,8 +48,9 @@ public class ComponentConstants {
    * HorizontalArrangement, VerticalArrangement, and Screen
    */
   public static final int GRAVITY_LEFT = 1;
-  public static final int GRAVITY_RIGHT = 2;
-  public static final int GRAVITY_CENTER_HORIZONTAL = 3;
+  public static final int GRAVITY_CENTER_HORIZONTAL = 2;
+  public static final int GRAVITY_RIGHT = 3;
+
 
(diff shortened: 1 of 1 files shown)


halatmit:
I'd not sure we should do this, since it will break existing projects (but seems more logical and also compatible with the docs).   Suggestions?

josmas:
Is it possible to write an upgrader? Otherwise I'd be weary to do so.

halatmit:
@josmas What do you think the upgrader should do?  I don't see how to do one in general.

josmas:
Is this affecting the numbers used in Blocks when setting alignment in arrangements? I'd say it would not only be blocks, but also property values in scm files, right?

If that's so, the upgrader should take all arrangements and swap 2s and 3s?

halatmit:
We can do the swap if the block has a literal 2 or 3, but not if it's a computed value.
Is that second case rare enough that we can go ahead and break things?   Another approach is to mark the block an add a message to the user warning them to fix it.   We did that sort of thing in Google App Inventor when we switch trig function default from radians to degrees.

josmas:
What do you mean by a computed value in this context? If the value is not either 2 or 3, it should not change.

halatmit:
I mean if  the value is set to an expression, or the value of a variable.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

josmas:
Right, that makes sense now.I'd say that would be rare enough, though.

halatmit:
I'll work on it.

halatmit:
Subsumed by #711 
Closing

PR #706, opened 2016-02-16
Update display list documentation for AI2 (The trivial change here is…

halatmit:
… to force a re-push)

### DIFF
--- a/appinventor/docs/reference/other/displaylist.html
+++ b/appinventor/docs/reference/other/displaylist.html
@@ -457,6 +457,7 @@
 
            Click here for a version of this page on which you can comment.</a>
 
+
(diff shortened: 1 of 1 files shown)


josmas:
What do you mean by _**forcing a re-push**_?

halatmit:
Someone I thought I'd pushed this to github before, but it's not there.
So I'm trying again.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

josmas:
Are you in the right branch? There's only one file change in here, and it's the trivial change you pushed last (no other changes than a space added to _displaylist.html_).

jisqyv:
Change already made

PR #711, opened 2016-02-20
Change horizontal alignment so that 2 is centered and 3 is right, in

halatmit:
accord with the documentation.
Note that this is an incompatible change that will break current projects.

Working on create version upgrade for fix horizintal alignment

Continue work on upgrader

Added upgrader code.   Now ready to debug.

Still debugging

Finishing AlignHorizontal upgrader

Change the meaning of the arguments to AlignHorizontal and add
upgrader for Screen and HorizontalArrangement blocks


halatmit:
@jisqyv @afmckinney 
I don't see where the merge conflict is coming from.  Please let me know.

jisqyv:
@halatmit I'm going to figure out the merge conflict today or tomorrow.

halatmit:
@jisqv Thanks.  Let me know if I can help.

halatmit:
@jisqyv I think this can be merged now, assuming the component and language version numbering is now correct.  do I need to do anything more?

jisqyv:
@halatmit We should have a brief conversation first. I'm concerned about breaking projects.

jisqyv:
@halatmit I believe it should be possible to actually upgrade the blocks rather then just disable them...

jisqyv:
@halatmit I also believe a form upgrader is needed to change the properties in the designer.

jisqyv:
@halatmit Is there a good reason to not just fix the documentation?

halatmit:
replaced by #760 which fixes the issue by only changing the documentation

PR #717, opened 2016-03-03
Retry failed searches in Gallery

toropeza:
This pull request fixes the errors thrown in the current gallery search. Will retry up to 3 times in order to get the correct results if a Transient Error is thrown.

Working version (Although this cannot be tested here since I am not getting Transient Errors on my instance) : https://searchfix-dot-aigallerythomas.appspot.com/
Use the following search keywords on my instance: "search", "debug"

@jisqyv @wolberd

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/server/GallerySearchIndex.java
+++ b/appinventor/appengine/src/com/google/appinventor/server/GallerySearchIndex.java
@@ -41,7 +41,8 @@ public class GallerySearchIndex {
       GalleryStorageIoInstanceHolder.INSTANCE;
   private static volatile GallerySearchIndex  instance= null;
   private static Cursor searchCursor;
-  private final int appResultMaxAccuracy = 100;
+  private static final int SEARCH_RETRY_MAX = 3;
+  private final int NUMBER_FOUND_ACCURACY = 100;
   /**
    * The default constructor of GallerySearchIndex
    */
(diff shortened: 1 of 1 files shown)


appinventor/appengine/src/com/google/appinventor/server/GallerySearchIndex.java
```diff
@@ -113,32 +113,43 @@ public GalleryAppListResult find (String searchWords, int start, int count) {
...
+    int attempts = 0;
+    while (attempts++ < 3){
+      try {
+        LOG.info("Sending query " + query);
+        results = getIndex().search(query);
+        break;
+      } catch (SearchException e) {
+        if (StatusCode.TRANSIENT_ERROR.equals(e.getOperationResult().getCode())) {
+          // retry
+          LOG.info("Query failed on attempt:"+ attempts + ". retrying...");
```
jisqyv: @toropeza This doesn't seem correct. If the error is NOT a TRANSIENT_ERROR, then you will still retry, but just be silent. Errors that are not TRANSIENT should cause a failure without retrying...

toropeza: @jisqyv @wolberd Didn't catch that thanks, I just pushed a fix.

jisqyv:
@toropeza LGTM MERGED

Issue #718, opened 2016-03-04
Add "/" To the End of a Firebase URL


jisqyv:
If a provided Firebase URL is lacking a trailing slash, add one.

E-Hon:
hi @jisqyv 
i created a pr for this and this bitesize issue is bitten :D

E-Hon:
one thing to ask though. why is a trailing slash necessary?

jisqyv:
@E-Hon I don't know why it is needed. It is a requirement of the firebase provided client library.

jisqyv:
Fix merged.

PR #720, opened 2016-03-06
Addressing-Add-/-To-the-End-of-a-Firebase-URL-#718

E-Hon:
see #718 
because i didn't have a valid firebase url, I tested the code by disabling other verification code. 

i also considered adding code to the property editor to automatically append a slash if it's not present but i thought it would be better to just process the url either in the android code or in app inventor but not both which may lead to inconsistency. hope it's fine.


jisqyv:
@E-Hon MERGED onto ucr

E-Hon:
Thx @jisqyv. Shall we also close its corresponding issue #718?

jisqyv:
@E-Hon I closed it. Thanks for the reminder :-)

PR #721, opened 2016-03-07
Fix title of "intex in list" in error message

halatmit:


halatmit:
This was a trivial bug resulting from name change "position in list" to "index in list"

jisqyv:
@halatmit LGTM

jisqyv:
@halatmit MERGED

PR #722, opened 2016-03-07
DropDownButton adapting to browser boundaries

brandonwang15:
@jisqyv @afmckinney 

Changed `DropDownButton `functionality in the Designer Window. See the TextColor dropdown for Buttons as an example.

**Problem:**
The `ContextMenu `previously would always drop 'down' and if there wasn't enough room below it, it would force the screen to extend so that it could display the entire menu. 

**Solution:**
`ContextMenu `will drop 'up' if there isn't enough room below it, rather than pushing down the screen boundaries. Note that in the case that dropping up will also push up the screen boundaries, it will default to dropping down.

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/widgets/DropDownButton.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/widgets/DropDownButton.java
@@ -46,11 +46,15 @@ public class DropDownButton extends TextButton {
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


jisqyv:
@brandonwang15 LGTM Moved to Gerrit

jisqyv:
@brandonwang15 MERGED

PR #725, opened 2016-03-09
Verify the Companion Package

jisqyv:
Make sure that the package name of the Companion is matches this
version of App Inventor. Different versions of App Inventor (different
forks) should have their own Companions, each with its own package
name. This check makes sure that a person is using the correct
Companion for the version of App Inventor they are using. The expected
version is set in YaVersion. The default value provided is for the MIT
App Inventor service provided by MIT.

Change-Id: I23f48ab017020448a01a5a8d781f831fe6cec450


jisqyv:
@afmckinney @halatmit Please Review

appinventor/blocklyeditor/src/msg/en/_messages.js
```diff
@@ -1304,6 +1304,7 @@ Blockly.Msg.en.switch_language_to_english = {
     Blockly.Msg.REPL_NOT_NOW = "Not Now";
     Blockly.Msg.REPL_COMPANION_OUT_OF_DATE1 = "The Companion you are using is out of date.<br/><br/>This Version of App Inventor should be used with Companion version";
     Blockly.Msg.REPL_COMPANION_OUT_OF_DATE_IMMEDIATE = "You are using an out-of-date Companion. You should update the MIT AI2 Companion as soon as possible. If you have auto-update setup in the store, the update will happen by itself shortly.";
+    Blockly.Msg.REPL_COMPANION_WRONG_PACKAGE = "Make sure you are using the correct MIT AI2 Companion";
```
halatmit: Probably better to say
The Companion you are using was built for different instance of App Inventor.  To obtain the correct companion look on the App Inventor screen under Help and Companion Information.

-- 
And ace we make sure the companion will always be there?

jisqyv: @halatmit I'll make the change. There will always be some information under the Companion Information menu. If there isn't a companion available from the service itself, it will at least give you the version number. At the moment it doesn't give you the "name" of the companion to use. This is probably best left for the "branding" commits, as we have no way of knowing what the name of the companion is (we can learn its package name, but that can have no relation to the name given tot he App both on the device and in the store).

halatmit: Is it possible to always have the companion link there?   People are
already being sloppy about creating a companion at a find able place.
Can't we always stash a copy with the service?


== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
@halatmit I've updated the error message.

appinventor/blocklyeditor/src/msg/en/_messages.js
jisqyv: @halatmit Let me think about it. But that shouldn't be part of this change. It will requite some rework of the build system, but I believe it can be done.

halatmit:
LGTM

jisqyv:
MERGED

PR #735, opened 2016-03-18
Mark Projects with our Host Name

jisqyv:
Because of the likely proliferation of various versions of App
Inventor, we want to mark a project with the history of which versions
have seen it. We do that with the "authURL" tag which we add to the
Form files. It is a JSON array of versions identified by the hostname
portion of the URL of the service editing the project. Older projects
will not have this field, so if we detect an older project (YAV <
OLD_PROJECT_YAV) we create the list and add ourselves. If we read in a
project where YAV >= OLD_PROJECT_YAV *and* there is no authURL, we
assume that it was created on a version of App Inventor that doesn't
support project tagging and we add an "*UNKNOWN*" tag to indicate
this. So for example if you examine a (newer) project and look in the
Screen1.scm file, you should just see an authURL that looks like
["ai2.appinventor.mit.edu"]. This would indicate a project that has
only been edited on MIT App Inventor. If instead you see something
like ["localhost", "ai2.appinventor.mit.edu"] it implies that at some
point in its history this project was edited using the local dev
server on someone's own computer.

Change-Id: I7a1012d163404eb6725944fbc56c0d50604a62c0

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/editor/youngandroid/YaFormEditor.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/editor/youngandroid/YaFormEditor.java
@@ -30,11 +30,13 @@ import com.google.appinventor.client.editor.youngandroid.palette.YoungAndroidPal
 import com.google.appinventor.client.explorer.SourceStructureExplorer;
 import com.google.appinventor.client.output.OdeLog;
 import com.google.appinventor.client.properties.json.ClientJsonParser;
+import com.google.appinventor.client.properties.json.ClientJsonString;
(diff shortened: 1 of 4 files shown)


jisqyv:
@afmckinney @halatmit Please Review

jisqyv:
@afmckinney @halatmit P.S. This change also adds a ZIP file comment to exported AIA files. It reads "Built with MIT App Inventor." This will help us detect the case where someone unzips a project into their own directory and then zips it back up.

jisqyv:
MERGED

afmckinney:
LGTM

PR #741, opened 2016-04-04
Added support for creating components with Kotlin.

davidlindenbaum:


josmas:
Hi @davidlindenbaum as discussed in the document, this should be optional, at least for now. Could it be implemented as a target?

jisqyv:
I am not in favor of this change. There is significant code maintenance added when you add another language. We should not do so without a strong compelling argument.

halatmit:
I am not in favor either.  Maybe this is something that can be addressed
extensions or with a separate instance (not run by MIT).



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

davidlindenbaum:
Would it be better if it was an optional build target so there's no impact until people write components in Kotlin? Also I wouldn't expect there to be much of a maintenance issue - Kotlin compiles to Java bytecode so there's no difference between a Kotlin component and a Java one, so it's essentially just a library.

josmas:
@davidlindenbaum yeah, an optional target would be very helpful, at least as a proof of concept that this can be done. Even if it does not get merged at this point, developers could use it locally for their own distributions. Also with the rate of adoption of Kotlin in the Android community, it would be good to be prepared to integrate with Kotlin libraries in the future, even if the language is not used for specific components.

halatmit:
@davidlindenbaum As Jeff wrote, our hesitation here is about code maintenance and support.  We also don't want to have to include extra libraries (like the Kotlin jars) in the app inventor system and further expand the size of the companion for a feature that on a few people would use.   Can you figure out a way to avoid this?   But also, can you deploy a Kotlin enabled instance and make it publicly available so people can experiment with writing Kotlin components and libraries?

davidlindenbaum:
@halatmit @jisqyv @josmas 
Kotlin is now completely optional - a normal ant build will be the same as before (including filesizes of the companion and compiled apps) and ant -Dusekotlin=true will build with support for Kotlin components. Adding <property name="usekotlin" value="true"/> in a build.xml will enable Kotlin by default.
As far as including the Kotlin jars in the repository, I guess there's not really a way around that, but it's not that much extra size or anything like that compared to the libraries already there.

jisqyv:
@davidlindenbaum I'm still a little concerned. In particular as soon as someone writes and submits a component written in Kotlin, we, the development team, now have an obligation to maintain that component going forward. I am a veteran of a two software systems that had a portion written by a contributor who used a non-standard library or language. Those libraries and languages never gained traction and eventual rotted due to lack of maintenance. Eventually we had to remove a very important part of the system because we simply could not keep it working as written and we didn't have the resources to re-write it in a supported language. Been there, done that, don't want a repeat...

josmas:
Hey guys, this is great. It doesn't have to get merged, but we now know we can use it if we ever need to.
And other distros can use it if they want to, even if it's not merged in the main repo, so thanks @davidlindenbaum and @BrandonAPrice for your work on this; much appreciated!

halatmit:
@davidlindenbaum @BrandonAPrice
Thanks for doing this, but we've decided not to merge it into main MIT AppInventor because we don't want to take on the obligation of supporting it.  As Jos says, the other distros can use this, and it would be good for them to experiment with it.

PR #742, opened 2016-04-05
Add Support for Local Logins

jisqyv:
This change adds support for local logins as well as maintaining
compatibility with Google authentication (when running on App Engine).

A “local login” is an account whose login name is an email address and
which has a password which we store in the UserData object (hashed via
PBKDF2). Two new flags “auth.usegoogle” and “auth.uselocal” control
whether or not Google Logins and/or Local logins are accepted.

If auth.usegoogle is false and auth.uselocal is true, a login page is
displayed which prompts for a person’s email address and password. A
“set/recovery password” link is also displayed which can be used to
change a person’s password or set a new password. This link is how local
login accounts are created. When clicked, the person is prompted for
their e-mail address. We then e-mail them a link which they click on to
confirm their account. This takes them to a page where they can set
their password.

Note: Because of restrictions imposed by Google, we cannot send e-mail
directly from App Engine. Instead we use a “helper” server which accepts
an HTTP POST operation from us (using the Urlfetch functionality of App
Engine). This helper does the actual sending of e-mail. This also
permits us to control where the email originates from (i.e., we can use
a server at MIT instead of a server at Google as some locations care
from where the mail comes). The source code for the “helper” server is
included in this commit (in the misc/passwordmail directory. See the
README file there for more details).

If auth.usegoogle and auth.uselocal are both set, the login page has a
link labeled “Use Google Account” which will then cause their Google
account to be used to login.

If auth.usegoogle is true and auth.uselocal is false, then the system
behaves as it does today. No local login page is shown, instead Google
accounts are used (which means that if someone isn’t logged in they see
Google’s account page as they do today).

We also add a “User Admin” feature that permits authorized users to
change the password (but not a Google password) for other users. Admins
can also “become” another user for purposes of examining or downloading
their projects. When in this “become” mode, the system is read-only. You
can look at other user’s project, and interact with them, but you cannot
save any work.

Change-Id: Id20c49b3dadbb5e5f2180364a928991dd31343bd


jisqyv:
@halatmit @afmckinney Please Review.

Note: This is a *large* change. However the scariest code is in ObjectifyStorageIo.java. If you want to spend time grocking code, that is the code to grock...

-Jeff

jisqyv:
@halatmit @afmckinney This code is now up on ai2-test.appinventor.mit.edu. You are both "admins" so you can see how the Admin panel works.

jisqyv:
This review is being conducted on Gerrit due to its size and complexity.  -Jeff

halatmit:
I don't see a way to usefully review this, since it's so complex.   Is there some other way we can test it?

jisqyv:
It's running on ai2-test and was used on the contest server.

-Jeff

jisqyv:
It may be easier to look at this change on Gerrit as Gerrit's diff tools
work better for large changes.

-Jeff

halatmit:
Somehow I've lost my link to Gerrit code reviews and this submission.

Please send it.



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
You can find it on Gerrit at:

https://gerrit.appinventor.mit.edu/#/c/1899/

-Jeff

halatmit:
- How do you use "become" mode, on another user?

- Can any admin turn on the admin bit for another user?  If so, that seems
too powerful.  Perhaps we want to have a superuser status.






== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.



On Sun, Apr 10, 2016 at 11:05 AM, Jeffrey Schiller <notifications@github.com

jisqyv:
@halatmit On the Admin Page, click on a person's email address. This should pull up a dialog box where you can set them as an admin (and gallery moderator iff the gallery is enabled, which it isn't on ai2-test). You can set their password (doesn't change Google password) and there is a button to switch to their account (in read-only mode).

I agree that Admin is pretty powerful. I'm thinking that we should add a "power user" flag which is like admin but more limited.

For now, I want to do this in another change, I want to get this one done...

Btw. The code here is a little out-of-date as I am adding more patch sets on Gerrit.

jisqyv:
MERGED

PR #747, opened 2016-04-11
Resolved issue #173 - preventing AIA upload as media asset

varunm22:
Resolved issue detailed at https://github.com/mit-cml/appinventor-sources/issues/173 and replaced window.alerts with gwt dialog boxes. 

@afmckinney @jisqyv


appinventor/appengine/src/com/google/appinventor/client/wizards/FileUploadWizard.java
```diff
@@ -218,4 +240,58 @@ private void finishUpload(FolderNode folderNode, String filename,
...
+      
+      Button info = new Button ("More Info");
+      info.addClickListener(new ClickListener() {
+        public void onClick(Widget sender) {
+          Window.open("http://appinventor.mit.edu/explore/ai2/share.html", 
+            "AIA_Help", "");
+        }
+      });
+      holder.add(info);
+    } else if (errorType.equals("noFile")) {
```
jisqyv: You should get rid of this bunch of if ... then else if ... stuff. Instead just pass the error message to the dialog. as in:

`createErrorDialog(MESSAGES.noFileSelected(), folderNode, fileUploadedCallback);`

-Jeff

jisqyv:
@varunm22 See line note above.

appinventor/appengine/src/com/google/appinventor/client/wizards/FileUploadWizard.java
varunm22: Hi @jisqyv, 
I mainly sent the error type so I could tell whether or not a specific error type needed other buttons. For example, the aia upload error requires a "More Info" button. Should I just have something that checks if the sent error message is equal to MESSAGES.aiaMediaAsset() and adds the More Info button if so?

jisqyv: @varunm22 OK, I see what you are attempting to do. However it would be much cleaner to provide the error message as an argument, as I have suggested, and then add an additional errorType argument which should be a Java enum type. Then instead of a bunch of if .. then .. else .. if statements, you can use a switch statement based on the enum to determine what additional buttons should be placed, etc.

varunm22:
Hi @jisqyv, I fixed up the code with your suggestions, please let me know if there's anything else I can improve!

jisqyv:
@varunm22 Looks good... except... If you look at other components, you will see we use ALL UPPER CASE for enum values. Can you make that change?

varunm22:
@jisqyv I fixed it, thanks!

jisqyv:
Thanks. I'll do my final testing this evening and if all goes well I'll
merge it into the master branch.

-Jeff

jisqyv:
@varunm22 LGTM

There are some whitespace errors, but I'll take care of these when I squash things down for Gerrit

jisqyv:
@varunm22 MERGED

PR #748, opened 2016-04-11
Pedometer component

alicesypark:
The pedometer implementation has been re-implemented to measure steps more accurately. The component interface has been modified, and no longer uses the GPS to measure stride lengths -- it consumes more battery power with GPS. Tooltip descriptions have also been added.


josmas:
@alicesypark thanks for this. It would be nice to be able to test this on appspot [(here is how)](https://docs.google.com/document/pub?id=1Xc9yt02x3BRoq5m1PJHBr81OOv69rEBy8LVG_84j9jc#h.b70u04nuv2sm). You will also need to provide a link to an apk for the modified companion. Kirn and Laura are also working on pushing to appspot; you could team up for that if you want.

jisqyv:
@alicesypark I like this update to the Pedometer. The code appears to work. However before we can merge it YaVersion.java needs to be updated to indicate a revision to the Pedometer component as well as the documentation needs to be updated.

josmas:
@davidlindenbaum @alicesypark The changes to _YaVersion_ are a bit more involved than changing the component version number; please have a look at the [instructions here](https://docs.google.com/document/d/1ukpZXcd0fIn1fV0cNp4J00wrfwwA1daswU1yY6dqqfs/pub#h.2g0jyctaq6go), but skip point 4.3 as that is not in place anymore.

jisqyv:
@davidlindenbaum @alicesypark This change has gone 10 days without an update. We are getting ready to build a "Component" release (which this change requires). We only do such release once every few months. So I am going to finish this change up and test it. If all goes well it will make the release.

alicesypark:
@jisqyv We have not been successful in testing this on appspot yet :( As for the changes to YaVersion, I will push the updates.

jisqyv:
@alicesypark I'm almost finished. I have made the necessary changes. I'm going to close the PR and submit a new one which is complete. I'll point you at it so you can see what was needed.

jisqyv:
@alicesypark I'm closing this PR. take a look at #759 which is my updated version. It has been squashed down to one commit and rebased onto our "ucr" branch which is where we put changes for the Upcoming Component Release (UCR). I have kept David as the author of the commit. Let me know (comment in #759) if other people should also be added as authors.

Note that I added back events and properties that you removed, but marked them as deprecated. This ensures that any projects loaded that had used the Pedometer previously do not break (the deprecated blocks will be disabled and marked as "bad"). I'm attaching here an AIA (project) file that was made with the older version. If you load it into a the new version you will see what happens.

-Jeff

P.S. I have to rename the project to be a ZIP file so GitHub would accept it. When you download it, rename it from to end in '.aia'
[pedometerupgradetest.zip](https://github.com/mit-cml/appinventor-sources/files/243240/pedometerupgradetest.zip)

PR #750, opened 2016-04-14
Properly adjust font sizes within a ListView

jisqyv:
Make sure fonts are scaled based on device density when in “Responsive”
mode. In Fixed mode, the layout level scaling takes care of it for us,
but we have to do it here when we are in responsive mode.

This fix addresses Issue #713

Change-Id: Id51898981ab5db451ed783bad2ea3c3fb2853846

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/ListView.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/ListView.java
@@ -260,6 +260,7 @@ public final class ListView extends AndroidViewComponent implements AdapterView.
     // TODO(hal): Generalize this so that different items could have different
     // colors and even fonts and sizes
     int size = items.size();
+    int displayTextSize = textSize;
     Spannable [] objects = new Spannable[size];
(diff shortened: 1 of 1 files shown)


jisqyv:
@halatmit @afmckinney We need to test this on several different devices. I have tested it with 5.1.1 and 2.3.6 (emulated). I'll post a link to a compiled companion with this change. You can use it against the production system.

jisqyv:
@halatmit @afmckinney Also note. This PR is against master, but in reality when it is ready we will merge it into UCR as it is a component change (though no changes to YaVersion is needed).

jisqyv:
@halatmit @afmckinney You can download a compiled companion at:

  http://dl.qyv.me/MITAI2Companion.apk

You can use this with the current system.

jisqyv:
@afmckinney I believe this one is ready for review...

jisqyv:
Note: This change should be put on the ucr branch.

afmckinney:
LTGM

PR #751, opened 2016-04-14
Cdk

hendriwinata:


josmas:
This seems to be a merge into master; the code is not ready to do that yet. Closing for now.

jisqyv:
We will soon be merging in the cdk, but we will squash it to one commit to maintain linear history.

PR #759, opened 2016-04-29
Update Pedometer Implementation

jisqyv:
Update pedometer implementation to use the accelerometer instead of
GPS. Move from the “INTERNAL” (hidden) drawer to the Sensors drawer


jisqyv:
Note: Another commit is coming which updates the documentation.

PR #763, opened 2016-05-03
Media Drag and Drop

aaosuarez:
6.UAP Project: Media Usability 
Implemented Moxieapp's Uploader to allow users to drag and drop files when uploading media assets to their project.
@halatmit @afmckinney @jisqyv

### DIFF
--- a/appinventor/appengine/war/Ya.css
+++ b/appinventor/appengine/war/Ya.css
@@ -1030,6 +1030,11 @@ select.ode-PropertyEditor[disabled] {
   color: #8fc202;
 }
 
+.ode-ComponentHelpPopup > .popupContent > table {
+  width: 100%;
+  word-break: break-word;
+}
+
 .compiler-ErrorMarker {
   color: red;
 }
--- a/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
@@ -1536,10 +1536,18 @@ public interface OdeMessages extends Messages {
   @Description("Caption for file upload wizard.")
   String fileUploadWizardCaption();
 
+  @DefaultMessage("Error: Malformed Filename")
+  @Description("Error message when file name contains characters that would require URL encoding.")
+  String malformedFilenameTitle();
+
   @DefaultMessage("File names can contain only unaccented letters, numbers, and the characters " +
       "\"-\", \"_\", \".\", \"!\", \"~\", \"*\", \"(\", and \")\"")
   @Description("Error message when file name contains characters that would require URL encoding.")
   String malformedFilename();
+
+  @DefaultMessage("Error: Bad Filename Size")
+  @Description("Error message when filenames are 0 or 101+ characters long")
+  String filenameBadSizeTitle();
 
   @DefaultMessage("File names must be between 1 and 100 characters.")
   @Description("Error message when filenames are 0 or 101+ characters long")
@@ -1557,9 +1565,26 @@ public interface OdeMessages extends Messages {
   @Description("Error message reported when a file couldn't be uploaded because of its size.")
   String fileTooLargeError();
 
+  @DefaultMessage("Error: No File Selected")
+  @Description("Error message reported when a file was not selected.")
+  String noFileSelectedTitle();
+
   @DefaultMessage("Please select a file to upload.")
   @Description("Error message reported when a file was not selected.")
   String noFileSelected();
+
+  @DefaultMessage("Error: Cannot upload .aia file as media asset")
+  @Description("Error message when user tries to upload aia file as media asset")
+  String aiaMediaAssetTitle();
+
+  @DefaultMessage("To use this file, click Projects > Import project (.aia) from" +
+    " my computer ...")
+  @Description("Error message when user tries to upload aia file as media asset")
+  String aiaMediaAsset();
+
+  @DefaultMessage("http://appinventor.mit.edu/explore/ai2/share.html")
+  @Description("URL for more info on using aia files properly")
+  String aiaMediaAssetHelp();
 
   @DefaultMessage("Request to save {1}" +
       "\n\nA file named {0} already exists in this project." +
--- a/appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockCheckBox.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockCheckBox.java
@@ -101,12 +101,22 @@ public final class MockCheckBox extends MockWrapper {
     updatePreferredSize();
   }
 
+  @Override
+  int getHeightHint() {
+    int hint = super.getHeightHint();
+    if (hint == MockVisibleComponent.LENGTH_PREFERRED) {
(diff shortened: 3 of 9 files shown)


appinventor/appengine/build.xml
```diff
@@ -379,6 +379,7 @@
         <pathelement location="${lib.dir}/guava/guava-18.0.jar" />
         <pathelement location="${lib.dir}/gwt_dragdrop/gwt-dnd-3.2.3.jar" />
         <pathelement location="${lib.dir}/gwt_incubator/gwt-incubator-20101117-r1766.jar" />
+        <pathelement location="${lib.dir}/moxieapps/org.moxieapps.gwt.uploader-1.1.0.jar" />
```
jisqyv: What is "moxieapps" (I just don't know). Is it being maintained? Is source available? Will this cause us issues later when we need/want to upgrade GWT?

appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/AssetList.java
```diff
@@ -63,7 +64,7 @@ public AssetList() {
       @Override
       public void onClick(ClickEvent event) {
         if (assetsFolder != null) {
-          FileUploadWizard uploader = new FileUploadWizard(assetsFolder);
+          DNDFileUploadWizard uploader = new DNDFileUploadWizard(assetsFolder);
```
jisqyv: Is FileUploadWizard still in use? Where? If not, shouldn't it be removed by this change?

appinventor/appengine/src/com/google/appinventor/client/wizards/DNDFileUploadWizard.java
```diff
@@ -0,0 +1,305 @@
...
+import com.google.appinventor.shared.rpc.project.FileNode;
+import com.google.appinventor.shared.rpc.project.FolderNode;
+import com.google.appinventor.shared.rpc.project.ProjectNode;
+import com.google.appinventor.shared.rpc.project.youngandroid.YoungAndroidAssetNode;
+import com.google.appinventor.shared.rpc.project.youngandroid.YoungAndroidAssetsFolder;
+import com.google.gwt.core.client.GWT;
+import com.google.gwt.event.dom.client.*;
+import com.google.gwt.user.client.Command;
+import com.google.gwt.user.client.Window;
+import com.google.gwt.user.client.ui.*;
```
jisqyv: Please use explicit imports, none with "*"

jisqyv:
@a-suarez Please answer line comments...

appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/AssetList.java
a-suarez: Yes, the normal FileUploadWizard is used to upload projects and in the Properties panel.

appinventor/appengine/build.xml
a-suarez: It's from Moxie Group (http://www.moxiegroup.com/moxieapps/gwt-uploader/). The Uploader is open-source, released in 2012 and updated then. I used it because it makes handling files simpler and can be built upon. Do you think having old code is a big enough issue to warrant holding off on this feature?

appinventor/appengine/src/com/google/appinventor/client/wizards/DNDFileUploadWizard.java
a-suarez: Will do!

appinventor/appengine/build.xml
jisqyv: I'm concerned that a version of GWT in the future will be incompatible with this code. It is already 4 years old. I would rather not have this functionality, then have it and then have to remove it later.

halatmit: Is there an alternative library?



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

a-suarez: I've found other GWT libraries (e.g. https://github.com/fredsa/gwt-dnd) but I used the Moxie uploader because it had clearer documentation and seemed to have more capability for future work. At this point, I'm starting my usability tests today so I can't switch this out. Would it be okay to test the impact of drag and drop to see if it is worth implementing with another library or from scratch?

jisqyv:
Andrew and I have discussed this and we'll go with the moxieapps library. I just need to do some testing and I'll give this a thumbs up once I complete my testing.

afmckinney:
Tested but need to handle a few conner cases I'll fix and move to gerrit.

afmckinney:
Will look at again and make decision. Worried about dependency on non maintained JS package.

appinventor/appengine/war/Ya.css
```diff
@@ -2301,3 +2301,28 @@ textarea::-webkit-input-placeholder {
   height: 0;
   clear: both;
 }
+.
```
mjgallag: @aaosuarez i'm guessing this period is a typo?

aaosuarez: Yes, I'd say so.
Good catch! Not sure how that got in.

ewpatton:
I have a version that uses the native HTML5 drag and drop support that supersedes this implementation.

moliata:
@ewpatton this should be left working and merged. Only a few browsers has HTML5 drag and drop support, this PR is better + already tested by millions of users, using distro's like Makeroid/Thunkable, etc...

ewpatton:
@moliata 
First, [according to MDN](https://developer.mozilla.org/en-US/docs/Web/Events/drop), HTML5 drag and drop has been supported in Chrome since verrsion 4, Firefox since version 3.5, and Safari since version 3.1. It's even supported in IE 10 and Opera 12, so I must respectfully disagree that only a few browsers support it.

Second, based on our monthly numbers and assuming that (1) every App Inventor derivative had the monthly numbers we have, (2) every one of those users was distinct on each platform, and (3) everyone used this feature if available, I would accept your claim that it has been tested by millions of people. However, I find all of these premises to be highly unlikely.

My implementation encompasses more behavior, such as drag and drop support for projects and extensions. It is also done in pure JavaScript so that we can integrate it with future versions of App Inventor that rely less on GWT for UI.

moliata:
@ewpatton I agree with point that, it would be greater to rely on GWT for UI less, but we need to make some kind of a workaround for IE 7-10 version browsers. As of my research, a lot of users are still using just IE 8, that would be a headache for them, even my neighbor is using IE 8 still. Best regards.

ewpatton:
App Inventor's policy has been that we do not support IE, so that's not a dealbreaker for us.

moliata:
@ewpatton never heard of that.

kkashi01:
@ewpatton  If your version supports multi-file, that'll be great. We have incorporated this into AppyBuilder, but it only allows for single-file add.

ewpatton:
It does not at the moment due to the fact that projects, keystores, media, and extensions all have different code paths for use load, and those code paths all accept a single item. We would need to only allow multi file if and only if all files are the same category. I will also need to add some RPCs that allow multi file uploads. It is definitely doable and I can incorporate that into the next iteration.

PR #774, opened 2016-05-19
Update build.xml

GodUseVPN:
The default childProcessRamMb set to 2048MB is too large for 32bit OS.


halatmit:
How thoroughly has this been tested?    What are the implications building large projects?  Are there projects that could be built now that cannot be built after the change?

GodUseVPN:
The change of build.xml only affect to building aiplayapp from dev machine.1GB RAM is enough.

jisqyv:
LGTM -- moved to Gerrit for mering

jisqyv:
MERGED

PR #775, opened 2016-05-19
Update Ode.java

GodUseVPN:
It's more friendly to display username instead of email

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/Ode.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/Ode.java
@@ -729,7 +729,7 @@ public class Ode implements EntryPoint {
         // Initialize UI
         initializeUi();
 
-        topPanel.showUserEmail(user.getUserEmail());
+        topPanel.showUserEmail(user.getUserName());
       }
 
       @Override


jisqyv:
@GodUseVPN I don't believe this change will work. Unless the Gallery is enabled (which it is only on our public instance) we do not know the user name, so we don't have anything to display. The standalone code does not implement the Gallery (and will require significant work to include it).

-Jeff

GodUseVPN:
Well,I have partially implemented GAE API used by the public instance, and I use QQ OAuth as authenticate module.The name field is used to display the QQ nickname,email is filled with uuid.So the name field is userfull for me,but it may be useless for others...

jisqyv:
@GodUseVPN Are you adding support so that people can use their QQ account to login to App Inventor? If so, can you submit a PR to add that authentication code to App Inventor?

GodUseVPN:
Not adding,I just replace the locale auth (login.jsp) with QQ auth.It may be useless outside China.
POC implement: http://app.wxbit.com/

jisqyv:
@GodUseVPN Works fine here. I signed up for a QQ account, downloaded the App to my phone and was able to login to your site. Is your site based on the stand alone code, or something else. I noticed that you had the Gallery present, and we don't yet have it working in the stand alone code...

GodUseVPN:
It's another stand alone implementation.I think implementing the GAE API used by AI is a better way then developing a stand alone branch.But the GAE Implementation like AppScale is too complicated.So, I check out the master branch and partially implement the GAE API for AI.Now, I have implemented the objectify and gcs api,not memcache, images and search api yet.

jisqyv:
Very interesting. When I wrote the stand alone code, most of the work was replacing the ObjectifyStorageIo module with an equivalent (LocalStorageIo) that also implemented the StorageIo interface.

I have a new version I am working on that is horizontally scaleable, you can add more server to handle more users. It uses a key/value store which is also redundant and scaleable. It became feature complete over the weekend.

GodUseVPN:
My implementation based on MySQL,almost make no change to the master code,except some "bugs" fixed,I'm not sure if they are bugs.

halatmit:
We believe this has been taken care of.

PR #777, opened 2016-05-21
Udoo

sirhcbuilt15:


jisqyv:
I believe this component would be a good candidate for an extension. We are on the verge of releasing our extension implementation!

jisqyv:
Actually, I just took a deeper look and I believe this code cannot be done as an extension at this time.

halatmit:
We need to think about how to handle this.   It's too big to add to the main system (which would impose overhead on all App Inventor users).   But it's also to complex for an extension.   We need to come up with the right way for this code to relate to App Inventor and not put the maintenance obligation on MIT.

josmas:
I am a bit puzzled about this pull request. @fmntf is it coming from you? or was it opened by someone else by mistake?

fmntf:
Hi! No, I did not open this pull request (@sirhcbuilt15 did)!

fmntf:
A few more things:
 * @jisqyv is correct, this code cannot be bundled as extension. For instance, it modifies Compiler.java adding more permissions (USB; ADK)
 * The code should be definitely rebased and reviewed

Please cancel this pull request (as code author, not pull request author). Then, if we could discuss a way to integrate it somehow in App Inventor (after the proper cleanup and reviewed) it would be great!

jisqyv:
@fmntf I'll close this PR. As Hal said, we need to figure out how to incorporate code such as this. Thanks for your work on it!

PR #784, opened 2016-05-26
Update CrashReport.java

GodUseVPN:
may caused nested RuntimeException

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/server/CrashReport.java
+++ b/appinventor/appengine/src/com/google/appinventor/server/CrashReport.java
@@ -42,25 +42,11 @@ public final class CrashReport {
    */
   public static RuntimeException createAndLogError(Logger log, HttpServletRequest req,
       String extraInfo, Throwable exception) {
-    return createAndLogError(log, req, extraInfo, new RuntimeException(exception));
-  }
-
-  /**
-   * Logs an error.
-   *
-   * @param log  logger for to log error
-   * @param req HTTP request (may be {@code null})
-   * @param extraInfo  additional information about the context of the crash
-   *                   (may be {@code null}), preferably formatted as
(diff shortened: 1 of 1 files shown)


jisqyv:
@GodUseVPN Have you seen this happen? Java Polymorphism should take care of calling the version that accepts a RuntimeException when createAndLogError is called with a RuntimeException. The more general version, which accepts a Throwable, should only be called if the exception is not an instance of RuntimeException (or one of its subclasses).

jisqyv:
Here is a sample program:

```java
public class test {
    public static void main(String [] argv) {
      z2 t = new z2();
      test1(t);
    }

    public static class z1 {
        int frob;
    }

    public static class z2 extends z1 {
        int frob1;
    }

    public static void test1(z1 arg) {
      System.out.println("test1 called, z1 is the class of the arg");
    }

    public static void test1(z2 arg) {
      System.out.println("test1 called, z2 is the class of the arg");
    }
}
```

When run, it outputs: "test1 called, z2 is the class of the arg" which indicates that the more specific version is called. The same thing "should" happen with createAndLogError.

GodUseVPN:
public class Test {
	
	public static void main(String[] args) {
		Exception e = new RuntimeException();
		print(e);
	}
	
	private static void print(RuntimeException e){
		System.out.println("1");
	}
	
	private static void print(Exception e){
		System.out.println("2");
	}
}

print "2"

Somewhere such as GalleryServlet pass an Exception type  to CrashReport 

1 pic

jisqyv:
@GodUseVPN I see. You have a point. Let me do my normal testing. I'll likely merge this today (soon).

jisqyv:
LGTM

jisqyv:
@GodUseVPN MERGED: Note: I re-indented your code to confirm to our normal indentation standard.

PR #785, opened 2016-05-26
Adding new property label format - html

jisqyv:
Author: Manish Khanchandani <naveenkhanchandani@gmail.com>
Author: Jeffrey I. Schiller <jis@mit.edu>

Change-Id: Icc8a4ab100be3c4c706051ace390bcf80fe33caa


jisqyv:
@afmckinney @halatmit This change replaces PR #99 (which I will now close). This code takes #99 and adds the changes needed to MockLabel so that HTML displays in the Designer as well as on the device. There is a catch. The Designer displays full HTML but the device can only display a limited subset (the same subset as is used in the Notifier component). So you cannot put in tables and things like that. I have tested it, etc.

afmckinney:
LGTM

jisqyv:
MERGED

PR #687, opened 2016-01-27
feature description

toropeza:
This commit allows moderators to add a description for apps that are featured. Let me know if there are recommendations for the design.

A working version can be found at http://fdesc.aigallerythomas.appspot.com/
Changes made:
-New styling on Featured app list
screen shot 2016-02-09 at 3 37 58 pm

-Featured description shows on GalleryPage
screen shot 2016-01-26 at 4 57 09 pm
-Popup panel for adding the description
screen shot 2016-01-26 at 4 57 24 pm

Thank you,
Thomas Oropeza

PR #798, opened 2016-06-09
Feature description

toropeza:
[Live Version](https://fdesc-dot-aigallerythomas.appspot.com/#5645784439586816)
Refer to https://github.com/mit-cml/appinventor-sources/pull/687 also.
What changed:
1. Wont crash with apps that do not have a description. (Like those that are currently featured)
2. three cards per row instead of two
3. different border
4. feature description no longer shows on the app view page
5. character limit for featured description
6. removed meta-data from feature apps. (cleaner look)
7. updated with current Master

screen shot 2016-06-09 at 10 43 16 am
screen shot 2016-06-09 at 10 43 22 am
screen shot 2016-06-09 at 10 43 48 am

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/GalleryGuiFactory.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/GalleryGuiFactory.java
@@ -38,6 +38,8 @@ public class GalleryGuiFactory implements GalleryRequestListener {
   private final String NUM_VIEW_ICON_URL = "/images/numView.png";
   private final String NUM_COMMENT_ICON_URL = "/images/numComment.png";
 
+
+
   /**
    * Generates a new GalleryGuiFactory instance.
    */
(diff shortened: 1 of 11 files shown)


toropeza:
@wolberd @jisqyv

jisqyv:
@toropeza You should make sure that the description field of the GalleryFeatureData is unindexed. Otherwise App Engine builds an index on it, which takes space and costs $$. Btw. It appears that a lot of the fields in the Gallery storage are marked as indexable, when they shouldn't be. Take a look at ObjectifyStorageIo to see how to make fields and/or whole objects unindexed. Also check out the Objectify documentation itself.

Also. Featured Apps should "open" (go to the Gallery App Info) when they are clicked anywhere within the gold border. Right now there is a lot "dead" space that when clicked on does nothing, and there is no hint that you need to click on the images or app name. It probably just makes sense to have a click anywhere in the Gold/Yellow bordered area bring you to the app.

-Jeff

toropeza:
@jisqyv Thanks Jeff, i'll look through all of the Data classes to try to clean those out and make changes to the featured app cards.

afmckinney:
Need to revisit Gallery.

PR #799, opened 2016-06-12
Make session timeout a fatal error

jisqyv:
Session timeouts can happen if a user is idle too long or when we
perform certain types of App Inventor updates. Without this change, the
standard red error bar is produced. However many users appear to ignore
this error and keep working, but this work will never be saved!

This patch instead provides a dialog box that requires them to reload
their browser. The reload should not only refresh their session, but
should load any new App Inventor client side code. There is a risk that
up to 30 seconds of work may be lost when this dialog appears (most
people might lose only 5 seconds, but the risk is up to 30 due to the
way the auto-save code operates).

Change-Id: I4301e6dbdef0c52681b90ede4b8d84d6f9445d97


jisqyv:
@afmckinney Please review. Not only do I want to put this in master, but I'll cherry-pick it onto the current release branch and deploy it hopefully on Monday. This will facilitate the component release on Wednesday by providing a way for people to reload into the new code instead of getting "red bar" errors after the release.

PR #808, opened 2016-07-01
Fixed the slash issue for Windows users

chenglonkhin:
Change "SLASH" to "/", which enables Window users to build the local instance of App Inventor successfully. 

@jisqyv


jisqyv:
@chenglonkhin MERGED into Master (and ucr)

Issue #624, opened 2015-12-02
Comment text is not generated in "Download blocks as image"


halatmit:
The function "Download Blocks as Image" does show the comment blocks, but they are empty.

shrutirij:
I think it's because the text isn't part of the workspace svg.

I'll look at how to fix this.

halatmit:
Thanks



== Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.
MIT Dept. of Elec. Eng. and Comp. Sci.

josmas:
heya @shrutirij let me know if you've no time to work on this and can be assigned to someone else.

shrutirij:
@josmas I've been swamped, so I'll work on this over this weekend. Is that ok?

josmas:
Heya @shrutirij sure, not a problem. Just let me know if you can't get to it in the next week or so and I'll open it up for grabs. Thanks for looking into it anyway; hope you are keeping well! :)

thejdeep:
Would love to work on this. Please help me proceed. Thanks.

wadleo:
Hi @josmas, @halatmit please can more information be provided for this task, thanks.

abhijit5893:
I have been playing around with this for a while and I think I have found out the issue.  In /GSOC/appinventor-sources/appinventor/blocklyeditor/src/exportBlocksImage.js (line 73) there is a function call "cloneNode". cloneNode is an inbuilt function and  a bug regarding textarea content not being cloned has been [reported](https://bugs.chromium.org/p/chromium/issues/detail?id=487352). So if we update the code with the latest version, the bug should be fixed. No change in code necessary.

chenglonkhin:
Has anyone made any progress on this yet? It seems like there was a fix, but hasn't been implemented.

halatmit:
I don't think anyone else worked on this.  See what you can do.

jisqyv:
This issue was fixed by #810 which is now in the master branch

PR #810, opened 2016-07-07
Fixed the Issue #624 (Download Blocks as Images, Comment Text Disappears)

chenglonkhin:
#624 

When a comment block is created and the user edits the comment block by typing text, this action changes the value property of the textarea, which is not duplicated using the .CloneNode() method. This is the reason why comment text disappears when a user attempts to download blocks as images. 

@halatmit

### DIFF
--- a/appinventor/blocklyeditor/src/exportBlocksImage.js
+++ b/appinventor/blocklyeditor/src/exportBlocksImage.js
@@ -70,6 +70,14 @@ out$.svgAsDataUri = function(el, optmetrics, options, cb) {
   options.scale = options.scale || 1;
   var xmlns = "http://www.w3.org/2000/xmlns/";  
   var outer = document.createElement("div");
+
+  var textAreas = document.getElementsByTagName("textarea");
+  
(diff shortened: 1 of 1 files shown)


halatmit:
@jisqyv @chenglonkhin @afmckinney 
This looks OK to me.   Except that when I start a new project with this instance, I get the warning:
This project was saved with a newer version of the App Inventor system. We will attempt to load the project, but there may be compatibility issues.

Is this a result of not updating YaVersion or something?  Lokhin, please check with Jeff or Andrew.

chenglonkhin:
@halatmit 
I can't seem to reproduce this warning on my local instance, but I'll make sure with Jeff or Andrew when they come into lab today.

jisqyv:
@halatmit I'm doing this review today. I'll make sure all is well...

jisqyv:
@chenglonkhin @halatmit LGTM, moving to Gerrit

jisqyv:
@chenglonkhin MERGED!

Issue #800, opened 2016-06-14
Calling activity starter with no info gives a permissions error


halatmit:
If you call the activity starter and do not provide and information it looks like the intent resolution mechanism picks an activity to launch that leads to a permissions error.

See
https://groups.google.com/forum/#!topic/mitappinventortest/-e4mNygdEoQ

José suggests:

When using Implicit intents[1], the system figures out what you are trying to do from the information provided. 'No information' seems to point to that particular one, as Hal mentioned, no idea why, but the 'FIX' here may be validating the data before running the intent, instead of catching an error. In the method 'buildActivityIntent', redirect to a Form.dispatchErrorOccurred if there's no info about the action.

halatmit:
See also the original forum thread

https://groups.google.com/forum/#!topic/mitappinventortest/tq0IysLiIs4

jisqyv:
#816 is merged

PR #816, opened 2016-07-14
Fix for Issue #800

chenglonkhin:
@halatmit 

#800 

When ActivityStarter is called without any action information, the intent is pointed to an action that requires permission and is denied that permission. It results in a runtime error which gives little information for users. 

106c0560-b197-4b24-8d52-270984f6a5f9

I added another error message that replaces this runtime error to inform the user that no information was found in ActivityStarter. It seems that the default action is android.intent.action.MAIN if no information was enter in the Action properties for the ActivityStarter.

[AIA](https://drive.google.com/file/d/0B9cmuy8O4JxKT1gyNmxnOXJhbVU/view?usp=sharing) and[ APK](https://drive.google.com/file/d/0B9cmuy8O4JxKUTY4aTNOMjloMlk/view?usp=sharing) to reproduce error.

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/ActivityStarter.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/ActivityStarter.java
@@ -436,19 +436,28 @@ public class ActivityStarter extends AndroidNonvisibleComponent
       requestCode = form.registerForActivityResult(this);
     }
 
-    try {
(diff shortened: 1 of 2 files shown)


jisqyv:
LGTM

jisqyv:
@chenglonkhin MERGED into UCR

Issue #817, opened 2016-07-16
Change divide-by-zero to signal an error rather than return Nan


halatmit:
This is an instance of "Scheme hubris" is fell for when I implemented App Inventor arithmetic.   I wanted to "show off" and follow the Scheme convention of returning Nan, since this was part of the well-thought-out Scheme and Common Lisp arithmetic.   But it's the wrong thing for App Inventor

halatmit:
PR #818

halatmit:
#818 was closed

PR #818, opened 2016-07-17
Change division so that dividing by 0 signals an error rather than re…

halatmit:
…turning NaN

### DIFF
--- a/appinventor/buildserver/src/com/google/appinventor/buildserver/resources/runtime.scm
+++ b/appinventor/buildserver/src/com/google/appinventor/buildserver/resources/runtime.scm
@@ -1457,6 +1457,10 @@
       (max lowest (min x highest)))))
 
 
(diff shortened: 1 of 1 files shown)


halatmit:
Issue #817

jisqyv:
@halatmit I'm not sure I agree with this change. With it, a divide by 0 error will crash a packaged application. There is no way to catch a RuntimeException. It would be better to create an error that can be caught in Form and handled by the user.

halatmit:
Good point.   I wasn't thinking.   I'll see if I can fix that.



== Hal

Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

halatmit:
@jisqyv @afmckinney I messed something up and confused myself.  I think this PR is obsoleted by #819

jisqyv:
Replaced with #819

PR #836, opened 2016-08-19
Improve Project List Loading time

jisqyv:
Generate the user’s project list by doing one bulk API call to to the
datastore to fetch project meta data.

Change-Id: Ie3a93126b22e2292bc4b270d4d54c5c382880cf9

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/server/storage/ObjectifyStorageIo.java
+++ b/appinventor/appengine/src/com/google/appinventor/server/storage/ObjectifyStorageIo.java
@@ -54,6 +54,7 @@ import com.google.appinventor.shared.storage.StorageUtil;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
+import com.google.common.collect.Lists;
 import com.google.common.io.ByteSource;
 import com.google.common.io.ByteStreams;
 
@@ -84,6 +85,7 @@ import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.ConcurrentModificationException;
 import java.util.List;
+import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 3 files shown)


jisqyv:
@halatmit @afmckinney @ewpatton Please Review

jisqyv:
@halatmit This change should speed things up, let people have more projects and save us $$

jisqyv:
@halatmit @ewpatton @afmckinney This code is now live on ai2-test.appinventor.mit.edu

appinventor/appengine/src/com/google/appinventor/server/storage/ObjectifyStorageIo.java
```diff
@@ -944,6 +946,39 @@ public void run(Objectify datastore) {
...
+            projectDatas.t = null;
+          }
+        }
+      }, false);
+    } catch (ObjectifyException e) {
+      throw CrashReport.createAndLogError(LOG, null,
+          collectUserErrorInfo(userId), e);
+    }
+    if (projectDatas.t == null) {
+      return null;
```
josmas: would it make sense to return an empty list instead of null?

kkashi01: +1 on returning empty list otherwise, may cause NPE if caller not handling null

halatmit: +1 on returning empty list

ewpatton: +1 as well. This changes the contract behavior relative to the lines removed from [ProjectServiceImpl.java](https://github.com/mit-cml/appinventor-sources/pull/836/files#diff-dacdac42ba5eb0de1b23d5ec187829daL243) that would return an empty list on failure.

jisqyv:
@halatmit @ewpatton @kkashi01 @josmas Actually, turning a null pointer into an empty list is probably not the correct thing to do. The Objectify code will return an empty Map if the person has no projects. I'm not sure if it ever can return a null pointer, but if it does, it is an indication that something is broken, not that the person has no projects.

So if something is broken, we do not want to turn it into an (error free) empty list. If we did, the end user would see no error, but would believe that all their projects are gone, which I believe is much scarier then an error!

That said, I'm not sure the client code properly handles an error in the project list call. We know that when we get a DeadlineLimitExceeded error (which we cannot prevent) the client is getting a 500 response. However this either isn't getting through to the user, or it is being obscured by the survey dialog (which is triggered because the project list code failed). I'm going to look into that. So expect some more commits here soon, both to deal with the null pointer better (likely by calling CrashReporter) as well as any fixes I believe should go on the client side.

-Jeff

halatmit:
@jisqyv sounds good.   thanks for being careful.

jisqyv:
@halatmit @afmckinney @josmas @ewpatton I just pushed another commit which addresses error handling better. If someone gives is an LGTM, I'll merge the commits together and merge them through Gerrit.

I have also determined the cause of Abraham seeing the survey splash repeatedly. I'll address that in a separate PR, as it is a different problem, triggered by this one.

-Jeff

halatmit:
LGTM on throwing the runtime exception

jisqyv:
@halatmit is your LGTM on the change, or just the throwing the runtime exception part?

halatmit:
Just the exception

halatmit:
@jisqyv LGTM for the entire PR.

jisqyv:
Latest commit is just the previous two squashed together. This is the commit on its way to Gerrit

halatmit:
LGTM

jisqyv:
MERGED

PR #837, opened 2016-08-21
Do not erase UserSettings when App Inventor Closed

jisqyv:
There is a race condition where if App Inventor is closed (browser
window closed) while the system is still initializing, the user’s global
settings will be over-written by an empty settings object because the
global settings have not finished loading. We detect that with this code
and avoid over-writing the settings.

Change-Id: Icfcf8204b860984bf36e2a7d525ebf5814a2ffd8

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/settings/user/UserSettings.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/settings/user/UserSettings.java
@@ -23,6 +23,7 @@ import com.google.gwt.user.client.DeferredCommand;
  */
 public final class UserSettings extends CommonSettings implements SettingsAccessProvider {
   private boolean loading;
+  private boolean loaded;
 
   /**
    * Creates new user settings object.
@@ -45,6 +46,7 @@ public final class UserSettings extends CommonSettings implements SettingsAccess
             OdeLog.log("Loaded global settings: " + result);
             decodeSettings(result);
 
+            loaded = true;
             loading = false;
           }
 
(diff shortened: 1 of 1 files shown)


jisqyv:
@halatmit @afmckinney @ewpatton Please Review

How this code was tested

I tested this code by adding a patch to ObjectifyStorageIo.getUserProjects to add a 5 second delay. This simulates a lot of projects to load. I also added my normal debug code (jishack) and turned on AppInventorFeatures.showSurveySplash.

I then started App Inventor normally and saw the survey splash screen. I clicked the “Never” option. I then created a project (I just loaded HelloPurr from the “repository”).

I then reloaded my browser. Of course I sat for 5 seconds on the Projects List screen (with no projects displayed) while the timer ran. Eventually my project was listed and I was then sent to it. All as expected.

I then reloaded my browser again, but while sitting in the 5 second wait, I closed the browser. I then once again visited App Inventor (my development version) and this time I was again confronted with the SurveySplash (which I clicked away) and then after the 5 second delay I saw my project list. The “HelloPurr” project was *not* automatically opened because the last opened project setting was also lost along with the survey splash settings. This demonstrated and reproduced the problem.

I then added the code in this PR and redid the above experiment and I never was the splash survey after the first time. The code works. I also tested the new user case to make sure that new users have their settings properly saved as well.

You can easily reproduce what I did:

Checkout https://github.com/jisqyv/appinventor-sources/tree/fixsavesettings-debug-before this is the debugging code without this change. Do the steps above. You should see the problem.

Then checkout https://github.com/jisqyv/appinventor-sources/tree/fixsavesettings-debug this is the debugging code on top of this change. Do the steps above. You should *not* see the problme.

Enjoy.

-Jeff

jisqyv:
@halatmit @ewpatton @afmckinney Note: My debugging code (jishack) incorporated in the testing above is set for only Chrome. You will have to edit it to include Firefox if you want to test with Firefox.

afmckinney:
LGTM

PR #856, opened 2016-11-25
Json string

halatmit:
Changes the way lists are printed, so they appear in Json notation.   Implements the ShowListsAsLisp property.
Still need to change runtime.scm to make use-json-format a macro.    Also need to implement the upgrader.

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
@@ -3019,6 +3019,10 @@ public interface OdeMessages extends Messages {
   @DefaultMessage("Sizing")
   @Description("")
   String SizingProperties();
+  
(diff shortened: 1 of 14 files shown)


halatmit:
I updated the PR to comment out the upgrade code and the designer property.  This should now be ready for review.

jisqyv:
@halatmit It doesn't work properly. Basically the ShowOldStyle is hidden, but it's default value is "false" so lists now only appear as JSON. We need a solution where ShowOldStyle is hidden *and* lists appear in the Companion in lisp syntax. We can discuss when we meet in person... this is tricky

halatmit:
Thanks for catching that.   We can discuss when we meet.


== Hal

Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
@halatmit @afmckinney A version of this change has been pushed onto the ucr branch.

PR #870, opened 2017-02-13
Update OdeMessages_es_ES.properties

jodiacer:
PitchProperties and RollProperties belong to the orientation sensor and have been wrongly translated to Tono and Lanzar in Spanish, which makes no sense in this context. They are related to orientation sensor angles, which are in turn related to navigation angles used in aircraft.

Pitch could be translated as "Tono" when related to sound, but never when relating to movement or navigation. In this case, it should be translated as "cabeceo" (meaning movement of an aircraft nose up and down, i.e. pitch) . 

Roll has been translated as "lanzar", as in "roll a dice" --> "lanzar un dado", but this is wrong in the current context. The right translation is "alabeo" (which means moving one wing up and the other one down, i.e. roll).


josmas:
I've never seen the word _alabeo_ being used in this context. A quick search shows _balance_ as an alternative, but I am not certain which one would be better.

I do agree on the pitch translation error though, but the property is currently used in two components. It makes sense as it is in TextToSpeech (_tono_ would be correct), so the right fix to merge here would be to create a separate _PitchProperties_ for the OrientationSensor instead of reusing the TextToSpeech one (and also agree on a correct translation for Roll).

josmas:
Oh! and thanks for the PR! :)

jodiacer:
Hi José!

Thank you for your really fast response :)

I didn't notice that Pitch was also present in the TextToSpeech component.
In this case, I agree with you, the best option would be to separate both,
because we need different translations. Of course, this will have an impact
on other files, but unfortunately, I have no idea on which ones :(

With respect to "alabeo", I think it is more accurate than "balance" when
talking about navigation angles. Navigation angles come from aircraft
navigatiion, where "alabeo" means exactly the same as roll. At least in
Spain, this is the right term, although I am not sure whether a different
term is used in Latin America.

I know that Wikipedia has often mistakes, but in this case, I think it is
right. I am sending you two links to the English and Spanish version of the
same entry (Flight control surfaces / Mandos de vuelo). There you can see
the correspondence between pitch and "cabeceo" and roll and "alabeo":

https://en.wikipedia.org/wiki/Flight_control_surfaces
https://es.wikipedia.org/wiki/Mandos_de_vuelo

Anyway, if you finally use "balance", it would still be much better than
the current translation, so no problem on my side.
I would normally use this term in relation to left and right loudspeakers,
but it has the same sense left vs right that we are looking for here.

Thanks again!

Best regards,


Pepe Díaz


2017-02-14 12:50 GMT+01:00 Jos <notifications@github.com>:

josmas:
Thanks @jodiacer I'll mark this as future work. It is important but not high priority at the moment.
Please feel free to investigate the sources a bit more if you want to make the suggested changes :)

jisqyv:
The correct thing to do here is to create two entries in OdeMessages for pitch. This isn't hard, and I can take care of it.

jodiacer:
This would be great!
Thank you!

jisqyv:
Yep, this would be a good starter for a GSoC student.

moliata:
Any progress?

SusanRatiLane:
Saving translation information for future fix for #2817

AppInventorWorkerBee:

jisqyv:

PR #871, opened 2017-02-25
Update _messages.js

conorshipp:
After the recent App Inventor 2 update, the backpack is now persistent, which means that when right-clicking on the backpack icon on the block editor and clicking "Show Backpack documentation", some of the text in the dialog is incorrect. I have corrected the text to show that of the release notes.


jisqyv:
LGTM -- Moving to Gerrit

jisqyv:
MERGED

PR #873, opened 2017-03-06
Add SearchHint Property to ListView and ListPicker

keving17:
In response to [this forum post](https://groups.google.com/forum/#!category-topic/mitappinventortest/rcLJaILwwb4), added SearchHint Property to ListView and ListPicker Components.


moliata:
MockListView commit is useless. Also, you should leave Search list... not Search List... Plus, this better would be called FilterBarHint.

PR #874, opened 2017-03-08
Add VersionName & VersionCode blocks

conorshipp:
I have added two blocks to return the VersionName and the VersionCode of users' app.

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/Form.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/Form.java
@@ -73,6 +73,14 @@ import com.google.appinventor.components.runtime.util.SdkLevel;
 import com.google.appinventor.components.runtime.util.ScreenDensityUtil;
 import com.google.appinventor.components.runtime.util.ViewUtil;
 
+import android.content.pm.PackageInfo;
+import android.content.pm.PackageManager;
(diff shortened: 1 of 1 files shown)


josmas:
@conorshipp could you explain when you'd want to use these two methods?

There are cleaner ways of writing the code also. You could define
```
String vname = "";
```

at the top of the method and then return it after the **try_catch** instead of having two returns.
The *catch* clauses (error handling) could do with a bit of thinking too.

josmas:
I believe (not certain) that if you change Form you have to update a few more things (such as [YaVersion](https://github.com/mit-cml/appinventor-sources/blob/master/appinventor/components/src/com/google/appinventor/components/common/YaVersion.java#L691) and upgrade code).

conorshipp:
> could you explain when you'd want to use these two methods?

If, for example, a user creates an app with a settings screen, they might want to show the version name at the bottom of the screen, similar to Google's apps.

> such as YaVersion and upgrade code
I'm still learning an don't really understand this.

josmas:
This [section of the docs](https://docs.google.com/document/d/1ukpZXcd0fIn1fV0cNp4J00wrfwwA1daswU1yY6dqqfs/pub#h.2g0jyctaq6go) may be helpful.

PR #875, opened 2017-03-08
Add new Toast

conorshipp:
I have added a new method called "Toast". This is different than the "Show Alert" method. It will show a 'toast' the bottom of the screen, above the device's navigation bar.

The duration can either be set to "0" or "1". 0 is Toast.LENGTH_SHORT and 1 is Toast.LENGTH_LONG. If the user sets the duration to a number that isn't 0 or 1, then it will automatically change to 0.

https://developer.android.com/guide/topics/ui/notifiers/toasts.html

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/Notifier.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/Notifier.java
@@ -492,4 +492,13 @@ public final class Notifier extends AndroidNonvisibleComponent implements Compon
   public void LogInfo(String message) {
     Log.i(LOG_TAG, message);
   }
+
+  @SimpleFunction(description="Show a toast message with duration")
+  public void Toast(String text, int duration) {
+    if(duration != 0 || duration != 1) {
+      Toast.makeText(getApplicationContext(), text, 0).show();
+    } else {
+      Toast.makeText(getApplicationContext(), text, duration).show();
+    }
+  }
 }


halatmit:
Instead of creating a new kind of thing (toast) that App Inventor users would need to know about, it would be better to add position control to Alerts, without losing the control over background color, text color, font size etc, that Alerts provide.   Also note the default text size for alerts, that is sensitive to the SDK level.   And also the use of XOffset and YOffset

A better approach would be to change line 443 in toastNow to provide gravity control based on a new property, and follow the pattern used for TextColor and BackgroundColor to implement this new property.   

Be aware that there are some tricky aspects to implementing the way that users would set the values of the property.  You don't what to just have the property be a number, but rather provide a dropdown of choices, similar to what's done with AlignVertical and AlignHorizontal for Screen and Vertical/Horizontal Arrangements.  Take a look at the way  AlignVertical or AlignHorizontal are implemented and mimic that for the setting the position of an alert. 

You'll also need to provide for internationalization of the names of the property and its value.

There are several files that will need to be changed.
Don't expect to get this right on the first try.   

I suggest that you do a first pass where the gravity property is just a number and get that working and let us review it.   Then go on the the more complex part of creating the dropdown selector for the values of the property.

conorshipp:
Thanks @halatmit for the feedback (and sorry for the late response!). I think that the gravity property is a good idea and if/when I get time I could take a look at this. But for now, I think I'm going to close this.

PR #881, opened 2017-04-11
Fix outdated Companion not triggering auto-update

Klomi:
On the MIT App Inventor server, the Companion (version 2.11ai2) included in aiStarter 2.3.0 can be automatically updated to the latest version, while on a local server built from source, it is failing to trigger an auto-update and instead prompting the user to install the latest Companion manually.

The Companion's JSON response is
`{"version" : "2.11ai2", "fingerprint" : "generic/google_sdk/generic/:2.2/FRF91/43546:eng/test-keys", "installer" : "Not Known"}`
where the package name is missing. Undefined package name is not accepted by the version checker so the function returns early. Further version checking and auto-update is not triggered. By adding `json["package"] != undefined` skips the package verification if the package name is not specified.


Klomi:
Added a commit.
`Blockly.Msg.REPL_OK` can be replaced by localization text and it is not necessarily `"OK"` in some languages.

Klomi:
Will split this pull request

PR #895, opened 2017-06-30
Fixed broken generic component method blocks

kvmathison:
There was an issue where some generic component method blocks did not have the
mutator labeling them as such. This was fixed by explicitly setting the
is_generic mutator along with the other mutators when these blocks are generated
in the drawer.

There was another issue where generic clock add blocks expected a time unit
dropdown which only exists on the instance version of the clock add blocks. This
was fixed by only setting the time unit if the block is not generic, since this
attribute is only used in the instance version of these blocks.


jisqyv:
LGTM -- Moving to Gerrit for merging

jisqyv:
MERGED

PR #914, opened 2017-08-18
Fix string join

halatmit:


appinventor/components/src/com/google/appinventor/components/runtime/util/JavaJoinListOfStrings.java
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Java implementation of join-strings since the Kawa version appears to run of space.
+ * See runtime.scm
+ *
+ * @author halabelson@google.com (Hal Abelson)
+ */
+public final class JavaJoinListOfStrings {
+
+private static final boolean DEBUG = true;
```
ewpatton: I think this should be false by default.

halatmit:
Evan .. Nice catch on the DEBUG flag.

ewpatton:
LGTM. I will move this over to Gerrit.

jisqyv:
@halatmit Btw. You might try your Scheme version, but use letrec instead of let. Apparently Kawa will only do tail calls with letrec.

halatmit:
Thanks, but it's fixed now with java. And when I did the tail-call version
in scheme, it looked like it was running out of memory, not stack, unlike
the recursive version, which got a stack overflow.


== Hal

Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.


On Sun, Aug 20, 2017 at 10:16 PM, Jeffrey Schiller <notifications@github.com

jisqyv:
@halatmit Hmm. I just did some testing with our version of kawa, and let works just fine. To prove this I wrote the factorial function recursively and fed it 20000 and it blew up on a stack overload. I then re-wrote it using tail recursion with normal let, just like you did here, and it worked fine. Was able to compute factorial of 40000 without a problem (I didn't try higher numbers because it was taking a while :-).

halatmit:
This particular error happens only on the emulator -- not on my phone.
And when it got the answer with Abraham's example, there were about 29
thousand lists it was trying to join.

Try Abraham's example on your phone.  I should work. Then try it on the
emulator.

== Hal

Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.


On Sun, Aug 20, 2017 at 10:24 PM, Jeffrey Schiller <notifications@github.com

ewpatton:
The issue here is that the space complexity of a factorial computation will be roughly constant versus the string joining which is linear in the length of the strings. 

My first reaction to this is that the tail recursive version is probably keeping twice as much data because rather than using something like a StringBuilder where a single object is increased in size, KAWA is creating an intermediate string to store the result and thus you will spend a lot more time garbage collecting by creating a number of one-off objects in Eden. If KAWA somehow holds the intermediate references you will end up with O(n^2) space complexity and likely an OOM situation. This might be mitigated by KAWA's FString type but I would need to check the sources to understand the interaction better. 

Evan

halatmit:
This sounds like a very reasonable guess about why it's running out of
space.
The issue is more likely to be the use of kawa string-append rather that
tail-recursion.
We can test this by using string-append! instead.

== Hal

Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

ewpatton:
Closed via e3d1a87

PR #929, opened 2017-09-15
Fix button height when font size is changed

jisqyv:
Make sure the button in the designer scales along with the font size.

This change is a simplification of work originally done by Olivier Midy
<osmidy@mit.edu>.

Change-Id: Iad2033001f26b102506f8bba2dff89f94f6b8ab6


moliata:
Great, works for me. This pull request should be merged, as it does fix the problem.

ewpatton:
Compared with the production version, the default button is 6px shorter with this branch. The dimensions in values.xml specify that the padding on the device should be 6px vertically and 4px horizontally. It may be worth adopting these values.

halatmit:
I agree we should be consistent.  Thanks.


== Hal

Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
@ewpatton Here is another approach. It scales properly and the default button is the same size as our current production version. I actually believe this is a more correct fix.

PR #934, opened 2017-09-22
Click event for Image component

moliata:
Hello there, I have added a Click event for Image component & I have updated YaVersion, docs.


conorshipp:
@moliata Since you've added a new method/event and incremented the version, you'll also need to write an upgrader.

See [this section](https://docs.google.com/document/d/1ukpZXcd0fIn1fV0cNp4J00wrfwwA1daswU1yY6dqqfs/pub#h.2g0jyctaq6go) of the docs:

PR #946, opened 2017-10-05
Adds AuthKey instructions to readme, and a note about Java 8

josmas:
A couple of changes in the README to specify Java version and adding the MakeAuthKey step for compiling, which seems to be missing.


moliata:
Great, in fact Jave 9 is out already!

Issue #915, opened 2017-08-20
investigate accelerometer behavior under rotation


halatmit:
Different device behave differently in reporting Xaccel and Yaccel when the device is rotated.  Make it easier to write programs that work across all devices and rotations, probably by using the Window Service RotationIndex.

halatmit:
Look here for how to get started.

https://stackoverflow.com/questions/5877780/orientation-from-android-accelerometer

halatmit:
This will be handled by PR #956 .

jisqyv:
Fixed by https://github.com/mit-cml/appinventor-sources/commit/0f01fbe50f55a613f8cdf2425eefb47dbf100f5b

PR #956, opened 2017-10-09
Change the accelerometer so that Xaccel and Yaccel behave on Landscap…

halatmit:
…ePrimary devices

the same way as they do on PortraitPrimary devices.
If you hold the device vetically (long edge vertical) then xAccel should be 0 and
yAccel should be -9.  Rotate 90 degree counterclock wise and xAccel,Yaccel should be 9,0.
This should be true on both PortraitPrimary and LandscapePrimary devices.

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/AccelerometerSensor.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/AccelerometerSensor.java
@@ -19,10 +19,14 @@ import com.google.appinventor.components.common.YaVersion;
 import com.google.appinventor.components.runtime.util.ErrorMessages;
 
 import android.content.Context;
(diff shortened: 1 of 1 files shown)


halatmit:
This is for the ucr branch since it changes the accelerometer component.   Note that pushing this to Hong Kong will require changing that the classes are using:  There will now be only one version of FindtheGold (the one they are currently using for PortraitPrimary devices:  not the LenovoTablet).

halatmit:
Will resolve #915 when it's merged.

jisqyv:
Merged into ucr (https://github.com/mit-cml/appinventor-sources/commit/0f01fbe50f55a613f8cdf2425eefb47dbf100f5b)

Issue #962, opened 2017-10-21
class name not match file path when DX of extensions


ColinTree:
i got a folder for an extension and the structure is like this:
OpenGL
|--objects
|　|--Drawable.java (required files)
|　|-- ...
|--GLCanvas.java (**this is the extensions file**)

after packed as an extension, it will all become this structure in AndroidRuntime.jar: (all file in a same folder even in the different package)
![image](https://user-images.githubusercontent.com/22613139/31850701-85546ce0-b689-11e7-9efd-8df761a9125f.png)

and then when packaging by build server, the DX-fail log is like shown in the picture

![image](https://user-images.githubusercontent.com/22613139/31850740-6321f394-b68a-11e7-85b7-c9e4c8a72509.png)

and the only solution now, is to pack the required files into a jar file, and then declear require library, but it is such a unnessary step if the AndroidRuntime.jar is well packaged.

ewpatton:
@ColinTree You may want to take a look at the [copyRelatedExternalClasses](https://github.com/mit-cml/appinventor-sources/blob/master/appinventor/components/src/com/google/appinventor/components/scripts/ExternalComponentGenerator.java#L300) method in ExternalComponentGenerator. Right now it only aggregates classes within the same package, not any descendant packages. You could create a patch that walks the package hierarchy to include other classes.

ColinTree:
@ewpatton 
I had fixed it locally.
But i m not sure which branch should i put it in?

ewpatton:
Since this doesn't materially affect the companion, you can put it onto master.

ColinTree:
Sure, here it is: #963

marksherman:
you leave this open until the PR is accepted.

PR #963, opened 2017-10-21
Fixed the problem: class name not match file path when DX of extensions

ColinTree:
Problem details: #962


ColinTree:
Notice that this is only tested on my Windows 10, i think we need some test on other operating systems.

jisqyv:
MERGED (https://github.com/mit-cml/appinventor-sources/commit/ccd4bc04898660d61897a984f83d580c7e87bbbd)

PR #1032, opened 2017-12-02
Travis Integration

barreeeiroo:
This allows to add a Continuous Integration service to AppInventor Sources
Commits are checked automatically if they build or not, without having to test them manually

---

Tested at [Makeroid](https://www.makeroid.io). We use it to test ALL our commits before merging. It really saves lot of time

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,11 @@ before_install:
 
 install: cd appinventor && ant MakeAuthKey
 
-script: ant
+script: ant noplay
 
 after_success: 
-  - ant PlayApp && sudo ant clean
+  - ant PlayApp
+  - sudo ant clean
 
 after_failure: sudo ant clean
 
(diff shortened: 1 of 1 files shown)


ewpatton:
@conorshipp's understanding is correct. We use Jenkins to validate any changes by running the unit tests. However, this is not the first step in the process because we are open source and anyone can make a PR that includes malicious code. We prefer to do the code review first to make sure that things look okay, then we move it to Gerrit, which triggers the Jenkins run. Once both items are satisfied we merge.

jisqyv:
@barreeeiroo The tests should also be run with Oracle Java version 7. The production environment on App Engine is still Java 7. If we use Java 8, some things which pass the tests may fail in production.

ewpatton:
Since we use Jenkins at MIT, we won't merge this change. Feel free to keep it on your fork.

PR #1132, opened 2018-01-17
Added 'devmode' target

pavi2410:
This adds a new target to launch GWT CodeServer as discussed in https://groups.google.com/forum/#!topic/app-inventor-open-source-dev/049_f9pBNcI


ewpatton:
@pavi2410 It would be worthwhile to include a devmode target in appinventor/build.xml as well that calls through to the appengine version (see the PlayApp definition for an example of how this is done). This way, one can run the target at the top level rather than having to cd down into the appengine directory.

pavi2410:
Should I name it 'devmode' itself?

moliata:
Hey @pavi2410, I would recommend 'devmode' to be title cased like PlayApp or like RunLocalBuildServer. Thanks for your contribution

pavi2410:
Also may have seen 'installPlay' too

ewpatton:
Yes, there is not a hard and fast rule about whether or not to capitalize the targets in App Inventor. I think it is okay to keep `devmode` as the target.

pavi2410:
Yes, I'm aware of that. GWT requires CSS properties to be written in camelCase. So we have to edit it,

ewpatton:
During development I typically add `<collapse-all-properties/>` to `YaClient.gwt.xml` so that all of the languages are collapsed into a single file. This greatly speeds up the client build.

pavi2410:
My opinion is based on this FAQ.
http://www.gwtproject.org/doc/latest/FAQ_DebuggingAndCompiling.html#Can_I_speed_up_the_GWT_compiler?

pavi2410:
Hello @ewpatton, I'll try to give more concise instructions on how to use 'Super Dev Mode'

1. Run `ant devmode`
2. Run `%AppEnginePath%/bin/dev_appserver.cmd --port=8888 --address=0.0.0.0 appengine/build/war/` where %AppEnginePath% is the environment variable pointing to the App Engine SDK path.
3. Open `localhost:9876` (GWT CodeServer) and drag the two bookmarklets (Dev Mode On & Off) to the bookmarks bar.
4. Open 'localhost:8888` (App Engine server)
5.**To see changes "live"**:
   1. Save your changes.
   2. Click on the "Dev Mode On" bookmarklet
   3. A popup will be shown with a button to compile `ode` module.
   4. Press that button to compile. (That button is actually a bookmarklet. So you can drag this button to the bookmarks bar as well. This will come handy for subsequent compilations)
   5. After that, GWT CodeServer will compile the module incrementally.
   6. Refresh the page and that's it! The changes are live.

I can confirm that all the steps listed above work as expected.

Proof:
1. I made changes to the title of the About dialog.
2. Pressed the "Compile" bookmarklet after the pressing "Dev Mode On" initially.
3. GWT CodeServer started the incremental compilation.
image

4. I refreshed the page and the title of the About dialog has really changed.
image

pavi2410:
By the way, a separate `*.gwt.xml` file will then not require us to manually edit that file manually when running `ant devmode`

pavi2410:
> Change the call in ClonedCheckBox constructor to use pbt.getElement().getFirstChildElement(), otherwise a <span> is passed to the constructor rather than the <input> element. It's unclear why this works in compiled mode but not dev mode.

Regarding this, do you think that this error is caused by that?
> notes = Browser: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/64.0.3282.167 Safari/537.36
foundIn = nls-1153-g59e44785-dirty
faultData = java.lang.AssertionError
projectId = -1

ewpatton:
@pavi2410 Yes that is the same error. If you break on caught exceptions using your browser dev tools you will see that the AssertionError is thrown in the CheckBox constructor called from ClonedCheckBox.

As for the YaClient.gwt.xml file, @jisqyv and I discussed this and we think that it is beneficial to add the `<collapse-all-properties/>` element to eliminate generating the many language/browser permutations.

pavi2410:
@ewpatton, The AssertionError is because the `pbt` element is null.

https://github.com/gwtproject/gwt/blob/c4069141d3056aee73e0e87515da04e5825bfa3a/user/src/com/google/gwt/user/client/ui/UIObject.java#L560

I haven't tried using `getFirstChildElement()` method yet.

ewpatton:
@pavi2410 In the open source forum you stated that we need to run `ant noplay` prior to running `ant devmode`. It seems useful to update the devmode dependencies to include `noplay` so that we can collapse that step.

AppInventorWorkerBee:

jisqyv:

halatmit:
Great.  I'll pick a place when I see who's coming.

== Hal

Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

ewpatton:
@pavi2410 I'd like to merge this in the next release. Would you mind fixing the build issues to make it work with current master?

pavi2410:
@ewpatton If everything is ok, can you please merge this?

ewpatton:
@pavi2410 One small request before I merge this. Can you take the description above and put it into a new section in the README just before the Need Help? section. This way people who are doing development have the documentation locally.

pavi2410:
Ok. I will do so.

ewpatton:
@pavi2410 Gentle reminder to update the README with your devmode instructions. We're going to be doing some more build system work so I'd prefer if we can get this in before that begins in earnest.

pavi2410:
Is that good enough? @ewpatton

PR #1160, opened 2018-02-22
Update whitelist.py and /remote_api

jisqyv:
Update whitelist.py so it works with the latest App Engine SDK. Also
update permissions on /remote_api to be compatible with the latest App
Engine SDK. Note: The change appears to remove the requirement for admin
rights to use the /remote_api facility. However this admin requirement
is now enforced by the Google supplied code which runs behind
/remote_api. Performing the additional check in web.xml just causes an
error about too many authentication attempts.

Also remove excess catch statement in Whitelist.java that has been
giving us compilation warnings for a very long time. Note:
Whitelist.java is no longer in use, but can be enabled via a simple edit
to OdeAuthFilter.java, so we are not removing it yet.

Change-Id: I30cbb0eba3fba3f20d8443a94dae837c827f0b20

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/server/Whitelist.java
+++ b/appinventor/appengine/src/com/google/appinventor/server/Whitelist.java
@@ -94,8 +94,6 @@ public class Whitelist {
       LOG.severe("No whitelist found.");
     } catch (SecurityException e) {
       LOG.severe("Whitelist found, but wrong permission.");
-    } catch (IOException e) {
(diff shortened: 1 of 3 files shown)


jisqyv:
I have testing whitelist.py against a real running instance and the dev server. The script can be a bit finicky depending on what version of the App Engine Python SDK you have installed. But this version works for me, whereas the previous version was without hope!

PR #1166, opened 2018-02-27
Implement translation for extensions

ColinTree:
Related #1018

## Supported translations:
* component name
* help string (have a higher priority than component description)
* properties
* events
* methods
* params (of events & methods)

## Method of implementing
1. When components are updated, `ComponentDatabase.initComponent(JSONObject)` calls `loadComponentTranslations(ComponentDefinition, JSONObject)`.
2. Then translations are loaded into:
    * Blockly - via `BlocklyPanel.appendExtensionTranslation(String, Map<String, String>)`. It then stores translation infomation into `component_database.js`.
    * Designer - via `ComponentsTranslation.extensionMap.put(String ,String)`
        * I added a new field `extensionMap` of HashMap in `ComponentsTranslation.java`, the structure is: `Map<String: componentType, Map<String: name, String: translation>>`.
        * the naming of the translation items are same with the original ones, like `COMPONENT-name` reflects to the translation of the component `name`.
3. I also rewrite the getters of translations in `component_database.js` & `ComponentsTranslation.java`, providing the extensions' translation a higher priority than the built-in translations, since the translations are classify by componentType.
4. That's it.

## Sample
```
{
    "TranslationTest": {
        "en": {
            "component": "Translation Test",
            "helpString": "Help string test",
            "properties": {
                "PropertyTest": "Property Test"
            },
            "events": {
                "EventTest": "Event Test"
            },
            "methods": {
                "MethodTest": "Method Test"
            },
            "params": {
                "paramTest": "param test"
            }
        }
    }
}
```

## 👉[Files for test.zip](https://github.com/mit-cml/appinventor-sources/files/1764677/Files.for.test.zip)
Zip file includes:
* cn/colintree/aix/TranslationTest/
  * TranslationTest.java
  * translation.json
* cn.colintree.aix.TranslationTest.aix
(en & zh_CN only)


ColinTree:
May need discuss: 

About the charset, I implements this on windows, so the default charset is not utf-8 but GBK (for Chinese environment) and it causes problem if the json is in utf8 (seems like this quite common) but the processor is not.

moliata:
Great job, congrats!

ColinTree:
Further plan:
For a better support of i18n, language packs (in Json) can be imported in a similar way, which allows user importing Third-party language packs, especially for extensions.

jisqyv:
@ColinTree This change breaks typeblocks. Load a project, go to the blocks editor and type in the workspace. Nothing happens. Open the console debugger and type in the workspace, observe error...

ColinTree:
the type block problem should be just an typo error...
see
https://github.com/ColinTree/appinventor-sources/blob/implement/extension-translation/appinventor/blocklyeditor/src/blocks/components.js#L557
instant should be instance 😂

ColinTree:
@jisqyv  typeblock problem fixed

PR #1168, opened 2018-03-03
implement FixedInterval for Clock.Timer

ColinTree:
Used time of the previous event will be taken away from the next interval (if true), or intervals start from the end of last event.

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/youngandroid/YoungAndroidFormUpgrader.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/youngandroid/YoungAndroidFormUpgrader.java
@@ -15,7 +15,6 @@ import com.google.appinventor.client.editor.simple.components.MockVisibleCompone
 import com.google.appinventor.client.output.OdeLog;
 import com.google.appinventor.client.properties.json.ClientJsonString;
 import com.google.appinventor.common.utils.StringUtils;
-import com.google.appinventor.components.common.ComponentConstants;
 import com.google.appinventor.components.common.YaVersion;
 import com.google.appinventor.shared.properties.json.JSONArray;
 import com.google.appinventor.shared.properties.json.JSONValue;
@@ -727,6 +726,12 @@ public final class YoungAndroidFormUpgrader {
       // (3) Duration Support was added
       // No properties need to be added to upgrade to version 3.
       srcCompVersion = 3;
+    }
+    if (srcCompVersion < 4) {
+      // The behaviour of the previous versions are not FixedInterval
+      componentProperties.put("FixedInterval", new ClientJsonString("False"));
+      // FixedInterval was added
+      srcCompVersion = 4;
     }
     return srcCompVersion;
   }
--- a/appinventor/components/src/com/google/appinventor/components/runtime/Clock.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/Clock.java
@@ -42,6 +42,7 @@ public final class Clock extends AndroidNonvisibleComponent
     implements Component, AlarmHandler, OnStopListener, OnResumeListener, OnDestroyListener,
                Deleteable {
   private static final int DEFAULT_INTERVAL = 1000;  // ms
+  private static final boolean DEFAULT_PRECISE_INTERVAL = true;
   private static final boolean DEFAULT_ENABLED = true;
 
   private TimerInternal timerInternal;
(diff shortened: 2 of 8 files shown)


ColinTree:
You can see the difference through this test project
[test project.zip](https://github.com/mit-cml/appinventor-sources/files/1777769/test.project.zip)

appinventor/appengine/src/com/google/appinventor/client/youngandroid/YoungAndroidFormUpgrader.java
```diff
@@ -728,6 +727,12 @@ private static int upgradeClockProperties(Map<String, JSONValue> componentProper
       // No properties need to be added to upgrade to version 3.
       srcCompVersion = 3;
     }
+    if (srcCompVersion < 4) {
+      // The behaviour of the previous versions are not FixedInterval
+      componentProperties.put("FixedInterval", new ClientJsonString("False"));
```
jisqyv: Delete this line. Because the default value is False, it isn't explicitly added to a project. So the upgrader is effectively a noop here.

ColinTree: Nahh, the default value of new components is true

jisqyv: I stand corrected.

appinventor/components/src/com/google/appinventor/components/runtime/Clock.java
```diff
@@ -42,6 +42,7 @@
     implements Component, AlarmHandler, OnStopListener, OnResumeListener, OnDestroyListener,
                Deleteable {
   private static final int DEFAULT_INTERVAL = 1000;  // ms
+  private static final boolean DEFAULT_PRECISE_INTERVAL = true;
```
jisqyv: I would recommend getting rid of this and just hardcoding the default. I say this because having this definition leads one to think that you can change the default by just changing this value. However changing the defaults has implications about how to handle different (old and new) versions of the Companion app (because default values are not emitted in the Yail).

ColinTree: Okay.. but we got a upgrader of properties, it should works

jisqyv:
@ColinTree Please check over your changes for trailing whitespace...

appinventor/components/src/com/google/appinventor/components/runtime/Clock.java
jisqyv: Upgraders deal with upgrading projects. I'm talking about dealing with someone using a new version of App Inventor with and older Companion App.

appinventor/appengine/src/com/google/appinventor/client/youngandroid/YoungAndroidFormUpgrader.java
ColinTree: As I try to import a old version project (from mit server), the FixedInterval is automatically set to true (checked) with out this line. Did I do any thing wrong?

I just:
* Commented this line:
  image
* upgrade the YoungAndroidVersion locally
  image

And i got this after imported:
image

It does not work as expected...

With this line i got this which is correct:
image

jisqyv:
@ColinTree We cannot accept this change as written.

The reality is that when we perform a release of MIT App Inventor, not everyone can immediately update their copy of the MIT AI2 Companion. So we are careful to ensure that a new release will work with an old Companion as long as new features are not used. So if your project worked prior to the release with the old (then current) Companion, it would still work after the release even though you didn't update your Companion.

The problem here is that Let's say I have a project with a Clock component. The first time I load it into App Inventor after the release, the project updater will add a FixedInterval property set to False. Because this is not the default value, the generated YAIL will have a (set-and-coerce-property "FixedInterval" False) line in it. This will be unrecognized by the older Companion which will then throw and exception.

I would recommend that you change the default value to False.

ColinTree:
@jisqyv 
I know what problem you are talking sir. But I did this similarly to the others work.
e.g. @ewpatton wrote similar here: https://github.com/mit-cml/appinventor-sources/commit/6e23c47c97882d517d4cd7b1bd6f4a414b7f094e

jisqyv:
@ColinTree Actually that is not what is going on in the change you referenced. The default values of all of the new properties were such that a project loaded into the workspace would not assert any non-default value unless the user explicitly enabled them (and they would then need the new Companion).

Howeve Evan then set some of the properties in the build of the Companion itself. So in this case (themes) user projects would still have the older "legacy" theme while the new Companion would have the modern theme. However loaded projects would not automatically adopt the new modern theme. Instead the project owner would have to explicitly set the theme in their project, but then they would need to have the new Companion to use that project with the Companion.

-Jeff

AppInventorWorkerBee:

ColinTree:
Changed the default value to false on backward compatible purpose

ewpatton:

PR #1180, opened 2018-03-09
Add the "Break" block

jisqyv:
This adds a break block that can be used to break out of for-range,
for-each, and while loops.  Running excuting break terminate execution
of the innermost loop.  It is an error to have a break block that is
not contained in a loop.  Note that this change does not implement
continue, only break.

Change-Id: Ib480b03d0db70d85016a7400579b8fe14435b12a


jisqyv:
@ewpatton @halatmit I'll be putting this code up on ai2-test later tonight or tomorrow afternoon. I have tested it locally and it works, including with the old companion (but of course with the old Companion you cannot execute the "break" block [you get an error]). Unit tests also pass. The primary workhorses are the forrange, foreach and while routines (in runtime.scm). I have also included Hal's -with-break versions. A future revision of App Inventor might want to use these newer versions (this is all about keeping compatibility between Companion versions).

I also have a testing app which I will also post later tonight or tomorrow.

jisqyv:
@ewpatton @kkashi01 @halatmit This code is now up and running on ai2-test.appinventor.mit.edu. You can find a simple test app (.aia and .apk) at:

  https://ipfs.io/ipfs/QmXLC67jYcf19caYfKos9jrVhAumWB4U15Speb5X373opw

-Jeff

Amerkashi:
@jisqyv  THANKS. Can this code be applied on top of 1142?

jisqyv:
@kkashi01 I'm not sure, but I can build you a commit that can. But not until later tonight.

-Jeff

jisqyv:
@kkashi01 https://github.com/jisqyv/appinventor-sources/commit/90350b96757d93d03b96392ae41a18fbea4595ef contains the difference between the version in #1142 and this version.

appinventor/buildserver/src/com/google/appinventor/buildserver/resources/runtime.scm
```diff
@@ -767,26 +767,93 @@
...
-    ((_ lambda-arg-name body-form list)
-     (yail-for-each (lambda (lambda-arg-name) body-form) list))))
-
 
-(define-syntax forrange
+;; Note: (Jeff Schiller) The macro below is intentially unhygienic. We
+;; need to make sure that if there is a *yail-break* form inside
+;; bodyform that it does not get shadowed by the macro (which it would
+;; if this was a hygienic macro).
+
```
halatmit: What a nice way to fix this!!

jisqyv: Now all I need to do is spell intentionally correctly.

jisqyv:
Just updated the comment to correct spelling...

PR #1183, opened 2018-03-14
Fix for missing ripple/touch effect on buttons when color was changed

NmdOfficial:
This little changes fixes the problem that when a user changes the background color of a button, then was the touch/ripple effect removed.


jisqyv:
@NmdOfficial You should resubmit this PR to the mit-cml "ucr" branch. "ucr" stands for next component release.

NmdOfficial:
Done

PR #1184, opened 2018-03-16
Fix for missing ripple/touch effect on buttons when color was changed

NmdOfficial:
The button "setBackgroundColor" function overrides the default drawable.
If we use "getBackground().setColorFilter(...)" we do not override it.

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/ButtonBase.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/ButtonBase.java
@@ -25,6 +25,7 @@ import android.graphics.drawable.ShapeDrawable;
 import android.graphics.drawable.shapes.OvalShape;
 import android.graphics.drawable.shapes.RectShape;
 import android.graphics.drawable.shapes.RoundRectShape;
+import android.graphics.PorterDuff;
 import android.util.Log;
 import android.view.View;
 import android.view.View.OnClickListener;
@@ -373,8 +374,15 @@ public abstract class ButtonBase extends AndroidViewComponent
         } else {
           // Clear the background image.
           ViewUtil.setBackgroundDrawable(view, null);
-          // Set to the specified color (possibly COLOR_NONE for transparent).
-          TextViewUtil.setBackgroundColor(view, backgroundColor);
+          //Now we set again the default drawable
+          ViewUtil.setBackgroundDrawable(view, defaultButtonDrawable);
+          /* Below removes the nice touch/ripple effect.
+           *    TextViewUtil.setBackgroundColor(view, backgroundColor);
+           * We ignore this and do a little bit code magic
+           * 
+           * @Author NMD (Next Mobile Developments) (c)2018
+           */
+          view.getBackground().setColorFilter(backgroundColor, PorterDuff.Mode.SRC_ATOP);
         }
       } else {
         // If there is no background image and the shape is something other than default,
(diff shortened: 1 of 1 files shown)


NmdOfficial:
Done @jisqyv

appinventor/components/src/com/google/appinventor/components/runtime/ButtonBase.java
```diff
@@ -373,8 +374,15 @@ private void updateAppearance() {
...
           ViewUtil.setBackgroundDrawable(view, null);
-          // Set to the specified color (possibly COLOR_NONE for transparent).
-          TextViewUtil.setBackgroundColor(view, backgroundColor);
+          //Now we set again the default drawable
+          ViewUtil.setBackgroundDrawable(view, defaultButtonDrawable);
+          /* Below removes the nice touch/ripple effect.
+           *    TextViewUtil.setBackgroundColor(view, backgroundColor);
+           * We ignore this and do a little bit code magic
+           * 
+           * @Author NMD (Next Mobile Development) (c)2018
```
jisqyv: @NmdOfficial We cannot accept a change that claims a copyright.

halatmit:
This is relevant to the email I just sent about keeping the MIT copyright
and adding "with contributions by ..."



== Hal

Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

NmdOfficial:
@jisqyv I removed The copyright now and added the email.

appinventor/components/src/com/google/appinventor/components/runtime/ButtonBase.java
```diff
@@ -373,8 +374,15 @@ private void updateAppearance() {
         } else {
           // Clear the background image.
           ViewUtil.setBackgroundDrawable(view, null);
-          // Set to the specified color (possibly COLOR_NONE for transparent).
-          TextViewUtil.setBackgroundColor(view, backgroundColor);
+          //Now we set again the default drawable
```
jisqyv: @NmdOfficial Should line 375 (which sets the BackgroundDrawable to null) be deleted, or does it serve a purpose (which I am not clear on)?

appinventor/components/src/com/google/appinventor/components/runtime/ButtonBase.java
```diff
@@ -373,8 +374,15 @@ private void updateAppearance() {
...
+          ViewUtil.setBackgroundDrawable(view, defaultButtonDrawable);
+          /* Below removes the nice touch/ripple effect.
```
jisqyv: This comment is confusing, as "Below" refers to code that has been removed. Instead change the comment to indicate why you (we) are now using view.getBackground().setColorFilter(). Also check for and remove trailing whitespace. Thanks.

appinventor/components/src/com/google/appinventor/components/runtime/ButtonBase.java
NmdOfficial: No the line should not be deleted.

appinventor/components/src/com/google/appinventor/components/runtime/ButtonBase.java
NmdOfficial: I removed the commit now

p4ssw0rd:
Not work when set background color to COLOR_NONE.

p4ssw0rd:
Has bug in classic theme.Setting a button background color may cause the background color of other buttons to change

PR #1198, opened 2018-03-29
Changes to Dialog buttons (layout and color). Fix for #1174

josmas:
changes the order in which the buttons appear, and make delete button red with css. #1174

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockComponent.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockComponent.java
@@ -246,8 +246,9 @@ public abstract class MockComponent extends Composite implements PropertyChangeL
           hide();
         }
       });
-      Button okButton = new Button(MESSAGES.okButton());
-      okButton.addClickHandler(new ClickHandler() {
+      Button deleteButton = new Button(MESSAGES.deleteButton());
+      deleteButton.addStyleName("destructive-action");
+      deleteButton.addClickHandler(new ClickHandler() {
         @Override
         public void onClick(ClickEvent event) {
           hide();
@@ -256,7 +257,7 @@ public abstract class MockComponent extends Composite implements PropertyChangeL
       });
       HorizontalPanel buttonPanel = new HorizontalPanel();
       buttonPanel.add(cancelButton);
-      buttonPanel.add(okButton);
+      buttonPanel.add(deleteButton);
       buttonPanel.setSize("100%", "24px");
       contentPanel.add(buttonPanel);
       contentPanel.setSize("320px", "100%");
--- a/appinventor/blocklyeditor/src/replmgr.js
+++ b/appinventor/blocklyeditor/src/replmgr.js
@@ -362,7 +362,7 @@ Blockly.ReplMgr.putYail = (function() {
                         if (work.failure) {
                             work.failure(Blockly.Msg.REPL_NETWORK_CONNECTION_ERROR);
                         }
-                        var dialog = new Blockly.Util.Dialog(Blockly.Msg.REPL_NETWORK_ERROR, Blockly.Msg.REPL_NETWORK_ERROR_RESTART, Blockly.Msg.REPL_OK, null, 0,
+                        var dialog = new Blockly.Util.Dialog(Blockly.Msg.REPL_NETWORK_ERROR, Blockly.Msg.REPL_NETWORK_ERROR_RESTART, Blockly.Msg.REPL_OK, false, null, 0,
                             function() {
                                 dialog.hide();
                                 context.hardreset(context.formName);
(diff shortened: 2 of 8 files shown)


appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockComponent.java
```diff
@@ -155,8 +155,8 @@ public void onClick(ClickEvent event) {
         }
       });
       HorizontalPanel buttonPanel = new HorizontalPanel();
-      buttonPanel.add(cancelButton);
```
ewpatton: For consistency, I think it is better for the Cancel to come before the OK/Delete buttons. There are other places in the UI where this occurs (create/delete project confirmation, even the browser dialog in Chrome). I think we should keep this original ordering where the cancel is first (left) and the ok is second (right). Otherwise, we should change all other dialogs as well, not just the two pertaining to MockComponents.

appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockComponent.java
```diff
@@ -246,17 +246,18 @@ public void onClick(ClickEvent event) {
...
+      deleteButton.addStyleName("destructive-action");
+      deleteButton.addClickHandler(new ClickHandler() {
         @Override
         public void onClick(ClickEvent event) {
           hide();
           MockComponent.this.delete();
         }
       });
       HorizontalPanel buttonPanel = new HorizontalPanel();
+      buttonPanel.add(deleteButton);
```
ewpatton: Ditto making the delete come after cancel.

josmas:
@ewpatton it does make sense to keep the order as I don't think customising the browser defaults is the way to go; I just went with the requirements in the issue, but can change it and resubmit. Thanks for the review!

josmas:
@ewpatton yeah, that's what I thought; will reverse the order of the two I changed, but will keep the red... am I getting that right?

josmas:
@ewpatton just re-reading your original comment in the issue and I see I got the order wrong in first place (hence the confusion). Anyway, to avoid having two commits, I amended and force-pushed my branch, so this should have the right changes now (just the button name, and the css addition).

josmas:
and to top it up, I clicked the wrong button too! 💃 haha!

josmas:
hey @ewpatton I do like it in material, but mostly because it is all very plain, just background and text. This css has a bunch of other things such as shadows and stuff. I'll give it a go anyway, and post a few screenshots.

josmas:
hey @ewpatton

white_on_red

I think in red with no shadow is the most legible. It still does not look as nice as material, mostly because of the slight emboss in the button, button borders and shadow, and so forth, that makes it all look a bit 2010. But that applies to pretty much the whole look and feel and it's no fixable in just one button.

What you (and others, please chip in!) think?

ewpatton:
@josmas I agree with you on all points. Let's go with the version without shadow for now, and we can think longer term about how to migrate the style globally to something like Material UI.

josmas:
sounds good @ewpatton 
I've amended and pushed again; will change labels here now.

ewpatton:
@josmas Thanks, although looking back at the original issue #1174, it looks like there is one place where we do need to make the change, which is in createDialog in YaBlocksEditor. The buttons are in the wrong order, and the coloring is not applied for the delete scenario. Since multiple paths through blocks editor go through `createDialog`, it might make sense to extend this with another parameter `destructive` that, if true, styles the button. 

Note that the dialog in question is triggered by dragging blocks to the trash can.

josmas:
@ewpatton I was wondering where that screenshot came from! I thought it could be a photoshop thing, cause _grepping_ ode_messages didn't show any matches. Got it placed now, will make the changes. Thanks!

ColinTree:
Is #1173 related?

josmas:
@ColinTree yeah, that second screenshot in #1173 is the one that I am changing at the moment.
The fact that they are two different dialogs for the same thing I would consider as a separate issue (although we should also do something about it!). Same for _enter + scape_ key behaviour.

josmas:
Hey @ewpatton changes are a bit _deeper_ than I had anticipated, because the `Dialog` is also used a bunch in the repl manager file. I also set as `destructive` the cancel button for the emulator connection and hard reset options. Let me know what you think.

appinventor/blocklyeditor/src/replmgr.js
```diff
@@ -778,7 +779,7 @@ Blockly.ReplMgr.startAdbDevice = function(rs, usb) {
     } else {
         message = Blockly.Msg.REPL_STARTING_EMULATOR;
     }
-    progdialog = new Blockly.Util.Dialog(Blockly.Msg.REPL_CONNECTING, message, Blockly.Msg.REPL_CANCEL, null, 0, function() {
+    progdialog = new Blockly.Util.Dialog(Blockly.Msg.REPL_CONNECTING, message, Blockly.Msg.REPL_CANCEL, true, null, 0, function() {
```
ewpatton: What is the rationale for marking the canceling emulator startup as destructive?

josmas: Yeah, it should be a warning really, not a red flag... I can change it back, not a problem.

josmas:
hey @ewpatton just pushed that last change; _Cancel_ button not red anymore.

josmas:
@ewpatton sorry about that! missed it completely. Pushed back up now. Thanks!

PR #1236, opened 2018-05-17
Update Clock Tests

jisqyv:
The recent update to the Clock component broke three unit tests (the
tests were actually checking for the incorrect behavior). This change
updates the tests to check for the correct behavior.

Change-Id: I8fafefeb5a673104c8446c9421b8a0bb98f46553


jisqyv:
@ewpatton This fixes the regression I mentioned the other day.

PR #1239, opened 2018-05-25
Update How Build Server progress is reported.

jisqyv:
Buildserver progress, as reported in the progress bar during a build,
used to work by having the client poll the App Inventor server, which
would in turn poll the buildserver to learn the progress of the build.

However this code would only work properly if there was only one
buildserver and only one build in progress. In reality we handle dozens
of builds simultaneously via a number of independent buildservers. So
the progress being returned to the user was the progress of *some* build
in the system, most likely not theirs!

The constant polling by the App Inventor server also causes problems
with the load balancer in front of the buildservers. This is because
most “jobs” are not real build jobs, but just polls for progress. This
may result in a non-optimal distribution of jobs between the various
buildservers.

This code changes things so that the buildserver does a callback to the
App Inventor server when appropriate milestones are reached in the
build job. The client still polls to App Inventor server for progress,
but the App Inventor server no longer polls the buildserver. This should
result in a better distribution of build jobs on the buildservers and
also returns to the end user the correct status of their build.

We also removed some obsolete code that attempts to generate the Yail
within the Buildserver. We *always* generate the yail in the client now
and it is sent to the buildserver.

Change-Id: Ibaa9a7e3ff6576335d5ab6261a4121f27a1daf1d

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/server/ReceiveBuildServlet.java
+++ b/appinventor/appengine/src/com/google/appinventor/server/ReceiveBuildServlet.java
@@ -74,11 +74,16 @@ public class ReceiveBuildServlet extends OdeServlet {
           LOG.info("Saving android.keystore for user: " + userId);
           storageIo.addFilesToUser(userId, StorageUtil.ANDROID_KEYSTORE_FILENAME);
           storageIo.uploadRawUserFile(userId, fileName, fileBytes);
+        } else if (fileName.equals("build.status")) {
+          int progress = Integer.parseInt((new String(fileBytes)).trim());
+          LOG.info("Received a build.status file contents = " + progress);
+          storageIo.storeBuildStatus(userId, projectId, progress);
         } else {
           String filePath = buildFileDirPath + "/" + fileName;
           LOG.info("Saving build output files: " + filePath);
           storageIo.addOutputFilesToProject(userId, projectId, filePath);
           storageIo.uploadRawFileForce(projectId, filePath, userId, fileBytes);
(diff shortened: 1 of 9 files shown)


jisqyv:
@halatmit @ewpatton You can test this code locally. I'm also planning on putting it up on ai2-test over the weekend.

halatmit:
Is there a way we can test this?

== Hal

Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
@halatmit Just build with the development server as you usually would. Then run a local buildserver in another window (ant RunLocalBuildServer). And then build a project. You can also look at the log (in the window of the dev_server) and see it reporting the progress.

halatmit:
Thanks.

== Hal

Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
@halatmit P.S. I have tested this code, including running it in the dev server with memcache disabled. In this case the progress bar just reports 50% until the build is finished (which interestingly enough isn't much different then when there is real progress, most stages of the build happen quickly. Now in production when things slow down, people may see more stages). I chose this fallback approach rather then storing the status in the datastore or someplace else where it will cost us $$ :-)

halatmit:
I tried this and it seems to work -- or at least the buildserver log
generates a lot of output, and finishes.  And the devserver log shows
doPosts from the buildserver


LGTM

== Hal

Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

PR #1292, opened 2018-06-26
Remove SdkLevel.java

pavi2410:


jisqyv:
@pavi2410 Looks like you did some work here. However there is a reason we have SdkLevel.java. In particular we sometimes want to reference an SDK level greater then the version of the Android SDK that we are using. android.os.Build only defines up to the version of the SDK it is part of.

pavi2410:
@jisqyv But as the SDK versions ints in BUILD.VERSION_CODES are final, they are compiled at compile-time.  Doesn't that work?

jisqyv:
@pavi2410 In the past we have had cases where SdkLevel defines a level greater then the SDK we had linked into the system. That isn't the case at the moment, but might change in the future.

As for the zxing package, that isn't our code. We took code from the Zxing project and modified it just enough to use in App Inventor. By not changing anything else it makes it easier in the future if we want to update to their latest code. The fewer changes we make, the easier the merging process becomes.

PR #1295, opened 2018-06-27
Modified the “while” macro

jisqyv:
Modified the “while” macro to avoid a bug in the Kawa in-line optimizer.

Change-Id: I4247adab9fbf80e2e3396f67c4aac5f4a21fa752

### DIFF
--- a/appinventor/buildserver/src/com/google/appinventor/buildserver/resources/runtime.scm
+++ b/appinventor/buildserver/src/com/google/appinventor/buildserver/resources/runtime.scm
@@ -800,25 +800,22 @@
 
 ;; Also unhygienic (see comment above about foreach)
 
+;; The structure of this macro is important. If the argument to
+;; call-with-current-continuation is a lambda expression, then Kawa
+;; attempts to optimize it. This optimization fails spectacularly when
+;; the lambda expression is tail-recursive (like ours is). By binding
+;; the lambda expression to a variable and then calling via the
+;; variable, the optimizer is not invoked and the code produced, while
+;; not optmized, is correct.
+
 (define-macro (while condition body . rest)
-  `(call-with-current-continuation
-    (lambda (*yail-break*)
-      (let *yail-loop* ()
-        (if ,condition
-            (begin
-              ;; Kawa attempts to be smart by detecting unreachable
-              ;; code and reporting it as an error. This causes an
-              ;; issue here if *yail-break* is determined to be always
-              ;; called within body-form, since *yail-loop* will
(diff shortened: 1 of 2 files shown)


jisqyv:
@p4ssw0rd @ewpatton @halatmit This change is an alternative to #1285. It doesn't require any modification to Kawa, but instead works around the problem (see the comment in runtime.scm).

appinventor/buildserver/src/com/google/appinventor/buildserver/resources/runtime.scm
```diff
@@ -800,25 +800,21 @@
...
-              ;; tail-recursive optimizer that breaks when the while
-              ;; loop is empty.
-              (or (and ((lambda () (begin ,body . ,rest))) #f)
-                  (*yail-loop*)))
-            #!null)))))
+  `(let ((cont (lambda (*yail-break*)
+                 (let *yail-loop* ()
+                   (if ,condition
+                       (begin (begin ,body . ,rest)
+                              (*yail-loop*)))))))
```
ewpatton: Minor nit: If we modify this so that the `let` returns `#!null` like in the previous version, then the `testYailWhile` does not need to be modified. I don't think we rely on the previous semantics anywhere, so we could also go with this version.

jisqyv:
@ewpatton @halatmit So far we only use "While" in the "While" block which doesn't have a return value. That said, I'm not sure what the semantics of what "while" should return. In the case of a "break", it is the value given to the "break" block (which is always #f in our implementation). If the condition test is True, then we loop. If the condition test is False, what do we return? The value of the last body executed? What if the body is never executed?

jisqyv:
@ewpatton P.S. I can make the "else" portion of the "if" statement be #f, which would cause the while statement to always return #f (which is what I believe it did before...)

jisqyv:
@ewpatton @halatmit I'm going to post a revision here which has while return the *the-null-value* which is what while did before we added support for the break block.

jisqyv:

halatmit:
I think it's OK to change this.
== Hal

Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

p4ssw0rd:
@jisqyv Got an error when connecting with companion.

error

jisqyv:
@p4ssw0rd Yep. I'm fixing that now, thanks for catching it.

jisqyv:
@p4ssw0rd I have updated ai2-test again. This version should work :-)

PR #1297, opened 2018-06-29
Reference #!null directly instead

jisqyv:
Reference #!null directly in the while macro instead of via
the *the-null-value* which apparently isn’t always bound.

Change-Id: I2e178f4c6a0317d4014d115912c16b9689bf1dd5


jisqyv:
Fix #1295

jisqyv:
@ewpatton This code is on ai2-test

PR #1305, opened 2018-07-06
DataChanged (CloudDB) now handles images

jisqyv:
Make the DataChanged event consistent with the GotValue event as to how
it handles images.

Change-Id: I08bdbd829cd423fe435081845b863ec4200eb7c4


jisqyv:
Fixes #1303

PR #1342, opened 2018-08-23
Fix a 500 Error

jisqyv:
Google Search reported this to us. If you go to “/b” you get an
ArrayIndexOutOfBounds exception. This is not a valid URL, normally in
practice a code follows the /b/ which points to the APK to download. So
given that /b is not valid, we turn the error into a 404, which makes
more sense.

Change-Id: Id3ac26a35eb2cb51038a74974850ee401b9fb91c


jisqyv:
@ewpatton Changes made...(and tested)

PR #1372, opened 2018-09-23
Increase the default URLFetch timeout

jisqyv:
Change-Id: I0fd9caafd4aeae00d451acc0699e833730415ace

### DIFF
--- a/appinventor/appengine/war/WEB-INF/appengine-web.xml
+++ b/appinventor/appengine/war/WEB-INF/appengine-web.xml
@@ -187,6 +187,10 @@
     <property name="clouddb.uuid.secret" value="changeme!"/>
     <property name="clouddb.secret" value="changeme too!" />
 
+    <!-- This appears to be needed after a suspected change in how
+         POSTs work with URLFetch around 09/20/2018 -->
+    <property name="appengine.api.urlfetch.defaultDeadline" value="40"/>
+
   </system-properties>
 
   <!-- Enable concurrency in the app engine server -->


jisqyv:
@ewpatton @halatmit From the observed behavior it *appears* that URLFetch POST requests get buffered inside App Engine and then finally sent out. I cannot be sure of this, but this *appears* to be the observed behavior. This means that it takes longer to initiate a build for a larger project because this buffering takes time. The default URLFetch timeout is 5 seconds. This change increases it to 40 seconds. So now a large project no longer fails to build, though you can see a noticeable delay from when you initiate the build to when the progress bar appears. But the build does seem to happen now, at least for the modest size project I tested. If necessary we can increase this delay further, but I suspect at 60 seconds we bump into the App Engine hard deadline limit...

Note: This change is running in production on ai2.appinventor.mit.edu as a "patch" to nb170.

I suspect this is needed due to a non-documented (at least I didn't find any documentation) change in the App Engine runtime. It is also possible that this is only an issue with the Java 7 runtime. I am testing to see what happens with the Java 8 runtime, which I believe we can switch to safely at this point.

halatmit:
Thanks.  Maybe if this gets sorted out, we could add some feedback like
"starting build".


== Hal

Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

Amerkashi:
@jisqyv Thanks for clarification. 

I see that you mentioned the change is already available in patch nb170. Has it been tested to see if projects over 10mb can be built? If yes, what's the limit now? 

BTW, I just tried building on ai2.appinventor.mit.edu and didn't notice the delay about progress bar that you mentioned.

jisqyv:
@Amerkashi I haven't found any reference to the limit being removed. When you tried a build, was it a largish project? Small projects do not see any noticeable delay.

Amerkashi:
@jisqyv Will check to see if I can find anyone who can test with large project. Will report back finding

PR #1382, opened 2018-10-02
Add Player Volume Getter

### LINKED ISSUE #969: Player Volume Getter
https://developer.android.com/reference/android/media/MediaPlayer.html

There is no formal "getter" for the volume of the Android MediaPlayer. However, it's very strange that we have a setter and no getter for the volume property within AI. It's easy enough to keep track of the volume ourselves, so it's not a big deal. but I'm posting this as a separate issue from #967.

conorshipp:
This commit adds a getter for the player's volume property.

Resolves #969

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/youngandroid/YoungAndroidFormUpgrader.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/youngandroid/YoungAndroidFormUpgrader.java
@@ -1255,11 +1255,15 @@ public final class YoungAndroidFormUpgrader {
       srcCompVersion = 5;
     }
     if (srcCompVersion < 6) {
-        // The PlayInForeground method was added.
-        // The OtherPlayerStarted event was added.
-        // Properties related to this component have now been upgraded to version  6.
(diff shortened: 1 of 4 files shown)


conorshipp:
I wasn't sure if this required a version increase or an upgrader so I left them out. If they're needed, just let me know and I can add them.

ewpatton:
@conorshipp Yes this should bump the version numbers. For the blocks upgrader, it should be "noUpgrade" because this just adds a new property.

conorshipp:
Upgraders done @ewpatton

conorshipp:
@ewpatton `this.volume` now gets set at the same time as `player.setVolume` if the `playerState` is `State.PREPARED`.

ewpatton:
Note that if I don't set a Source file in the Player, than the following statements do not return what one might expect (true), because the state is not PREPARED without a Source.

screen shot 2018-12-21 at 12 49 09 pm

The = block should return true in the case to be consistent with user expectations.

conorshipp:
Superseded by #1913

PR #1515, opened 2019-01-08
Documentation december2

### LINKED ISSUE #1368: Block name in documentation does not match block name in App Inventor
The help page for the "neg" block uses the name "-" instead. Probably the name was changed at some point to avoid conflict with the minus operation.

There are probably other places where the blocks and documentation diverge. We need an audit of the documentation with corrections.

### LINKED ISSUE #1444: Update TinyDB documentation for namespace usage
Most documentation of TinyDB still needs to be updated to explain namespaces.

Example in http://ai2.appinventor.mit.edu/reference/components/storage.html#TinyDB:

"Each app has its own data store. **There is only one data store per app. Even if you have multiple TinyDB components, they will use the same data store.** To get the effect of separate stores, use different keys. You cannot use the TinyDB to pass data between two different apps on the phone, although you can use the TinyDB to share data between the different screens of a multi-screen app."

### LINKED ISSUE #1461: TinyWebDB creation instructions are out-of-date
Report from user: https://groups.google.com/d/msg/mitappinventortest/uIvvEgjxgPw/nuaBaotCAQAJ

SusanRatiLane:
Small documentation refinements from poweruser-identified issues.

Fixes #1368
Fixes #1461 
Fixes #1444


ewpatton:
Technically, the Namespace property creates separate data stores on disk. I'm not sure we want to stick with the statement that "However, there is only one data store per app."

kdclang:
I was confused by the Namespace property and didn't realize this was possible. Could say, "By default, there is only one data store per app." Or just remove the sentence "However, there is only one data store per app."

SusanRatiLane:
How about: 

"Each namespace represents a single data store that is shared by the entire app. If you have multiple TinyDB components with the same namespace within an app, they use the same data store, even if they are on different screens."

Optionally:

"If you only need one datastore for your app, it's not necessary to define a namespace."

jsheldonmit:
Does it make sense to have"advanced" documentation?

My fear with something like this is that by being exhaustive, covering all functionality, that we obscure the basic information that 90 percent of users will be just fine with.

Along these lines, Susan mentioned to me the idea that, to paraphrase, we're applying bandaids, without considering (yet) the basic design.

Would it make sense to talk more about the relative priority of documentation, and if we decide it's important, plan some design time to think more deeply about how the overall system could be improved?

On Jan 11, 2019 4:16 PM, Susan Rati Lane <notifications@github.com> wrote:

How about:

"Each namespace represents a single data store that is shared by the entire app. If you have multiple TinyDB components with the same namespace within an app, they use the same data store, even if they are on different screens."

Optionally:

"If you only need one datastore for your app, it's not necessary to define a namespace."

PR #1532, opened 2019-01-21
FIX: Little typo on Javadoc

barreeeiroo:
I have just fixed a little issue on Javadoc target at build.xml
It was raising an error, but it was easy to fix it


jisqyv:
@barreeeiroo Yes #1495 fixed this problem as well. I just reviewed and merged it.

PR #1535, opened 2019-01-23
Fix for WebRTC on Safari

jisqyv:
Use the “urls” property of ICEServer instead of the deprecated “url”
property.

Change-Id: I391f06eaffd26261a7a106e9fb35eedf2e0d84ee


jisqyv:
@ewpatton @barreeeiroo OK. So I did this quicker then I said. Testing on Safari (12), Chrome and Firefox. Fixes #1530

PR #1549, opened 2019-02-08
Limit Certain Permissions

jisqyv:
Google has defined a set of permissions that they consider “dangerous”
and can only be used in an application under certain specific
circumstances and then only with special permission from Google. This is
enforced when an application is submitted to the Google Play Store, not
on a device.

This change causes these permissions to not be included in the MIT AI2
Companion. However it is controlled by the “limitPermissions()” flag in
the AppInventor features module. Change the return value to “false” to
give the Companion these permissions.

Change-Id: I0fa3b2e928e7aa53c70cd339f50ed11008fe1876


jisqyv:
@ewpatton @SusanRatiLane @halatmit Note: A version of this change was used to build the Companion currently in the Play Store.

PR #1565, opened 2019-02-15
Implement anonymous procedure

ColinTree:
Need help / suggestion:
- [ ] TOOLTIPs
- [ ] HELPURLs
- [ ] type blocks


ColinTree:
## Blocks added (in category procedure)

image

## Some samples

* calling a global defined procedure  
  image

* Calculating with mode switcher  
  image

wxbit:
Amazing! 👍

wxbit:
I suggest rename "get global procedure" to "get  procedure".

ewpatton:
@ColinTree Rather than having the "get procedure" block take a name parameter, why not provide a dropdown similar to the call blocks? This would allow us to adjust the function reference when the developer renames a procedure.

ewpatton:
Also, because this adds behavior to the companion, the base branch should be `ucr`, not `master`.

wxbit:
A name parameter can be called at runtime.

ewpatton:
Computed, maybe? name is a string, and strings are not callable. In any case, all named procedures will be known at compile time, so I think it makes sense to constrain it accordingly in the blocks.

wxbit:
Sometimes you need to define a series of procedure, the block allows you to decide which one to call based on user input or action.

ColinTree:
You can determine the procedure name that you want to call in runtime, e.g. specified by config from your server during runtime.

ewpatton:
@wxbit @ColinTree I am well aware of the general use case of such a block (in fact, I wrote an [extension](https://groups.google.com/d/msg/app-inventor-open-source-dev/NU4Qcj1PXwg/Ww0D-TN7AgAJ) to do something similar).

My concern is about making it more accessible to the average user. One of the benefits of a visual programming language is the ability to reduce the chance for syntactic and semantic errors. By making it so that the developer can only select valid names, we reduce the chances of introducing semantic errors with this block. We see that people sometimes make spelling errors in the "open another screen" block, for example, and this leads to confusion. I'd rather that we not make the same mistake with this block by making it open ended.

There are other approaches that we could take to achieve the general behavior.
1. If the user want to be able to dynamically look up procedures by name, they could construct a mapping using an associative list to do so.  This can be done using the existing mechanisms in App Inventor. 
2. Use a mutation on the block to allow for the block to be transformed into the parameterized form from the version with the closed set of procedure names, and put the option in the context menu if people really need it. I expect this is a feature only our most advanced users will need. For other distributions, it could be made visible by adding a path in drawer.js.

ColinTree:
@ewpatton Thanks for your reply!

In my opinion, avoid programming mistakes should be the responsibility of developers, designers like us should put more focus on adding more and more possibilities & functionalities into the platform/language.

But I can see your concern is very reasonable, that many of App Inventor users (developers) are still new for programming. So one solution of this should be adding one more block that has a dropdown of defined procedure names and current existing blocks are still kept for them to choose.

What do you think?

ewpatton:
Language designers can't protect developers from logical mistakes, true, but there are certainly plenty of ways in which we can reduce other types of mistakes. For example, this is why some languages have garbage collectors (Java) to prevent memory leaks due to malloc/free mismatches, or memory safety (Rust) to prevent buffer overflow attacks. Look to how variable renaming in App Inventor works as a good example of how we prevent mistakes when it comes to referencing items by name. When you rename a variable, all of its references are renamed at the same time. This doesn't allow one to accidentally decouple the reference from the original definition. I'm simply advocating that we strongly consider that first, and I have given two alternatives approaches that could provide similar flexibility while helping introduce users to the concepts of functional programming in a more guided way.

ewpatton:
Also inviting @jsheldonmit @kdclang @stezelMIT to comment from the educational perspective.

wxbit:
Average user can use the dropdown box block, dynamically look up procedures by name is an advanced usage for experienced users. The PR bring more possibilities for App Inventor.

ColinTree:
@ewpatton Block `getWithDropdown` had been added.
image

kdclang:
Agree with @ewpatton that the dropdown is necessary. I would also recommend hiding the more advanced blocks (this might be something we need to address with the planned feature to hide blocks/components) because it could confuse new programmers to see so many blocks under the procedures drawer.

AppInventorWorkerBee:

ColinTree:
Known issue:
- [ ] Renaming procedure would not rename the reference of get-procedure
- [ ] Get-with-name would not work with non-ASCII procedure names

Zhangzqs:
666

kunalkool:
@ColinTree how to work with run in background blocks and run after period? Can you show some example , how to use that ? And  can you explain about that?
Regards,
Kunal Mishra

Zhangzqs:
> @ColinTree how to work with run in background blocks and run after period? Can you show some example , how to use that ? And can you explain about that?
> Regards,
> Kunal Mishra

You can put an function start(procedure, time), which puts a run after period block, millis=time, passes in an anonymous function, and calls the start(procedure, time) function to form a recursion.

PR #1570, opened 2019-02-17
Several refactoring and improvements on login servlets

jerry73204:
Hi MIT team,
I introduce this PR in anticipation for further alternative authentication support ([LDAP](https://github.com/jerry73204/appinventor-sources/tree/ldap-support) or some). The work is mainly done on refactoring, bug-fixes and improvements on login servlets.

For the first great change, the LoginServlet is broken into LoginGoogleServlet and LoginLocalServlet. Both extending the BaseLoginServlet class. Future authentication maybe named as Login*Servlet as well. This avoids mixing several authentication mechanisms and infers better clarity.

The second is that inline HTMLs are moved to single *.jsp files. Existing *.jsp files are also re-writed and are more HTML5-alike.

I would leave one change for discussion. The route `/login/<id>/setpw` to `/login/setpw?pwid=<id>`, such that the code finds the id by `getParameter()` instead of complicated string operations, and avoids troubles on web.xml. This might break existing mail system. Hence, I need some suggestions to make proper changes here.

Other minor changes include:
- Add essential `URLEncoder.encode()`s. URLs without it are prone to injection attack.
- Rename `getUserFromEmail()` to `getOrCreateUserFromEmail()` to avoid confusing.


jisqyv:
@jerry73204 @ewpatton Thank you for your work on this. I've placed it on hold for now. Mostly because we have three different backends for MIT App Inventor at this point, only the App Engine version is currently in the open source. So to accept this change I would have to do the corresponding work on the non-public code, which I don't have time to do right now (we are a very small team with a large work queue).

However don't despair, I should get to this in the next few months.

jerry73204:
@jisqyv I expect this patch affects little functionality except `/login/<id>/setpw` route. Hope this helps to you a little.

BTW, I'm wondering why the [source code](https://github.com/mit-cml/appinventor-sources/blob/5cbfd9d2b60efa321c37242cd204c9eced05b381/appinventor/appengine/src/com/google/appinventor/server/LoginServlet.java#L131) manually saves encrypted cookie. I suppose GAE runs Jetty 9 that serves Servlet API 3. It could have secure http-only session cookies as the [doc](https://www.eclipse.org/jetty/documentation/9.3.25.v20180904/session-management.html) states.  I have little experience GAE so I may miss something.

Plus, do you have 3rd-party account support in plan, like LDAP, AD or other means?

AppInventorWorkerBee:

PR #1588, opened 2019-03-05
Handle transitive nature of disabled blocks in conditional permissions

wxbit:
Null value should't add to map, and blocks under a disabled block should be ignored.

### DIFF
--- a/appinventor/buildserver/src/com/google/appinventor/buildserver/FormPropertiesAnalyzer.java
+++ b/appinventor/buildserver/src/com/google/appinventor/buildserver/FormPropertiesAnalyzer.java
@@ -169,21 +169,31 @@ public class FormPropertiesAnalyzer {
       DocumentBuilder builder = factory.newDocumentBuilder();
       Document doc = builder.parse(new ByteArrayInputStream(source.getBytes(StandardCharsets.UTF_8)));
       NodeList blocks = doc.getElementsByTagName("block");
-      for (int i = 0; i < blocks.getLength(); i++) {
+      Node parent;
+      FOR: for (int i = 0; i < blocks.getLength(); i++) {
         Element block = (Element) blocks.item(i);
         String type = block.getAttribute("type");
 
         // Skip non-component and disabled blocks
         if (!isComponentBlock(type) || "true".equals(block.getAttribute("disabled"))) continue;
-
+        parent = block;
+        while ((parent = parent.getParentNode()) != null) {
+          if ("block".equals(parent.getNodeName()) 
+              && "true".equals(((Element)parent).getAttribute("disabled"))) {
+            continue FOR;
+          }
+        }
+        
         Element mutation = getMutation(block);
         if (mutation != null) {  // Should always be true, in theory...
-          String componentType = mutation.getAttribute("component_type");
           String blockName = getBlockName(type, mutation);
-          if (!result.containsKey(componentType)) {
-            result.put(componentType, new HashSet<String>());
+          if (blockName != null) {
+            String componentType = mutation.getAttribute("component_type");
+            if (!result.containsKey(componentType)) {
+              result.put(componentType, new HashSet<String>());
+            }
+            result.get(componentType).add(blockName);
           }
-          result.get(componentType).add(blockName);
         }
       }
     } catch(SAXException | IOException | ParserConfigurationException e) {
--- a/appinventor/buildserver/src/com/google/appinventor/buildserver/FormPropertiesAnalyzer.java
+++ b/appinventor/buildserver/src/com/google/appinventor/buildserver/FormPropertiesAnalyzer.java
@@ -172,15 +172,12 @@ public class FormPropertiesAnalyzer {
         public void startElement(String uri, String localName, String qName, Attributes attributes)
             throws SAXException {
           if ("block".equals(qName)) {
-            if ("true".equals(attributes.getValue("disabled"))) {
+            if ("true".equals(attributes.getValue("disabled")) || skipBlocksCounter > 0) {
               skipBlocksCounter++;
-              return;
-            } else {
-              if (skipBlocksCounter > 0) {
-                skipBlocksCounter++;
-              }
-              blockType = attributes.getValue("type");
             }
+            blockType = attributes.getValue("type");
+          } else if ("next".equals(qName) && skipBlocksCounter == 1) {
+            skipBlocksCounter = 0;
           } else if (skipBlocksCounter == 0 && "mutation".equals(qName)) {
             String blockName = null;
             if ("component_event".equals(blockType)) {
--- a/appinventor/buildserver/src/com/google/appinventor/buildserver/FormPropertiesAnalyzer.java
+++ b/appinventor/buildserver/src/com/google/appinventor/buildserver/FormPropertiesAnalyzer.java
@@ -173,7 +173,7 @@ public class FormPropertiesAnalyzer {
             throws SAXException {
           if ("block".equals(qName)) {
             if ("true".equals(attributes.getValue("disabled"))) {
-              skipBlocksCounter = 1;
+              skipBlocksCounter++;
               return;
             } else {
               if (skipBlocksCounter > 0) {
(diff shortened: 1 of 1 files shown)


AppInventorWorkerBee:

appinventor/buildserver/src/com/google/appinventor/buildserver/FormPropertiesAnalyzer.java
```diff
@@ -169,21 +169,31 @@ private static void collectComponentTypes(JSONObject componentProperties,
...
-      for (int i = 0; i < blocks.getLength(); i++) {
+      Node parent;
+      FOR: for (int i = 0; i < blocks.getLength(); i++) {
         Element block = (Element) blocks.item(i);
         String type = block.getAttribute("type");
 
         // Skip non-component and disabled blocks
         if (!isComponentBlock(type) || "true".equals(block.getAttribute("disabled"))) continue;
-
+        parent = block;
```
ewpatton: Rather than do this here, what I think would be more appropriate would be to make a new `isBlockDisabled` function that checks the block's disabled status AND puts the information into a map to memoize it. As implemented, this is O(n*m) in the depth of the tree and number of leaf nodes, rather than O(n) (every ancestor is checked by all of its descendants). This would also prevent having to jump out of the inner loop to the outer loop.

appinventor/buildserver/src/com/google/appinventor/buildserver/FormPropertiesAnalyzer.java
```diff
@@ -169,21 +169,31 @@ private static void collectComponentTypes(JSONObject componentProperties,
...
+          }
+        }
+        
         Element mutation = getMutation(block);
         if (mutation != null) {  // Should always be true, in theory...
-          String componentType = mutation.getAttribute("component_type");
           String blockName = getBlockName(type, mutation);
-          if (!result.containsKey(componentType)) {
-            result.put(componentType, new HashSet<String>());
+          if (blockName != null) {
```
ewpatton: Good catch. I didn't cover the case when component_component_block is used.

wxbit:
In fact, based on my tests, the performance loss of O(n*m) is not so obvious compared to parsing xml. It may be better to exclude disabled nodes when constructing xml.

ewpatton:
The XML represents the whole of the blocks workspace. Removing disabled nodes would mean altering people's programs. How large was the project you were testing on? How deep was the deepest block tree? You definitely won't see performance issues on small programs but for some projects with over 1000 nodes I expect you can start to see discernible differences.

wxbit:
I tested with a 800KB+ bky file, it has more than 2600 blocks.The constructed xml is used for code block statistic only, it doesn't matter.

ewpatton:
That’s all well and good, but I still think we should refactor it to make it clear what’s going on by compartmentalizing the disabled test as its own function. I’m not comfortable with the code using a label to jump from an inner loop to an outer loop. This would be the only place in the entire code that does this, but would go away if we just encapsulated it properly.

halatmit:
I agree with Evan.

== Hal

Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

wxbit:
> That’s all well and good, but I still think we should refactor it to make it clear what’s going on by compartmentalizing the disabled test as its own function. I’m not comfortable with the code using a label to jump from an inner loop to an outer loop. This would be the only place in the entire code that does this, but would go away if we just encapsulated it properly.
> […](#)
> On Mar 5, 2019, at 18:34, 皮皮蟹 ***@***.***> wrote: I tested with a 800KB+ bky file, it has more than 2600 blocks.The constructed xml is used for code block statistic only, it doesn't matter. — You are receiving this because you were assigned. Reply to this email directly, view it on GitHub, or mute the thread.

I think use label is better than this.And I will try to filter out the disabled block.

`        boolean isDisabledBlock = false;
        while ((parent = parent.getParentNode()) != null) {
          if ("block".equals(parent.getNodeName()) 
              && "true".equals(((Element)parent).getAttribute("disabled"))) {
            isDisabledBlock = true;
            break;
          }
        }
        if (isDisabledBlock) {
          continue;
        }`

ewpatton:

AppInventorWorkerBee:
Build failed. See the Jenkins log for more details. Please confirm that you can build this commit and ask another team member to test your build as well.

AppInventorWorkerBee:
Build failed. See the Jenkins log for more details. Please confirm that you can build this commit and ask another team member to test your build as well.

ColinTree:
I whink you need another `final` on this line XD @wxbit 
https://github.com/mit-cml/appinventor-sources/pull/1588/files#diff-83ea1781154eca510f05c45adbae82cfR161

AppInventorWorkerBee:

appinventor/buildserver/src/com/google/appinventor/buildserver/FormPropertiesAnalyzer.java
```diff
@@ -160,87 +158,62 @@ private static void collectComponentTypes(JSONObject componentProperties,
...
+      reader.setContentHandler(new DefaultHandler() {
+        private int skipBlocksCounter = 0;
+        private String blockType;
+
+        @Override
+        public void startElement(String uri, String localName, String qName, Attributes attributes)
+            throws SAXException {
+          if ("block".equals(qName)) {
+            if ("true".equals(attributes.getValue("disabled"))) {
+              skipBlocksCounter = 1;
```
ewpatton: I don't think that this is the right thing to do here. Rather, you should increment `skipBlocksCounter`. This is because it is possible to disable a block and then disable an ancestor, and when walking the tree you will reset the counter back to 1 when reaching the disabled descendant. I was able to construct a set of blocks using this technique that would include the SEND_SMS permission even though all the blocks look disabled in the editor.

Screen Shot 2019-03-22 at 12 51 22 AM

Both the = block and the if block are disabled, but a compiled app will still have the permission as if the SendMessageDirect block is enabled.

wxbit: Thanks

appinventor/buildserver/src/com/google/appinventor/buildserver/FormPropertiesAnalyzer.java
```diff
@@ -160,87 +158,62 @@ private static void collectComponentTypes(JSONObject componentProperties,
...
-            result.put(componentType, new HashSet<String>());
+      XMLReader reader = XMLReaderFactory.createXMLReader();
+      reader.setContentHandler(new DefaultHandler() {
+        private int skipBlocksCounter = 0;
+        private String blockType;
+
+        @Override
+        public void startElement(String uri, String localName, String qName, Attributes attributes)
+            throws SAXException {
+          if ("block".equals(qName)) {
```
ewpatton: One last corner case (hopefully): Because `<next>` elements are contained with `<block>` elements, once you have a disabled block, any blocks following it are considered disabled even though they are logically separate statements. This results in the following block stack not generating the SEND_SMS permission even though SendMessageDirect is enabled.

Screen Shot 2019-03-22 at 10 32 42 AM

My first thought was to reset the `skipBlocksCounter` to 0 when we encounter the `<next>` tag, but that won't work because the `<next>` might be associated with a value or statement input rather than the disabled block. Instead, when we encounter a disabled `<block>` we should add an entry to a stack (could be a `LinkedList`) and update the `skipBlocksCounter` for that entry. We pop the element off the stack if `skipBlockCounter` is 0 and either we hit an opening `<next>` tag or a closing `</block>` tag.

wxbit: I have never thought `<next>` is in `<block>`,

wxbit: image

Test result:

1 Sharing:ShareMessage
2 PhoneCall:MakePhoneCall
3 NxtDrive:Stop
4 NxtDrive:TurnClockwiseIndefinitely
5 NxtColorSensor:GetColor

PR #1596, opened 2019-03-09
Change UseExternalScanner default value from true to false

lightingft:
Changes UseExternalScanner default value to false for Barcode Scanner

Issue #1371

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/BarcodeScanner.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/BarcodeScanner.java
@@ -58,7 +58,7 @@ public class BarcodeScanner extends AndroidNonvisibleComponent
   private static final String LOCAL_SCAN = "com.google.zxing.client.android.AppInvCaptureActivity";
   private static final String SCANNER_RESULT_NAME = "SCAN_RESULT";
   private String result = "";
-  private boolean useExternalScanner = true;
+  private boolean useExternalScanner = false;
   private final ComponentContainer container;
   private boolean havePermission = false; // Do we have CAMERA permission?
 
@@ -171,7 +171,7 @@ public class BarcodeScanner extends AndroidNonvisibleComponent
    *                            false to use internal copy of ZXing.
    *
    */
-  @DesignerProperty(editorType = PropertyTypeConstants.PROPERTY_TYPE_BOOLEAN, defaultValue = "True")
+  @DesignerProperty(editorType = PropertyTypeConstants.PROPERTY_TYPE_BOOLEAN, defaultValue = "False")
   @SimpleProperty()
   public void UseExternalScanner(boolean useExternalScanner) {
     this.useExternalScanner = useExternalScanner;


AppInventorWorkerBee:

ewpatton:
@lightingft This change is more complicated than just changing the two lines. In particular, you're going to need to do the following:

1. Update the BARCODESCANNER_COMPONENT_VERSION number in YaVersion.java
2. Because default values for properties are not set in the project, you will need to add a path in YoungAndroidFormUpgrader.java that will remove the property if it exists (because it wasn't the default and now is) or add it (with the value False, because it was the default and now isn't).
3. versioning.js will need an entry, but it will be "noUpgrade"
4. Set alwaysSend in the `@DesignerProperty` annotation for backward compatibility with older companions (which will have the old default and therefore need to be informed otherwise).
5. Update the documentation to reflect the new default value.

We also need to test this with the Windows emulator as it runs a really old version of the companion by default to make sure we don't inadvertently break anything there.

lightingft:
For testing this with the Windows emulator, is there any way to bypass the update companion prompt, or to somehow go through with it? It gives me the error that the companion is outdated, and then I just get sent to a blank screen where nothing happens.

appinventor/appengine/src/com/google/appinventor/client/youngandroid/YoungAndroidFormUpgrader.java
```diff
@@ -511,6 +511,19 @@ private static int upgradeBarcodeScannerProperties(Map<String, JSONValue> compon
       // The UseExternalScanner property was added.
       srcCompVersion = 2;
     }
+    if (srcCompVersion < 3) {
+      // The UseExternalScanner property was set to True by default.
+      if (componentProperties.containsKey("UseExternalScanner")){
```
ewpatton: Whitespace between `)` and `{`

appinventor/appengine/src/com/google/appinventor/client/youngandroid/YoungAndroidFormUpgrader.java
```diff
@@ -511,6 +511,19 @@ private static int upgradeBarcodeScannerProperties(Map<String, JSONValue> compon
...
+        String value = ((ClientJsonString)componentProperties.get("UseExternalScanner")).getString();
+        if (value.equals("True")){
```
ewpatton: The default value was True, so this will never be the case because we don't store the default value. Instead, you want to remove the key if the value is "False".

appinventor/appengine/src/com/google/appinventor/client/youngandroid/YoungAndroidFormUpgrader.java
```diff
@@ -511,6 +511,19 @@ private static int upgradeBarcodeScannerProperties(Map<String, JSONValue> compon
...
+          componentProperties.remove("UseExternalScanner");
+        }
+      }
+      else {
+        componentProperties.put("UseExternalScanner", new ClientJsonString("False"));
```
ewpatton: Likewise, if the key isn't there, then it was the previous default (True), so you should be setting this to True here, not False.

lightingft:
Suggested changes done.

ewpatton:

appinventor/docs/reference/components/other.html
```diff
@@ -790,8 +790,8 @@ <h2 id="BarcodeScanner">
                     <img alt="Picture of Barcode Scanner component" src="images/barcodescanner.png" />
                     <p>
                       This non-visible component uses the phone's camera to read a 1-dimensional barcode
-                      or 2-dimensional barcode (QR code). In order for this component to work, the
-                      Barcode scanner app from ZXing must be installed on the phone. This app is
+                      or 2-dimensional barcode (QR code). If the UseExternalScanner property is toggled,
+                      the Barcode scanner app from ZXing must be installed on the phone. This app is
```
jisqyv: I wouldn't say "toggled" here. I would say "set". Also change the reference to the Android Market to "Google Play" (which is what the market is called today).

lightingft:
@jisqyv Fair point, although having the internal scanner as a default option would make sense, since it is probably more frequent that a user does not have the external app.

For example, I tried calling DoScan with ExternalScanner enabled, and simply nothing happened, because I do not have the app, whereas the internal scanner works fine, and works for both QR and bar codes without the need for the external app.

I have made the changes requested. I am still not sure how to deal with the failing test. Should we change the test itself somehow, or the .scm of Screen1 in productLookup?

jisqyv:
@lightingft Can we detect if no external scanner is present (error from sending the intent?) and use the Internal scanner if there is no external scanner available?

jisqyv:
@lightingft As to the unit test failure. The problem is that one of the tests is not getting back what it expected, and that is due to your change. So you just edit the test so that what you are returning is now what is expected. This is the normal way to handle such things.

lightingft:
@jisqyv Yes, that is indeed a possibility to make the BarcodeScanner fall back to the internal scanner if an external scanner is not found. Should I revert the default option to true and add this change?

jisqyv:
@lightingft I believe that makes the most sense (changing back the default)

lightingft:
@jisqyv Made the change. It seems that if the external scanner is not found, an exception is raised. So what I did was fallback to setting the useExternalScanner property to false, and call the method again. Seemed to work well with both the external scanner installed, and without it.

I have updated the documentation & the YoungAndroid references and code for the change as well.

jisqyv:
@lightingft You can back out the version update because you are not changing the API

ewpatton:
One concern that I have here is whether this makes sense in this day and age. In particular, I am concerned about us (and by extension anyone developing an App Inventor app) about needing to rely on the presence of third party software. In particular one of the issues I recall seeing was reports where a user had an app installed that included the ZXing activity but it wasn’t public, so there would be a failure condition. It is possible that this updated fix addresses that scenario, but in my opinion I think we should try and exert more control here for a consistent user experience, which is neigh on impossible if we defer by default to another service.

lightingft:
I think combining the two solutions could be an approach that we could take. So the value could be false by default, but if it's set to true and the Scanner is not found, it falls back to the default scanner, thus removing the dependence on the external app.

lightingft:
@ewpatton I agree on having consistency by not relying on external services. I have re-added the change which sets the default value of useExternalScanner to false, and if useExternalScanner is toggled on, but the service is not found, it will fallback to the internal scanner.

I believe it is more misleading in general to have a component use an external service by default, since most users will probably not even be aware that an external service is launched without reading the documentation and looking more into it.

jisqyv:
@lightingft My comment from June 17 still stands. The reason for setting useExternalScanner to true is because the internal one is not as powerful. I am opposed to changing the default here.

lightingft:
@jisqyv @ewpatton Right, I figure we can then still keep the default as true.

I have reverted the changes for the false value (and have undone all the version changes), and kept the change which falls back to the default internal scanner upon catching an error of the scanner not being found.

ewpatton:
@elatoskinas Sounds reasonable to me. @jisqyv Do you concur?

PR #1628, opened 2019-03-24
Have Companion Create a unique ID

jisqyv:
Have the Companion create a unique “InstallationId” which it reports to
the Rendezvous Server. This will help us keep track of how many
Companions are out there and what version of Android (and the Companion)
they are running on.

Note: The id is generated by the ACRA Library and is persistent across
upgrades of the Companion. However it will be re-created if the
Companion is uninstall and then re-installed. It is not related in any
way to any Android provided identifier, which minimizes the privacy
impact of this identifier.

Change-Id: I7b85afe1451f4d44b4e0d412b3f37904425e3112


jisqyv:
Just fixed a typo in the commit comment...

PR #1642, opened 2019-03-30
Remove References to the Package Installer

jisqyv:
This PR consists of ~~two~~ one commit~~s~~. The first "Remove vestigial code..." should definitely go in. It removes unused code that may contain security vulnerabilities.

The second commit removes all references to the Package Installer and instead installs Companion updates and user packaged apps via whatever browser is installed on the user's device. This approach is clunkier and offers a much poorer user experience, but may be required in order to meet the restrictions placed on us by the Google Play Store. Conversations are yet to happen, but here is the code if we need it.


jisqyv:
Note: This PR is on hold pending our discussions

halatmit:
We'll talk with Google and see if there is a way to not mess up the user experience.

PR #1650, opened 2019-04-08
Minor edit to correct numbering in the comment/documentation of testShowScale() method.

pasignature:
Minor edit to correct numbering in the comment of testShowScale() method inside the MapTest class, which unit tests the Map component.

Edited ! to 1 in the comment of testShowScale() method of the MapTest.java class.


AppInventorWorkerBee:

jisqyv:

PR #1657, opened 2019-04-12
Implement a StopListening method for SpeechRecognizer (#1013)

ankritisachan:
The major changes in this commit involve implementing IntentBasedSpeechRecognizer and ServiceBasedSpeechRecognizer. UseLegacy property has also been added.

Resolves [#1013](https://github.com/mit-cml/appinventor-sources/issues/1013)

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/SpeechListener.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/SpeechListener.java
@@ -24,5 +24,5 @@ public interface SpeechListener {
   /**
    * In case we need to report any errors, such as the activity or service dying for an unexpected reason
    */
-  void onError(String message);
+  void onError(int errorCode);
 }
--- a/appinventor/components/src/com/google/appinventor/components/runtime/SpeechListener.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/SpeechListener.java
@@ -1,10 +1,29 @@
+// -*- mode: java; c-basic-offset: 2; -*-
+// Copyright 2009-2011 Google, All Rights reserved
+// Copyright 2011-2012 MIT, All rights reserved
+// Released under the Apache License, Version 2.0
+// http://www.apache.org/licenses/LICENSE-2.0
+
 package com.google.appinventor.components.runtime;
 
+/**
+ * This interface is used to declare functions associated with results and errors of SpeechRecognizer,
+ * used in both ServiceBasedSpeechRecognizer and IntentBasedSpeechRecognizer
+ */
 public interface SpeechListener {
-  // For the service approach in which we can report partial results to the blocks
+
+  /**
+   * For the service based approach in which we can report partial results to the blocks
+   */
   void onPartialResult(String text);
-  // For the final result, used by both approaches
+
+  /**
+   * For the final result, used by both approaches
+   */
   void onResult(String text);
-  // In case we need to report any errors, such as the activity or service dying for an unexpected reason.
+
+  /**
+   * In case we need to report any errors, such as the activity or service dying for an unexpected reason
+   */
   void onError(String message);
 }
--- a/appinventor/components/src/com/google/appinventor/components/runtime/SpeechListener.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/SpeechListener.java
@@ -1,6 +1,5 @@
 // -*- mode: java; c-basic-offset: 2; -*-
-// Copyright 2009-2011 Google, All Rights reserved
-// Copyright 2011-2012 MIT, All rights reserved
+// Copyright 2019 MIT, All rights reserved
 // Released under the Apache License, Version 2.0
 // http://www.apache.org/licenses/LICENSE-2.0
 
--- a/appinventor/components/src/com/google/appinventor/components/runtime/SpeechListener.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/SpeechListener.java
@@ -1,0 +1,10 @@
+package com.google.appinventor.components.runtime;
+
+public interface SpeechListener {
+  // For the service approach in which we can report partial results to the blocks
+  void onPartialResult(String text);
+  // For the final result, used by both approaches
+  void onResult(String text);
+  // In case we need to report any errors, such as the activity or service dying for an unexpected reason.
+  void onError(String message);
+}
--- a/appinventor/components/src/com/google/appinventor/components/runtime/SpeechRecognizer.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/SpeechRecognizer.java
@@ -1,6 +1,6 @@
 // -*- mode: java; c-basic-offset: 2; -*-
 // Copyright 2009-2011 Google, All Rights reserved
-// Copyright 2011-2012 MIT, All rights reserved
+// Copyright 2011-2019 MIT, All rights reserved
 // Released under the Apache License, Version 2.0
 // http://www.apache.org/licenses/LICENSE-2.0
 
@@ -39,7 +39,7 @@ import android.speech.RecognizerIntent;
 @UsesPermissions(permissionNames = "android.permission.RECORD_AUDIO," +
         "android.permission.INTERNET")
 public class SpeechRecognizer extends AndroidNonvisibleComponent
-    implements Component, SpeechListener, Deleteable {
+    implements Component, OnClearListener, SpeechListener {
 
   private final ComponentContainer container;
   private String result;
@@ -56,8 +56,13 @@ public class SpeechRecognizer extends AndroidNonvisibleComponent
    */
   public SpeechRecognizer(ComponentContainer container) {
     super(container.$form());
+    container.$form().registerForOnClear(this);
     this.container = container;
     result = "";
+    recognizerIntent = new Intent(RecognizerIntent.ACTION_RECOGNIZE_SPEECH);
+    recognizerIntent.putExtra(RecognizerIntent.EXTRA_LANGUAGE_MODEL, RecognizerIntent.LANGUAGE_MODEL_FREE_FORM);
+    recognizerIntent.putExtra(RecognizerIntent.EXTRA_PARTIAL_RESULTS, true);
+    UseLegacy(useLegacy);
   }
 
   /**
@@ -98,10 +103,6 @@ public class SpeechRecognizer extends AndroidNonvisibleComponent
       return;
     }
     BeforeGettingText();
-    recognizerIntent = new Intent(RecognizerIntent.ACTION_RECOGNIZE_SPEECH);
-    recognizerIntent.putExtra(RecognizerIntent.EXTRA_LANGUAGE_MODEL, RecognizerIntent.LANGUAGE_MODEL_FREE_FORM);
-    recognizerIntent.putExtra(RecognizerIntent.EXTRA_PARTIAL_RESULTS, true);
-    UseLegacy(useLegacy);
     speechRecognizerController.addListener(this);
     speechRecognizerController.start();
   }
@@ -113,7 +114,9 @@ public class SpeechRecognizer extends AndroidNonvisibleComponent
    */
   @SimpleFunction
   public void Stop() {
-    speechRecognizerController.stop();
+    if (speechRecognizerController != null) {
+      speechRecognizerController.stop();
+    }
   }
 
   /**
@@ -151,7 +154,6 @@ public class SpeechRecognizer extends AndroidNonvisibleComponent
   public void onResult(String text) {
     result = text;
     AfterGettingText(result, false);
-    onDelete();
   }
 
   /**
@@ -164,7 +166,8 @@ public class SpeechRecognizer extends AndroidNonvisibleComponent
   }
 
   @Override
-  public void onDelete() {
+  public void onClear() {
+    Stop();
     speechRecognizerController = null;
     recognizerIntent = null;
   }
@@ -183,6 +186,7 @@ public class SpeechRecognizer extends AndroidNonvisibleComponent
       + "partial results are also provided.")
   public void UseLegacy(boolean useLegacy) {
     this.useLegacy = useLegacy;
+    Stop();
     if (useLegacy == true || Build.VERSION.SDK_INT<8) {
       speechRecognizerController = new IntentBasedSpeechRecognizer(container, recognizerIntent);
     } else {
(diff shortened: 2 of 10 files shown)


AppInventorWorkerBee:

ewpatton:

appinventor/components/src/com/google/appinventor/components/runtime/SpeechRecognizerController.java
```diff
@@ -0,0 +1,7 @@
+package com.google.appinventor.components.runtime;
```
ewpatton: Add copyright and license preamble.

appinventor/components/src/com/google/appinventor/components/runtime/SpeechRecognizerController.java
```diff
@@ -0,0 +1,7 @@
...
+
+abstract class SpeechRecognizerController {
+  abstract void start();
+  abstract void stop();
+  abstract void addListener(SpeechListener speechListener);
+}
```
ewpatton: Add blank line at the end of the file.

appinventor/components/src/com/google/appinventor/components/runtime/SpeechRecognizerController.java
```diff
@@ -0,0 +1,7 @@
+package com.google.appinventor.components.runtime;
+
+abstract class SpeechRecognizerController {
```
ewpatton: The logic of making this an abstract class would be to include the addListener logic here and not repeat it in the individual implementations. Either do that or make this an interface.

appinventor/components/src/com/google/appinventor/components/runtime/SpeechRecognizer.java
```diff
@@ -110,4 +127,59 @@ public void AfterGettingText(String result) {
...
+  @SimpleProperty(
+      category = PropertyCategory.BEHAVIOR,
+      description = "If set, an app can retain their older behaviour.")
+  public boolean UseLegacy() {
+      return useLegacy;
+  }
+
+  @DesignerProperty(editorType = PropertyTypeConstants.PROPERTY_TYPE_BOOLEAN,
+      defaultValue = "True")
+  @SimpleProperty(description = "Used to set UseLegacy property based on user's choice.")
```
ewpatton: This description seems like a tautology to me. Instead, you should include a more user-friendly description, e.g. "If set to true, a separate dialog is used to recognize speech. If false, speech is recognized without changing the user interface. When false, the SpeechRecognizer also provides partial results."

appinventor/components/src/com/google/appinventor/components/runtime/SpeechRecognizer.java
```diff
@@ -110,4 +127,59 @@ public void AfterGettingText(String result) {
...
+  public void UseLegacy(boolean useLegacy) {
+    this.useLegacy = useLegacy;
+  }
+
+  public void initialize(){
```
ewpatton: Rather than use this method, you could simply have the UseLegacy setter instantiate the right object.

appinventor/components/src/com/google/appinventor/components/runtime/SpeechRecognizer.java
```diff
@@ -32,16 +30,21 @@
...
+@UsesPermissions(permissionNames = "android.permission.RECORD_AUDIO," +
+        "android.permission.INTERNET")
 public class SpeechRecognizer extends AndroidNonvisibleComponent
-    implements Component, ActivityResultListener {
+    implements Component, SpeechListener, Deleteable {
 
   private final ComponentContainer container;
   private String result;
+  private Intent recognizerIntent;
+  SpeechRecognizerController speechRecognizerController;
```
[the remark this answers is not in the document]
ankritisachan: Evan, I couldn't understand this comment. Please clarify a bit more.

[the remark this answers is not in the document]
ankritisachan: Okay, I guess you want "private SpeechRecognizerController speechRecognizerController" ?

appinventor/components/src/com/google/appinventor/components/runtime/SpeechRecognizer.java
```diff
@@ -110,4 +127,59 @@ public void AfterGettingText(String result) {
     EventDispatcher.dispatchEvent(this, "AfterGettingText", result);
   }
 
+  /**
+   * Method from SpeechListener interface.
+   */
+  @Override
+  public void onPartialResult(String text) {
+    result = text;
+    AfterGettingText(result);
```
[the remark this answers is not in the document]
ankritisachan: Evan, it was also the case that if the 'stop' method is called, results are displayed via the 'onPartialResult' method and 'not' via the 'onResult' method. So, this parameter will be 'true' even when the user has finished speaking and stop method is called.
Can you suggest what should I write on the tool-tip of AfterGettingText for 'partial' parameter?

appinventor/components/src/com/google/appinventor/components/runtime/SpeechListener.java
```diff
@@ -0,0 +1,10 @@
+package com.google.appinventor.components.runtime;
+
+public interface SpeechListener {
```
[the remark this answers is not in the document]
ankritisachan: What does 'document the class' mean?

appinventor/components/src/com/google/appinventor/components/runtime/SpeechRecognizer.java
```diff
@@ -110,4 +127,59 @@ public void AfterGettingText(String result) {
...
+      category = PropertyCategory.BEHAVIOR,
+      description = "If set, an app can retain their older behaviour.")
+  public boolean UseLegacy() {
+      return useLegacy;
+  }
+
+  @DesignerProperty(editorType = PropertyTypeConstants.PROPERTY_TYPE_BOOLEAN,
+      defaultValue = "True")
+  @SimpleProperty(description = "Used to set UseLegacy property based on user's choice.")
+  public void UseLegacy(boolean useLegacy) {
```
ankritisachan: If I am not wrong, this is the setter method, right?

appinventor/components/src/com/google/appinventor/components/runtime/SpeechRecognizer.java
ankritisachan: Is setter called every time an app is run? 
If I move speechRecognizer instantiating code in the setter, won't it be instantiated only when the setter is called/used? 
A user need not always use the setter to set the value of UseLegacy, isn't it?

appinventor/components/src/com/google/appinventor/components/runtime/SpeechRecognizer.java
ankritisachan: If I am not wrong, this is the setter method, right?

ewpatton: @ankritisachan Correct.

appinventor/components/src/com/google/appinventor/components/runtime/SpeechRecognizer.java
ewpatton: The setter is only called if either the property in the designer is set to the non-default value or if the property is set from the blocks. You can call it from the constructor with the default value if you want to make sure that it is always called at least once.

ewpatton:
@ankritisachan You marked a number of the conversations resolved, but there isn't a new commit for me to review. Did you forget to push?

ankritisachan:
Sorry Evan, there is bit left to be done to finally push a new commit. I marked them as resolved for my reference. Actually, I got engaged with my end semester exams since last week. I will do the push by tomorrow.
Sorry again, for the delay.

ankritisachan:
Hi, how am I supposed to fix the above checks-failure?
Also, in the new commit, I have not yet included changes in the YoungAndroidFormUpgrader.java and versioning.js. I though it would be better to add those changes once all the other main changes with SpeechRecognizer component are okay. 
Please let me know if I should add those changes too.

ewpatton:
@ankritisachan Ah. There is a test in blocklyeditor that includes evaluating the translation from blocks to YAIL that uses the SpeechRecognizer, and by changing the signature of the event the test output is no longer correct. You'll want to replace `($result)` with `($result $partial)` on line 44 of blocklyeditor/tests/com/google/appinventor/blocklyeditor/data/copyCat/copyCatExpected.yail to compensate.

Also, you can go ahead and write the upgraders. Since we are only adding functionality in this PR the upgrades are no-ops (i.e., all existing code will work with the additions). The important thing will be to bump the version number in YoungAndroidFormUpgrader.java. There are plenty of examples in the other components.

ankritisachan:
@ewpatton Changes made. You can proceed with the review now. Sorry for the delay, again.

appinventor/components/src/com/google/appinventor/components/runtime/SpeechRecognizer.java
```diff
@@ -103,11 +126,67 @@ public void BeforeGettingText() {
...
+  }
+
+  /**
+   * Method from SpeechListener interface.
+   */
+  @Override
+  public void onResult(String text) {
+    result = text;
+    AfterGettingText(result, false);
+    onDelete();
```
ewpatton: Is it necessary to do this here? The semantics of onDelete are supposed to be such that it's only called when the component is being removed from the screen in the companion app.

appinventor/components/src/com/google/appinventor/components/runtime/SpeechRecognizer.java
```diff
@@ -103,11 +126,67 @@ public void BeforeGettingText() {
...
+
+  @DesignerProperty(editorType = PropertyTypeConstants.PROPERTY_TYPE_BOOLEAN,
+      defaultValue = "True")
+  @SimpleProperty(description = "If true, a separate dialog is used to recognize speech. "
+      + "If false, speech is recognized without changing the user interface and "
+      + "partial results are also provided.")
+  public void UseLegacy(boolean useLegacy) {
+    this.useLegacy = useLegacy;
+    if (useLegacy == true || Build.VERSION.SDK_INT<8) {
+      speechRecognizerController = new IntentBasedSpeechRecognizer(container, recognizerIntent);
```
ewpatton: See my comment above about stopping before setting speechRecognizerController to something else. Otherwise people might be able to start multiple recognizers which is likely undesirable.

appinventor/components/src/com/google/appinventor/components/runtime/SpeechRecognizer.java
ankritisachan: Okay, yes I get that. Actually, I thought onDelete() needs to be explicitly called, which, I guess, is not the case when I checked in other classes. But, if it is called automatically, when does that happens?

appinventor/components/src/com/google/appinventor/components/runtime/SpeechRecognizer.java
```diff
@@ -103,11 +126,67 @@ public void BeforeGettingText() {
...
+   */
+  @Override
+  public void onError(String message) {
+    result = message;
+    AfterGettingText(result, false);
   }
 
+  @Override
+  public void onDelete() {
+    speechRecognizerController = null;
```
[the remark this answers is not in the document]
ankritisachan: I am currently unable to understand when this onDelete() method is called. Please explain a bit.
Also what kind of changes can happen in designer while speech recognition is running? Can you please provide an example for the same.

appinventor/components/src/com/google/appinventor/components/runtime/SpeechRecognizer.java
ankritisachan: Also, if we are avoiding creation of multiple instances of SpeechRecognizer here, then, along with stopping speechRecognizerController, shouldn't we also initialize it to 'null' before re-initializing with appropriate controller?

ankritisachan: I think, initializing to null is not required. but still just confirming once.
I was thinking about cases when people can start multiple recognizers. Will that happens when start button is clicked again while SpeechRecognizer is already listening (at the same moment)?

appinventor/components/src/com/google/appinventor/components/runtime/SpeechRecognizer.java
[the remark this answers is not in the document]
ewpatton: After reviewing the code, I realize there is a mistake here. onDelete is unreachable in the current system. Instead, we should be implementing OnClearListener instead. The underlying problem is that when one changes the contents of the designer, we send the code to rebuild the entire screen and all of its components. If in the companion app you've started speech recognition, and then make a change to the designer (say, change the color of a button), the screen will be created and a new SpeechRecognizer initialized while the one from the previous version of the screen is still running. Eventually, it will get GCed, but in the mean time there is no way to interact with it so it will continue to record speech.

appinventor/components/src/com/google/appinventor/components/runtime/SpeechRecognizer.java
ewpatton: True, it seems like you should have a flag to represent when speech is/isn't in progress so that you don't call through to the start method more than once. In the Intent-based case, I think it just brings forward the existing speech recognizer. It's unclear what happens in the Service-based case. It would be best just to avoid the ambiguity altogether.

appinventor/components/src/com/google/appinventor/components/runtime/SpeechRecognizer.java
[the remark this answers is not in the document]
ewpatton: Note when you switch to using OnClearListener, you also have to call form.registerOnClearListener(this) in the constructor to register the component.

appinventor/components/src/com/google/appinventor/components/runtime/SpeechRecognizer.java
ankritisachan: @ewpatton I have added Stop() method.
Also, regarding adding flag, it will not be required I guess. From what I checked,
Consider cases:
- Using IntentBasedSpeechRecognizer: Since another custom activity opens up for listening speech, if we try again to click on start button, this custom activity stops and hence listening stops.
- Using ServiceBasedSpeechRecognizer: If I try to click start button again, I will just get an error message "RecognitionService Busy"(already handled in ServiceBasedSpeechRecognizer class), which disappears in a second since recognizer is still listening, and no speech is lost. SpeechRecognizer will behave as if start button was never clicked.

ankritisachan:
Gentle reminder: Please review these changes.

PR #1726, opened 2019-06-11
 Korean translation update

SusanRatiLane:
Update to Korean translation by Lee Jaewoo.


SusanRatiLane:
> @SusanRatiLane Do we need to combine the two fixes into one PR, or is it possible to have the tooltip fix as a separate change? If they aren't tightly coupled, it might make sense to keep them as two changes logically.

This is the first translation to translate the component blocks tooltips, which is why the bugfix went hand-in-hand with finalizing the translation.

There's no big penalty to separating them, and I thought about doing it. If the translation is released without the tooltip bugfix, the component block tooltips will just display in English.

jisqyv:
> This is the first translation to translate the component blocks tooltips, which is why the bugfix went hand-in-hand with finalizing the translation.
> 
> There's no big penalty to separating them, and I thought about doing it. If the translation is released without the tooltip bugfix, the component block tooltips will just display in English.

@SusanRatiLane I would keep the changes separate (tooltip fix and the translation itself). We would want the commit with the fix to be merged first.

-Jeff

SusanRatiLane:
Branch updated with current master. Now the only changes should be the Korean Translation. This is ready for review again.

PR #1728, opened 2019-06-12
Blocks Toolkit

SusanRatiLane:
Feature adapted from Lynda Tang's graduate work.

Adds project-level property "BlocksToolkit" (internally also called SubsetJSON), which is a JSON string describing UI components and code blocks to be displayed in the Designer and Blocks Editor. This allows projects to limit the options in the GUI to fit the curriculum and not overwhelm students.


jisqyv:
When I use the "Export to File" option of the BlocksToolkit it creates a file with the ".txt" extension. Should it use ".json" instead?

appinventor/blocklyeditor/src/drawer.js
```diff
@@ -81,6 +87,209 @@ Blockly.Drawer.buildTree_ = function() {
...
+  console.log("HELLO creating subset block info array here");
+  try {
+    console.log("trying to set subset string");
+    var subsetJsonString;
+    var formName = Blockly.mainWorkspace.formName;
+    var screenName = formName.split("_")[1];
+    if (window.parent.BlocklyPanel_getComponentInstancePropertyValue) {
+      subsetJsonString = window.parent.BlocklyPanel_getComponentInstancePropertyValue(formName, screenName, "BlocksToolkit");
+    } else {
+      subsetJsonString = JSON.stringify(window.reactGetSubsetString());
```
jisqyv: I don't see where window.reactGetSubsetString is defined?

appinventor/blocklyeditor/src/drawer.js
```diff
@@ -81,6 +87,209 @@ Blockly.Drawer.buildTree_ = function() {
...
+    }  
+    console.log("subsetJsonString");
+    console.log(subsetJsonString);
+    var subsetArray = JSON.parse(subsetJsonString);
```
jisqyv: If subsetJsonString is the empty string (which is the case when loading an older project) then this line throws and exception. Should check to see if the string is empty and provide an appropriate define ([] or {}?).

PeterMathijssen:
In the BlocksToolKit i am missing the extension palet.

SusanRatiLane:
> In the BlocksToolKit i am missing the extension palet.

It doesn't support the extensions palette right now. It's worth checking out what that would take. Since extensions can be added and removed, which affects what's in the toolkit, I deemed it out of scope.

However, it's important that extensions palette not be turned off in the Designer, and I don't think i've tested that.

PeterMathijssen:
Yes the extension palet is turned of now when using a custom Blocks Toolkit.

appinventor/blocklyeditor/src/drawer.js
SusanRatiLane: Huh.  I didn't immediately find it either. I'll poke around at it. This is Lynda's code, but I never had a problem with it.

Console logs should probably be cleaned up a bit too.

appinventor/blocklyeditor/src/drawer.js
SusanRatiLane: Heh, and I just hit this when testing. Working on it.

SusanRatiLane:
> Yes the extension palet is turned of now when using a custom Blocks Toolkit.

OK, this should be fixed, and I believe the fix is on the test server right now.

PeterMathijssen:
I now have to extension categories in the palette
image

SusanRatiLane:
I have fixed some bugs and changed Match Project to match the event/method/property component blocks as well as the components and built-in blocks. This code could use some refinement, but out respect for the upcoming release, I'd like to kick this back into the review queue.

PR #1738, opened 2019-07-06
Move component creator README to the right place

pavi2410:


AppInventorWorkerBee:

PR #1739, opened 2019-07-11
Add internationalization to component block tooltips in Blocks Editor

SusanRatiLane:
We currently allow translations of component block tooltips (Method/Event/Property Descriptions in the OdeMessages file), but the blocks editor doesn't display the internationalized strings. This PR fixes that bug.


SusanRatiLane:
This PR still needs review. The Korean (and Polish when released) translations have strings for the component block tooltips.

PR #1778, opened 2019-08-17
Correct example in tooltip for split

halatmit:


halatmit:
I'm not sure I did the chinese correctly.   Ting-Chia might want to have a look.

ewpatton:
@halatmit I'm still not sure this is correct. The space after -potato needs to be included as well, e.g., `"-potato "`, otherwise the split strings will also include the unmatched space. Visually, it won't appear in "Do It" results but it will be significant.

halatmit:
@evan Good catch on that space.
I fixed this, but now there is a merge conflict with the Korean messages.js file.
Could you or Jeff fix that and do the merge?

jisqyv:
@halatmit I'll look at it...(the merge conflict)

halatmit:
I think this happened because someone updated the Korean translation  and I
forgot to pull origin master when I went to patch my PR.

== Hal

Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jisqyv:
@halatmit I pushed to your branch a merge with master, with the conflict resolved. I couldn't quite figure out how to deal with the potato example in Korean because the Kanji do not quite line up with our English usage, so the result is confusing. So I just removed that particular example.

ewpatton:
@halatmit Was there a reason you closed this PR without merging it?

PR #1795, opened 2019-08-23
Add environment sensors

espertus:
I don't expect this to really be merged in. This is a way of getting feedback.

Instance: http://spertus-ai2.appspot.com
APK: https://github.com/espertus/appinventor-sources/raw/a41a83627ce5e0f268abb51435d14ef07a4a6f5c/appinventor/build/buildserver/MIT%20AI2%20Companion.apk


AppInventorWorkerBee:

ewpatton:
@espertus I wouldn't be opposed to taking this PR with some changes. I like the idea of having an abstract superclass shared among the different sensors, which might make sense to backport to the other sensors in the system. It also would be a good launching off point to implement some of the data source/sink work discussed as part of the design for #1776.

espertus:
@ewpatton Thanks for the feedback. Do you mean the new components could be added to App Inventor rather than being extensions (which I had assumed they would be)? If so, I'll update YaVersion (and document how to do that).

I'm not sure which other sensors could be backported, since I believe the new ones are the only ones that return a single value.

espertus:
Here's a test program, for anyone who's interested.

[Test.aia.zip](https://github.com/mit-cml/appinventor-sources/files/3537414/Test.aia.zip)

ewpatton:
Thanks @espertus. I think the biggest thing missing at this point is an update to the sensors documentation page with the new sensors. We will also need you to add a line to the preamble of the new files including the MIT copyright statement.

ewpatton:
Also, because this would require a new companion release, I'm going to change the base branch on GitHub from `master` to `ucr`. Since the two branches are in sync right now this shouldn't cause any problems.

espertus:
I want to make sure all of the icon images are reviewed and correctly licensed or replaced.

I got lightsensor.png from feathericons.com, which uses the MIT license, which I believe requires attribution.

I created the images for the barometer, hygrometer, and thermometer. I am not a graphic designer. Do they meet your standards and/or would anyone like to improve them? The icon for the barometer is the weakest, in my opinion, since it is just a dial. I don't know how to do better in 16x16. Nor do the interwebs, when I search.

espertus:
@ewpatton Should items in sensors.html end in periods? The existing file is inconsistent. (May I make it consistent?)

espertus:
I have updated the documentation. I consider this PR complete, although I will happily make any suggested changes.

ewpatton:

espertus:
In case the license on lightsensor.png is a problem, here are two icons I created myself, either of which could replace it.
light2
light1

appinventor/components/src/com/google/appinventor/components/runtime/LightSensor.java
```diff
@@ -0,0 +1,81 @@
+// -*- mode: java; c-basic-offset: 2; -*-
+// Copyright 2009-2011 Google, All Rights reserved
```
ewpatton: Ditto Google copyright note.

appinventor/components/src/com/google/appinventor/components/runtime/LightSensor.java
```diff
@@ -0,0 +1,81 @@
...
+import android.hardware.SensorManager;
+
+/**
+ * Physical world component that can measure the light level.
+ * It is implemented using
+ * android.hardware.SensorListener
+ * (http://developer.android.com/reference/android/hardware/SensorListener.html).
+ */
+@DesignerComponent(version = YaVersion.LIGHTSENSOR_COMPONENT_VERSION,
+                   description = "Non-visible component that can measure the light level.",
```
ewpatton: 4 space indent

appinventor/components/src/com/google/appinventor/components/runtime/SingleValueSensor.java
```diff
@@ -0,0 +1,190 @@
+// -*- mode: java; c-basic-offset: 2; -*-
+// Copyright 2009-2011 Google, All Rights reserved
```
ewpatton: Ditto Google copyright note

appinventor/components/src/com/google/appinventor/components/runtime/SingleValueSensor.java
```diff
@@ -0,0 +1,190 @@
...
+@SimpleObject
+public abstract class SingleValueSensor extends AndroidNonvisibleComponent
+    implements OnStopListener, OnResumeListener, SensorComponent, SensorEventListener, Deleteable {
+  private static final int DEFAULT_REFRESH_TIME = 1000; // ms
+  private Sensor sensor;
+  protected int sensorType;
+  protected float value;  // most recent value read
+  protected final SensorManager sensorManager;
+  protected boolean enabled;
+  protected int refreshTime;
```
ewpatton: Since we only send property values if they aren't set to default, you should initialize refreshTime to be DEFAULT_REFRESH_TIME. Otherwise, it will default to 0 and so the refresh rate will be much faster than expected.

appinventor/components/src/com/google/appinventor/components/runtime/Thermometer.java
```diff
@@ -0,0 +1,65 @@
+// -*- mode: java; c-basic-offset: 2; -*-
+// Copyright 2009-2011 Google, All Rights reserved
```
ewpatton: Ditto Google copyright note

appinventor/components/src/com/google/appinventor/components/runtime/Thermometer.java
```diff
@@ -0,0 +1,65 @@
...
+import com.google.appinventor.components.common.YaVersion;
+
+import android.hardware.Sensor;
+
+/**
+ * Physical world component that can measure the ambient air temperature if
+ * supported by the hardware.
+ */
+@DesignerComponent(version = YaVersion.THERMOMETER_COMPONENT_VERSION,
+                   description = "Non-visible component that can measure the ambient temperature.",
```
ewpatton: 4 space indent

appinventor/components/src/com/google/appinventor/components/runtime/Thermometer.java
```diff
@@ -0,0 +1,65 @@
...
+    super(container.$form(), Sensor.TYPE_AMBIENT_TEMPERATURE);
+  }
+
+  @Override
+  protected void onValueChanged(float value) {
+    TemperatureChanged(value);
+  }
+  
+  /**
+   * Indicates the temperature changed.
```
ewpatton: Since the Javadoc becomes the English description, we should note here that temperature is in Celsius.

espertus:
I appreciate all of the useful feedback, which I will address today. Could you please also let me know which lightsensor we should use: the better one under the MIT license, the one I placed in the public domain, or something else?

appinventor/components/src/com/google/appinventor/components/runtime/SingleValueSensor.java
```diff
@@ -0,0 +1,190 @@
...
+    
+  @Override
+  public void onResume() {
+    if (enabled) {
+      startListening();
+    }
+  }
+
+  @Override
+  public void onStop() {
```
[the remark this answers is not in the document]
espertus: Thanks for catching that. That's a big deal!

I modeled the code off Accelerometer.java, which is also missing onPause(). Should I fix that in this PR or another one?

appinventor/components/src/com/google/appinventor/components/runtime/SingleValueSensor.java
espertus: That would be bad. Thanks very much for catching this.

appinventor/components/src/com/google/appinventor/components/runtime/Thermometer.java
espertus: FYI, I've updated "How to Add a Component" to reflect proper indentation.

appinventor/components/src/com/google/appinventor/components/runtime/SingleValueSensor.java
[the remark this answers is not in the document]
espertus: It's not clear to me that we need onStop() if we have onPause(), since I believe all calls to onStop() are preceded by onPause(). Do you have an opinion on this?

appinventor/components/src/com/google/appinventor/components/runtime/Thermometer.java
espertus: I added the missing @param tag but see that didn't help, so I changed the javadoc as you suggested.

I made a similar change in Barometer.

espertus:
Should I follow this i18n documentation: https://docs.google.com/document/d/1xk9dMfczvjbbwD-wMsr-ffqkTlE3ga0ocCE1KOb2wvw/edit#heading=h.sajmtj7l1183

appinventor/components/src/com/google/appinventor/components/runtime/SingleValueSensor.java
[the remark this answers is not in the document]
ewpatton: Based on Figure 1 from [Understanding the Activity Lifecycle](https://developer.android.com/guide/components/activities/activity-lifecycle?hl=en), onPause should always be called before onStop, so I agree that we should just use onPause instead.

ewpatton:
> I appreciate all of the useful feedback, which I will address today. Could you please also let me know which lightsensor we should use: the better one under the MIT license, the one I placed in the public domain, or something else?

I prefer the first of the two you posted earlier in this thread (the half sun icon).

ewpatton:
> Should I follow this i18n documentation: https://docs.google.com/document/d/1xk9dMfczvjbbwD-wMsr-ffqkTlE3ga0ocCE1KOb2wvw/edit#heading=h.sajmtj7l1183

Hmm. We should update that document. It's now the case that ComponentTranslations generates a file called `AutogeneratedOdeMessages` that does most of this work for us, so we don't need to duplicate information. It uses the description field in the `@Simple...` annotation, if provided, or the Javadoc if the annotation doesn't include a description. For the Javadoc case, it uses the text up to the first Javadoc annotation (e.g., `@param`). `AutomatedOdeMessages` is a superclass of `OdeMessages`, so any automated messages can be overridden if needed. Therefore, you shouldn't need to do any editing of `OdeMessages` at this point unless there's a non-default behavior that you need.

tl;dr: In most cases, all of this is automated now so unless you're doing something non-standard you needn't touch any of the files as mentioned in that section.

espertus:
I agree. Where should I add the license information?

espertus:
I was wrong about the light sensor image coming from feathericons, which uses the MIT license. It is from a pay site, but I can purchase a license for the icon. I am happy to do that and to upload the icon to this PR. Is there any rule against incorporating purchased imagery?

espertus:
@ewpatton I followed your advice to automatically generate documentation for the new components, which I pasted into sensors.html. I documented the process in "How to Add a Component".

appinventor/components/src/com/google/appinventor/components/runtime/Hygrometer.java
```diff
@@ -44,8 +44,11 @@ protected void onValueChanged(float value) {
   
   /**
    * Indicates the relative humidity changed.
+   *
+   * @pram the new relative humidity
```
ewpatton: Should be `param`, not `pram`

appinventor/docs/reference/components/sensors.html
```diff
@@ -540,33 +539,30 @@ <h3>Methods</h3>
...
-                     <dt> <code class="c2"> Available </code> </dt>
-                      <dd> Reports whether or not the device has an ambient air pressure sensor.</dd>
+<p>A sensor component that can measure the ambient air pressure.</p>
 
-                      <dt> <code class="c2"> Enabled </code> </dt>
-                      <dd> If enabled, then device will listen for changes in air pressure.</dd>
+<h3>Properties</h3>
+<dl>
+  <dt><code><em>AirPressure</em></code></dt>
+  <dd></dd>
```
ewpatton: Description is missing?

appinventor/components/src/com/google/appinventor/components/runtime/Hygrometer.java
espertus: Thanks for catching this.

appinventor/docs/reference/components/sensors.html
espertus: Fixed.

espertus:
I really appreciate the thorough review and that you caught both big and small issues. I'll try to check more carefully in the future.

PR #1796, opened 2019-08-24
GSOC'19 Recycler View Component Implementation in MIT App Inventor

27vinayvyas:
Pull Request for GSoC '19.
Project titled "Recycler View Component Implementation in MIT App Inventor".

Project Description : https://docs.google.com/document/d/1XsG29ZxBeDDHUeGODAViKNa3qcawRP42E_lRuxwF-f8/edit#heading=h.xghftkqvb5c3

Known Issues
1. 'Automatic' attribute of width property is breaking.
2. Selection value property  to be added.


AppInventorWorkerBee:

josmas:
@27vinayvyas hey, your google doc does not seem to have view permissions.

27vinayvyas:
I have changed the permission status now. Thanks.

SusanRatiLane:

SusanRatiLane:
The dyslexic font PR was never merged. The code should not be in this PR.

appinventor/components/src/com/google/appinventor/components/runtime/ListAdapterWithRecyclerView.java
```diff
@@ -0,0 +1,347 @@
...
+            }else if(layoutType==Component.LISTVIEW_LAYOUT_TWO_TEXT){
+            String first =firstItem[position];
+            String second=secondItem[position];
+            
+            holder.textViewFirst.setText(first);
+            holder.textViewSecond.setText(second);
+            }else if(layoutType==Component.LISTVIEW_LAYOUT_TWO_TEXT_LINEAR){
+            String first =firstItem[position];
+            String second=secondItem[position];
+            String combined = first + second;
```
SusanRatiLane: There should be a space inserted between first and second strings.

A different option that occurred to me is that you could keep both textViewFirst and textViewSecond, but turn off content wrapping in the LayoutParams object when you define those LinearLayout objects in the RvViewHolder. This would, I think, cut the content off at the end of one line rather than wrapping the text strangely. You could then  use both the Text and DetailText color/size parameters.

I can see arguments for either solution.

appinventor/components/src/com/google/appinventor/components/runtime/ListAdapterWithRecyclerView.java
```diff
@@ -0,0 +1,347 @@
...
+      else if(layoutType==Component.LISTVIEW_LAYOUT_IMAGE_TWO_TEXT){
+
+        TextView textViewFirst=new TextView(context);
+        idFirst=View.generateViewId();
+        textViewFirst.setId(idFirst);
+        LinearLayout.LayoutParams layoutParams1=new LinearLayout.LayoutParams(LinearLayout.LayoutParams.WRAP_CONTENT,LinearLayout.LayoutParams.WRAP_CONTENT);
+        layoutParams1.topMargin = 10;
+        textViewFirst.setLayoutParams(layoutParams1);
+        textViewFirst.setTextSize(detailTextSize);
+        textViewFirst.setTextColor(detailTextColor);
```
SusanRatiLane: You have TextSize/Color and DetailTextSize/Color parameters reversed. textViewFirst should use the TextSize/Color params, and testViewSecond should use DetailText.

The parameter names are displayed alphabetically, making it kind of confusing because they are placed far apart. I discussed this with Ankriti on her PR, and she agreed to change the names to TextMainSize/Color and TextDetailSize/Color so that they will sort together and make clearer which is which.

appinventor/components/src/com/google/appinventor/components/runtime/RecyclerView.java
```diff
@@ -0,0 +1,721 @@
...
+    return showFilter;
+  }
+
+ public void setAdapterr(){
+
+    int size = currentItems.size();
+
+    String[] first=new String[size];
+    String[] second=new String[size];
+    boolean[] select=new boolean[size];
```
SusanRatiLane: Select is probably a bad name for a variable. It's not a keyword in java, but it is in SQL and other things java might interact with. So it doesn't seem to be causing trouble, but I'd rename it anyway.

Selected or selections seems fine.

appinventor/components/src/com/google/appinventor/components/runtime/RecyclerView.java
```diff
@@ -0,0 +1,721 @@
...
+    
+    ArrayList<Drawable> third = new ArrayList<Drawable>();
+    
+    if(layout==Component.LISTVIEW_LAYOUT_SINGLE_TEXT){
+    for(int i=0;i<size;i++){
+      JSONObject object = currentItems.get(i);
+      first[i]=object.has("Text1")?object.getString("Text1"):"";
+      select[i]=false;
```
SusanRatiLane: It looks like you're initializing the select array to all falses, then passing that array to the adapter constructor.

If the array is always going to be the same when the adapter is created, I don't think there's a reason to pass it as a parameter at all. Just initialize it in the adapter constructor.

SusanRatiLane:
OK, I have added the SelectedItems get-only property and changed the previous selected property to LastClicked because it is set when an item is clicked whether it's selecting or de-selecting.

SusanRatiLane:
Update: We are waiting until some changes go in this winter before merging this PR, but it is still on the priority list. Everything takes longer than expected.

SusanRatiLane:
@ewpatton @jisqyv @ellelili2025  This is ready for another look. In particular, if you have any thorny ListView apps, it would be great to make sure they still work with this new code.

This branch must be merged after https://github.com/mit-cml/appinventor-sources/pull/1751. Ankriti's work in 1751 is the designer-side and Vinay's in this one are the component-side of the same component. Ankriti's branch has been merged into this one. Merging her PR first will preserve correct authorship.

This also includes integration with Evan's Dictionaries enhancement.

SusanRatiLane (approved):

SusanRatiLane (approved):
Ready to merge

PR #1798, opened 2019-08-25
Add legacy connection information to README.md

espertus:
Consistent with the rest of the document, I was terse. Let me know if I give more detail (or change to your taste).

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ You should now be up and running; you can test this by pointing your browser to:
 
     http://localhost:8888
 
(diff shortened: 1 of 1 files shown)


AppInventorWorkerBee:

jisqyv:
Note: We will need a better solution at some point because we will not be supporting Legacy Connections forever. Once we convert from http to https, Legacy Connections will not work (which was the motivation for WebRTC in the first place). We will need to figure out how to handle folks using the dev_server...

PR #1802, opened 2019-08-25
[GSoC][WIP] Listview Custom Layout

ankritisachan:
This PR contains work, to allow App Inventor programmers, to create a custom layout for a ListView row.

Work to be done:

- [ ] Create a fully functional ListViewRow component
- [ ] Enable ListView component to accept only one newly created ListViewRow component
- [ ] Enable ListData property of ListView component to create the table-column headers according to the contents of ListViewRow component to enable adding data to each component of a ListViewRow.
- [ ] Enable visualization of added data in Designer and Android device.


AppInventorWorkerBee:

SusanRatiLane:
From testing: 

1. Tab navigation in the Add Data popup needs attention. I'd expect that hitting tab would move focus to the next element.

2. When I use the "MainText, DetailText (Horizontal)" layout, I see this on my device:

Screen Shot 2019-08-29 at 3 28 01 PM

3. The ElementsFromString property is still available. How is this supposed to interact with the new ListView Data features? Right now, it appears that ElementsFromString will be used unless ListData is specified. That may be the intended behavior. Just want to be sure we've thought about it.

4. In the Designer, the default height of the ListView object appears to be 100x100 pixels, regardless of the content. I would expect sizing to be dynamic when Height or Width is set to Automatic.

5. The Selection property in the designer does not appear to do anything. Isn't this property intended to select a default element when the ListView is loaded on the device?

6. Should there be a way to set selection back to nothing? If it's available, this is usually done by tapping the already-selected element.

7. What is SelectionDetailText intended to do?

8. I like that the main text and detail text can be given separate text sizes, but I think it's confusing that DetailTextSize and TextSize appear so far apart because properties are displayed in alphabetical order. I think it might make better sense to change the names to TextSizeDetail and TextSizeMain.

-- That's all so far. Thanks!

PR #1819, opened 2019-09-05
Remove ancient debug logging

espertus:
It's mortifying that we've been logging to consoles unnecessarily for years.

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockForm.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockForm.java
@@ -6,7 +6,6 @@
 
 package com.google.appinventor.client.editor.simple.components;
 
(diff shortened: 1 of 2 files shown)


AppInventorWorkerBee:

ewpatton:
Some of this is more recent, but I agree we can at this point probably remove this logging as it serves its purpose (although it’s unclear whether there is a significant performance penalty if the dev tools aren’t open).

This is ok to test.

espertus:
> Some of this is more recent

Good to know. It hadn't occurred to me to do a `git blame`.

FWIW, what called my attention to it was when I did my own logging for debugging and had to dig through irrelevant messages.

ewpatton:
@jisqyv Is there any reason we shouldn't remove this logging?

jisqyv:
@ewpatton No reason, it can go...

PR #1828, opened 2019-09-11
Update AssetList.java

wxbit:
bugfix

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/AssetList.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/AssetList.java
@@ -110,17 +110,19 @@ public class AssetList extends Composite implements ProjectChangeListener {
               nodeName.length());
 
         String fileSuffix = node.getProjectId() + "/" + node.getFileId();
-        String treeItemText = "<span style='cursor: pointer'>" + nodeName + "</span>";
+        String treeItemText = "<span style='cursor: pointer'>";
+        Image mediaIcon = null;
         if (StorageUtil.isImageFile(fileSuffix)) {
-          Image mediaIcon = new Image(images.mediaIconImg());
-          treeItemText = "<span>" + mediaIcon + nodeName + "</span>";
+          mediaIcon = new Image(images.mediaIconImg());
         } else if (StorageUtil.isAudioFile(fileSuffix )) {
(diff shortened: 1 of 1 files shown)


AppInventorWorkerBee:

ewpatton:

appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/AssetList.java
```diff
@@ -110,17 +110,19 @@ private void refreshAssetList() {
               nodeName.length());
 
         String fileSuffix = node.getProjectId() + "/" + node.getFileId();
-        String treeItemText = "<span style='cursor: pointer'>" + nodeName + "</span>";
+        String treeItemText = "<span style='cursor: pointer'>";
+        Image mediaIcon = null;
```
ewpatton: Alternatively, instead of setting and checking the variable, within the if-else blocks use `treeItemText += new Image(...);` since at most only one path will be taken.

appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/AssetList.java
```diff
@@ -110,16 +110,15 @@ private void refreshAssetList() {
...
         } else if (StorageUtil.isAudioFile(fileSuffix )) {
-          Image mediaIcon = new Image(images.mediaIconAudio());
-          treeItemText = "<span>" + mediaIcon + nodeName + "</span>";
+          treeItemText += new Image(images.mediaIconAudio()) + nodeName + "</span>";
+        } else if (StorageUtil.isVideoFile(fileSuffix )) {
+          treeItemText += new Image(images.mediaIconVideo()) + nodeName + "</span>";
         } else {
-          Image mediaIcon = new Image(images.mediaIconVideo());
-          treeItemText = "<span>" + mediaIcon + nodeName + "</span>";
+          treeItemText += nodeName + "</span>";
```
ewpatton: You didn't need to move this inside of each branch, you could have left it where it was in the previous change. `nodeName` and `</span>` are always added, so you don't need to duplicate it over every branch.

wxbit: Haha, how could I be so illogical……

wxbit:
@ewpatton Should I also add icons for other types of files? It looks like it's more consistent.

PR #1829, opened 2019-09-11
Add Deprecation Warning on FusionTables

jisqyv:
Back in January we added a warning about the pending deprecation of
FusiontablesControl (due to Google deprecating the service). This
warning is triggered when you drag a FusiontablesControl into a project.

This change will cause a warning to be displayed when loading a project
that already contains a FusiontablesControl. It is implemented as a Form
Upgrader (thus we bump the FUSIONTABLESCONTROL_COMPONENT_VERISON and
YOUNG_ANDROID_VERSION). So after this change goes live each
project (that has a FusiontablesControl) will generate a warning
just *once* when it is loaded.

Change-Id: I1a0b7b12ddea7ecd39c37977799f7d098814e19f


jisqyv:
@wxbit I figured that I would start "slow" on warnings, but perhaps your are correct, I'll modify this PR to be more aggressive.

halatmit:
We need to also look at any tutorials and examples that use fusion tables

jisqyv:
@wxbit OK, I've changed the code to give a warning each time a project is loaded. Thanks for your help.

appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
```diff
@@ -372,6 +372,13 @@
   @Description("Error message if the gallery id does not exist")
   String galleryIdNotExist();
 
+  // Deprecating FusionTables
+
+  @DefaultMessage("The FusiontablesControl has been deprecated and will " +
```
ewpatton: Why not reuse the existing warning text? If we are going to keep different warning text, consider prefixing this with "This project uses FusiontablescControl. " The other text also includes a link to "Learn More", which this one doesn't have.

appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockFusionTablesControl.java
```diff
@@ -27,7 +30,8 @@
 public class MockFusionTablesControl extends MockNonVisibleComponent {
 
   public static final String TYPE = "FusiontablesControl";
-  private static boolean warningGiven = false; // Whether or not we have given experimental warning
+  private static boolean warningGiven = false; // Whether or not we have given experimental warning (on drag out)
```
ewpatton: We should set liveWarningGiven to true if we also set warningGiven to true. The reason for this is if you create a blank project and drag in a new FusiontablesControl you will end up with two dialogs, one for the drag out and one for the attach. If you've received the drag out dialog it doesn't make sense to then immediately get the onAttach dialog. Of course, it might be the other way around...

appinventor/blocklyeditor/src/versioning.js
```diff
@@ -1527,7 +1527,10 @@ Blockly.Versioning.AllUpgradeMaps =
 
     // The LoadingDialogMessage property was added
     // The ShowLoadingDialog property was added
-    4: "noUpgrade"
+    4: "noUpgrade",
```
ewpatton: Since we are no longer relying on upgrading to trigger the warning, this should be removed (the control version is still 4).

appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockFusionTablesControl.java
SusanRatiLane: To play devil's advocate, if you got the popup upon opening the project and STILL try to add a new control, maybe you need to be reminded again....

jisqyv:
@ewpatton I believe I have address all review comments (good catches!). I have reduce the warning flags to just one. It is set either at project load or when a FusiontablesControl is dragged out into a project that never had one before. Also using the same warning dialog. Reverted versioning.js, I missed that one...

PR #1836, opened 2019-09-15
Add Timeout property to Web component (#1782)

austenjoa:
Allows for users to handle Web request timeouts through two methods; a TimeOut block for each Web component, as well as through the Screen error handler.


AppInventorWorkerBee:

austenjoa:
Opened PR to incorrect branch

PR #1913, opened 2019-11-16
Add more getters and setters for properties

### LINKED ISSUE #714: add more getters and setters for properties
For example, font boldface and italic.   Maybe lots more.

bartmathijssen:
Resolves #714

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/youngandroid/YoungAndroidFormUpgrader.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/youngandroid/YoungAndroidFormUpgrader.java
@@ -320,6 +320,12 @@ public final class YoungAndroidFormUpgrader {
       } else if (componentType.equals("SpeechRecognizer")) {
         srcCompVersion = upgradeSpeechRecognizerProperties(componentProperties, srcCompVersion);
 
+      } else if (componentType.equals("Switch")) {
+        srcCompVersion = upgradeSwitchProperties(componentProperties, srcCompVersion);
+
+      } else if (componentType.equals("TableArrangement")) {
+        srcCompVersion = upgradeTableArrangementProperties(componentProperties, srcCompVersion);
+
       } else if (componentType.equals("TimePicker")) {
         srcCompVersion = upgradeTimePickerProperties(componentProperties, srcCompVersion);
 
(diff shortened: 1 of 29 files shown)


AppInventorWorkerBee:

appinventor/components/src/com/google/appinventor/components/runtime/TableArrangement.java
```diff
@@ -66,7 +66,7 @@ public int Columns() {
    */
   @DesignerProperty(editorType = PropertyTypeConstants.PROPERTY_TYPE_NON_NEGATIVE_INTEGER,
       defaultValue = "2")
-  @SimpleProperty(userVisible = false)
+  @SimpleProperty
```
ewpatton: Don't make the row/column setters visible as there is no way to set the row/column of the views to fill those positions if they change.

ewpatton:

appinventor/components/src/com/google/appinventor/components/runtime/CloudDB.java
```diff
@@ -390,6 +390,7 @@ public void run() {
 
   @DesignerProperty(editorType = PropertyTypeConstants.PROPERTY_TYPE_STRING,
     defaultValue = "DEFAULT")
+  @SimpleProperty
```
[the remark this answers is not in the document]
bartmathijssen: I made the mistake of pushing the changes to the CloudDB and FirebaseDB components, this is wasn't my plan originally and is why the component versions of these components haven't been upgraded either. I don't see a purpose for the user to change these settings at runtime. What is your take on this?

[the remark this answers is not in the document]
ewpatton: I would remove these because often they might also need different security tokens and those might not be changeable at runtime.

appinventor/appengine/src/com/google/appinventor/client/youngandroid/YoungAndroidFormUpgrader.java
```diff
@@ -344,6 +350,42 @@ private static void upgradeComponentProperties(Map<String, JSONValue> componentP
...
+      } else if (componentType.equals("NxtSoundSensor")) {
+        srcCompVersion = upgradeNxtSoundSensorProperties(componentProperties, srcCompVersion);
+
+      } else if (componentType.equals("NxtTouchSensor")) {
+        srcCompVersion = upgradeNxtTouchSensorProperties(componentProperties, srcCompVersion);
+
+      } else if (componentType.equals("NxtUltrasonicSensor")) {
+        srcCompVersion = upgradeNxtUltrasonicSensorProperties(componentProperties, srcCompVersion);
+
+      } else if (componentType.equals("Ev3Motors")) {
```
ewpatton: I'd like to try and keep this function as alphabetized as possible, so please move the Ev3* checks up to the appropriate location.

appinventor/components/src/com/google/appinventor/components/common/YaVersion.java
```diff
@@ -1111,9 +1160,13 @@ private YaVersion() {
...
   // For SWITCH_COMPONENT_VERSION 1
   //  - Initial Version
-  public static final int SWITCH_COMPONENT_VERSION = 1;
+  // For SWITCH_COMPONENT_VERSION 2:
+  // - FontBold, FontItalic, FontTypeface properties made visible in block editor.
+  public static final int SWITCH_COMPONENT_VERSION = 2;
 
-  public static final int TABLEARRANGEMENT_COMPONENT_VERSION = 1;
+  // For TABLEARRANGEMENT_COMPONENT_VERSION 2:
+  // - Added GetColumns, GetRows methods
```
ewpatton: These are properties, not methods.

appinventor/appengine/src/com/google/appinventor/client/youngandroid/YoungAndroidFormUpgrader.java
```diff
@@ -814,6 +876,55 @@ private static int upgradeEmailPickerProperties(Map<String, JSONValue> component
...
     }
+    if (srcCompVersion < 4) {
+      // TextAlignment, FontBold, FontItalic, FontTypeface properties made visible in block editor.
+      srcCompVersion = 4; 
+    }
+    return srcCompVersion;
+  }
+
+  private static int upgradeEv3ColorSensorProperties(Map<String, JSONValue> componentProperties,
+                                                     int srcCompVersion) {
```
ewpatton: Code style: continuing indentation should be 4 spaces (if you're using IntelliJ you can adjust this by turning off alignment of parameters). Make this change to all of the new methods in this file.

appinventor/appengine/src/com/google/appinventor/client/youngandroid/YoungAndroidFormUpgrader.java
```diff
@@ -1364,6 +1499,23 @@ private static int upgradeSpeechRecognizerProperties(Map<String, JSONValue> comp
     return srcCompVersion;
   }
 
+  private static int upgradeSwitchProperties(Map<String, JSONValue> componentProperties,
+      int srcCompVersion) {
```
ewpatton: These two are correctly indented.

appinventor/appengine/src/com/google/appinventor/client/youngandroid/YoungAndroidFormUpgrader.java
bartmathijssen: I fixed the styling in the latest commit. I was wondering if it wouldn't be better to have code styles defined in an .editorconfig file or even have a code linter installed. This way there shouldn't be any issues with code style when making a pull request.

bartmathijssen:
@ewpatton I made all the requested changes. Unfortunately I wasn't able to build the HTML files, only the markdown. I must be doing something wrong. Another question. Why are the HTML files needed in the version control? The markdown files are already there, wouldn't that be enough?

ewpatton:
The theory is that not everyone will have Jekyll set up correctly (as you’ve seen) and so by committing both the markdown and HTML we reduce the chances that people are providing out of date documentation. In the original design we did have the HTML as a build output that was ephemeral (not checked in) but decided that since we couldn’t count on people having their machines configured correctly that we would just check both in. I can push a commit that updates the HTML. I expect this also means we need to update the README.

ewpatton:
@bartmathijssen Do you want to bring this up-to-date with ucr? I can do a review once it's in sync.

bartmathijssen:
@ewpatton Everything should be up-to-date now. Also, when building the sources to generate the new documentation I found that some generated folders, like `vendor` and `.sass-cache` are not included into the .gitignore. Wouldn't it be better if it did include these folders.

ewpatton:
@bartmathijssen Yes, but I would recommend doing that in a separate PR.

daki7711:
@bartmathijssen @ewpatton I know it could be a little pain to match this to current App Inventor but can we try to revive this?

ewpatton:
Yes. @bartmathijssen do you think you would be able to revisit this, or should we try to allocate some time from a core dev?

ghost:
@ewpatton I have not worked on this code base in a very long time and don't have time to go back and revisit this pull request. I think it would be better to allocate time from the core developers.

PR #1916, opened 2019-11-22
Remove the FusionTablesControl from visibility

jisqyv:
Moved the FusionTablesControl from the STORAGE to the INTERNAL
category. This means that projects that include the control can load,
but people working on new projects cannot add the control to their
project.

Also updated the documentation to include the warning that it is
shutdown.

Change-Id: Ie05d54b9df0a5bae21661529ded01ce0a549347c


jisqyv:
Note that I am using past tense when referring to December 3rd. It will be past tense soon enough :-)

jisqyv:
@ewpatton Well, the documentation for FustionTables will soon be moot.

PR #1921, opened 2019-11-26
Use PROPERTY_TYPE_TEXTAREA instead of PROPERTY_TYPE_STRING

### LINKED ISSUE #1914: Long text in textfield of textbox or label

When i want to add a long text to the about screen textfield i can make the field bigger.

Would it be helpful if we can do the same with the text fields of a textbox and a label in the designer? Those are the components that might have the most text inside of them but now only have one small textfield to put the text inside of it. So editing is difficult.

When learning AI users will use the designer textfields more then in a later stadium where they will learn how to use blocks. 

Designer textfields are just a single line not a textarea like the about textfield. 

When learning AI and putting larg amounts of text in the bigger textfield users have a better view of what they are doing. 

This is a list of the designer textfields that could be made into a textarea.

**UserInterface**
Label.Text
TextBox.Text
Spinner.ElementsFromString
ListView.ElementsFromString

**Maps**
LineString.PointsFromString
Polygon.HolePointsFromString
Polygon.PointsFromString
FeatureCollection.FeaturesFromGeoJSON

bartmathijssen:
Resolves #1914

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/FeatureCollection.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/FeatureCollection.java
@@ -39,7 +39,7 @@ public class FeatureCollection extends MapFeatureContainerBase implements MapFea
   }
 
   @SuppressWarnings("squid:S00100")
-  @DesignerProperty
+  @DesignerProperty(editorType = PropertyTypeConstants.PROPERTY_TYPE_TEXTAREA)
   @SimpleProperty(description = "Loads a collection of features from the given string. If the " +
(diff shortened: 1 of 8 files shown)


AppInventorWorkerBee:

ewpatton:

jisqyv:
Shouldn't this be based on ucr instead of master as it impacts the components on devices.

ewpatton:
Since it’s just changing the property editor shown in the designer, there isn’t an API change that will affect the companion. I think it is safe for master.

jisqyv:
@ewpatton I stand corrected. :-)

PR #1924, opened 2019-11-29
update community joining link in README.md

shivanik6z:
the old link was directing to the google group which is not active anymore and it took me a while to join the community via new link. I hope this new link will minimise the confusion and will help other new contributors too.


AppInventorWorkerBee:

PR #1942, opened 2019-12-09
Project List Folders

austenjoa:
Add folders in web project list display with a drag & drop interface

Unfinished Tasks:
- [x] Modify Move To Trash button to work with folders as well
- [ ] Add RPC call upon adding & deleting user folders to add folder persistency for empty folders
- [ ] Modify trash view to accommodate folders as well?


AppInventorWorkerBee:

ewpatton:
@SusanRatiLane and I should review this for the purposes of the course, but @jisqyv should sign off on the final version before it is merged.

ewpatton:

appinventor/appengine/src/com/google/appinventor/client/explorer/project/ProjectManager.java
```diff
@@ -12,13 +12,10 @@
 import com.google.appinventor.shared.rpc.project.ProjectNode;
 import com.google.appinventor.shared.rpc.project.UserProject;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
jisqyv: You should revert this change, we do not want exports of the form foo.*, but instead explicit exports of modules.

austenjoa: Will fix this in next commit, had my IDE misconfigured so it automatically did a star import.

appinventor/appengine/src/com/google/appinventor/client/wizards/NewFolderWizard.java
```diff
@@ -0,0 +1,113 @@
+package com.google.appinventor.client.wizards;
+
+import com.google.appinventor.client.Ode;
+import com.google.appinventor.client.widgets.LabeledTextBox;
+import com.google.appinventor.client.widgets.Validator;
+import com.google.appinventor.client.youngandroid.TextValidators;
+import com.google.gwt.event.dom.client.*;
```
austenjoa: Fixing this for next commit

appinventor/appengine/src/com/google/appinventor/server/ProjectServiceImpl.java
```diff
@@ -51,7 +52,7 @@
  *
  * <p>Note that this service must be state-less so that it can be run on
  * multiple servers.
- *
+ * S
```
austenjoa: Removing this mistake in next commit

SusanRatiLane:
@austenjoa I see in the description that you're working on moving entire folders. I was not able to move individual projects into a folder either. Is that expected at this stage?

appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/ProjectList.java
```diff
@@ -6,21 +6,24 @@
...
 import com.google.appinventor.client.explorer.project.ProjectManagerEventListener;
-import com.google.appinventor.shared.rpc.project.GalleryApp;
-import com.google.gwt.event.dom.client.ClickEvent;
-import com.google.gwt.event.dom.client.ClickHandler;
-import com.google.gwt.event.dom.client.MouseDownEvent;
-import com.google.gwt.event.dom.client.MouseDownHandler;
+import com.google.appinventor.shared.rpc.project.UserProject;
+import com.google.common.collect.Ordering;
+import com.google.gwt.core.client.GWT;
+import com.google.gwt.event.dom.client.*;
```
austenjoa: Fixing wildcard import in next commit

austenjoa:
> @austenjoa I see in the description that you're working on moving entire folders. I was not able to move individual projects into a folder either. Is that expected at this stage?

@SusanRatiLane Individual project movement should be working, but there is an issue with UI that I need to fix- currently, it is impossible to select the project for drag/drop without highlighting the text first and then dragging it afterwards (the name labels are where the drag handlers are configured). This obviously is bad UI-- I discussed this with Evan and will try to fix this via a CSS style for the table.

SusanRatiLane:
> > @austenjoa I see in the description that you're working on moving entire folders. I was not able to move individual projects into a folder either. Is that expected at this stage?
> 
> @SusanRatiLane Individual project movement should be working, but there is an issue with UI that I need to fix- currently, it is impossible to select the project for drag/drop without highlighting the text first and then dragging it afterwards (the name labels are where the drag handlers are configured). This obviously is bad UI-- I discussed this with Evan and will try to fix this via a CSS style for the table.

Ah, OK, selecting the text worked for me, so I can reproduce your current features.

I think there should be a visual distinction between folders and projects. Perhaps make the folder names bold?

austenjoa:
My original plan was to add an additional icon to the left of the name to distinguish the two (and that might still be worth adding later), but bolding the names in addition is definitely good idea. I'll make sure to add that in the next commit + the highlighting names fix.

SusanRatiLane:
Testing observations of varying importance:

I can drag a project into a folder, but once it is in the folder, dragging it to "Parent Folder" does not move it back out. This is what intuition would tell me to do.

The pointer displays as a text-select cursor when hovering over "Parent Folder" rather than a pointer as it does for all other clickable areas

It looks like you opted to de-bold the folder names rather than italicize?

If I move an empty folder to trash, it disappears. That's not totally unreasonable, especially since empty folders don't persist, but it might surprise some folks. 

Is there a reason for "Parent Folder" to have a checkbox at all? I see that it's there but disabled.

austenjoa:
> I can drag a project into a folder, but once it is in the folder, dragging it to "Parent Folder" does not move it back out. This is what intuition would tell me to do.

This is definitely unintended and seems to be a regression from a previous commit I made-- I will fix this.

> The pointer displays as a text-select cursor when hovering over "Parent Folder" rather than a pointer as it does for all other clickable areas

I cannot seem to replicate this-- the closest thing that I get is that the regular mouse pointer icon appears between list items, but I get the clickable pointer/white glove point when hovering over parent folder. This seems to be consistent when I tested on Firefox & Chrome. What browser did you get this issue on? I applied the same CSS style to Parent folder as all the other folders, so I'm not sure what would cause this issue for just the parent folder and not the other folders.

> It looks like you opted to de-bold the folder names rather than italicize?

image
image

Inspecting the HTML element, I think I applied both bold and italicize to it- is there anything I need to do besides apply both styles to get it to work properly?

> If I move an empty folder to trash, it disappears. That's not totally unreasonable, especially since empty folders don't persist, but it might surprise some folks.

This is a feature that would correspond to "Modify trash view to accomodate folders"; I don't think there's any coherent way to implement moving an empty folder to trash without having a trash folder display. Should this be considered higher priority? I agree that this behavior is inconsistent, but not sure if it is higher priority than folder persistency.

> Is there a reason for "Parent Folder" to have a checkbox at all? I see that it's there but disabled.

I put that there as a placeholder- I'm not 100% on how the HTML Grid element works in GWT, but I'll see if I can set it to be empty without messing up the rest of the grid arrangement.

SusanRatiLane:
OK, confirmed on dragging projects out of folders. And the checkbox on "Parent Folder" is now gone.

This is what I see: 
Screen Shot 2019-12-13 at 10 17 51 AM
(Fred is the folder.)

We don't see a CSS definition for ode-FolderNameLabel in the code we pulled from the branch. The CSS changes should be in war/Ya.css. Did you check them in there?

austenjoa:
It should be fixed now- I mistakenly made the changes to build/war/Ya.css instead of the source css file. Perhaps the text selection issue was also caused by this as well?

ewpatton:
@austenjoa Yep that did the trick.

SusanRatiLane:
> It should be fixed now- I mistakenly made the changes to build/war/Ya.css instead of the source css file. Perhaps the text selection issue was also caused by this as well?

That fixed both issues!

Getting more contorted:

If I create a Folder Fred and create a project Test3 in it, then move Fred to trash, I see just Test3 in trash. I believe that matches your to-do list in the PR description. Then if I restore Test3 and return to My Projects, I don't see anything I restored until I reload, at which point Fred reappears with Test3 inside. This may be less a bug than a consequence of the incomplete feature.

It's possible to create a project and a folder on the same level with the same name. Do we want to permit this?

OTOH, this looks good -- If I create a folder Fred with Test3 in it, then move Fred to Trash and create a NEW folder Fred with Test4 in it, then restore the original Fred from trash, I get one folder Fred with both Test3 and Test4 in it.

austenjoa:
> If I create a Folder Fred and create a project Test3 in it, then move Fred to trash, I see just Test3 in trash. I believe that matches your to-do list in the PR description. Then if I restore Test3 and return to My Projects, I don't see anything I restored until I reload, at which point Fred reappears with Test3 inside. This may be less a bug than a consequence of the incomplete feature.

This should be fixable even with the feature incomplete- I believe I just have to refresh the table when switching back from the Trash to Project view (since it currently doesn't refresh the table, even though the new folder is registered it isn't redrawn as a row yet).

> It's possible to create a project and a folder on the same level with the same name. Do we want to permit this?

I can change this in the next commit as well.

ewpatton:
@SusanRatiLane Based on our discussion in the office, it sounds like we should defer this to a future sprint rather than try and squeeze it into the January release. Does that accurately reflect your assessment?

SusanRatiLane:
Lessons learned from this project were applied to #2514 refactor proof-of-concept, and the feature is on track for a future release.

PR #1945, opened 2019-12-11
Fix gitignore to include Vagrant generated files

bartmathijssen:
Pull Request #1540 added the Vagrant build to App Inventor but overlooked to include the Vagrant generated files to the .gitignore file. We don't want contributors to add these files to the sources by accident.


AppInventorWorkerBee:

PR #1953, opened 2019-12-19
Updates for weblate launch

SusanRatiLane:
**Note: This should not be merged until updates have been added to handle JSON-format Blockseditor translation files.**

This is a collection of changes needed for launching the new weblate translation system. 

The weblate instance (http://weblate.appinventor.mit.edu) points to a separate repository appinventor-translations, currently hosted on my Github account (SusanRatiLane). This will probably be moved in the future to be owned by mit-cml, but that can be done at any time.

Changes in this PR:
 - new _messages.json files for all translations of Blocklyeditor. This is a one-time conversion of the .js files. The js files have not been removed yet.

 - Fixes to errors in the js files that caused trouble when trying to parse them into json files

 - New qqq.json files for Blocklyeditor translations. These will contain context descriptions for the translation strings, though they don't have much useful in them right now.  This matches the style currently used in Blockly.

 - Update to Brazilian Portuguese translation of OdeMessages, which was used as a beta test.


SusanRatiLane:
Changes here are superseded by #1996

PR #1966, opened 2019-12-28
Add timezone support to Clock component

barreeeiroo:
This PR allows the ability to play with timezones in which components are stored at

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/Clock.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/Clock.java
@@ -22,9 +22,11 @@ import com.google.appinventor.components.runtime.util.TimerInternal;
 import com.google.appinventor.components.runtime.util.ErrorMessages;
 
 
+import java.util.Arrays;
 import java.util.Calendar;
 import java.util.GregorianCalendar;
 import java.util.Date;
+import java.util.TimeZone;
 import java.text.DateFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
@@ -603,6 +605,21 @@ public class Clock extends AndroidNonvisibleComponent
   @SimpleFunction (description = "Text representing the time of an instant")
   public static String FormatTime(Calendar instant) {
     return Dates.FormatTime(instant);
+  }
+
+  /**
+   * Changes the timezone of a given instant.
+   *
+   * @param instant : instant to modify
+   * @param timezone : timezone to apply to instant
+   * @return : updated instant
+   */
+  @SimpleFunction(description = "Updates the timezone in which the instant is saved at")
+  public static Calendar ChangeTimezone(Calendar instant, String timezone) {
+    if (Arrays.asList(TimeZone.getAvailableIDs()).contains(timezone))
+      return instant.setTimeZone(TimeZone.getTimeZone(timezone));
+    else
+      throw new YailRuntimeError("Your given timezone is not a valid timezone.");
   }
 
   @Override


AppInventorWorkerBee:

appinventor/components/src/com/google/appinventor/components/runtime/Clock.java
```diff
@@ -605,6 +607,21 @@ public static String FormatTime(Calendar instant) {
...
+  /**
+   * Changes the timezone of a given instant.
+   *
+   * @param instant : instant to modify
+   * @param timezone : timezone to apply to instant
+   * @return : updated instant
+   */
+  @SimpleFunction(description = "Updates the timezone in which the instant is saved at")
+  public static Calendar ChangeTimezone(Calendar instant, String timezone) {
+    if (Arrays.asList(TimeZone.getAvailableIDs()).contains(timezone))
```
ewpatton: I'm wondering if it makes sense to cache this the first time this method is called so it doesn't have to be constructed every time someone calls ChangeTimezone?

barreeeiroo: Got, I just changed it. This weekend I'll check if there's anything else I can do in the PR, and mark it as ready for review.

barreeeiroo:
@ewpatton totally forgot about this PR in the meetup. The only missing block I think that this PR need is a "getTimezones" that returns a List of all available timezones.
Is there any other block you may want to add?

appinventor/components/src/com/google/appinventor/components/runtime/Clock.java
```diff
@@ -605,6 +611,31 @@ public static String FormatTime(Calendar instant) {
...
+  /**
+   * Changes the timezone of a given instant.
+   *
+   * @param instant  instant to modify
+   * @param timezone  timezone to apply to instant
+   * @return  updated instant
+   */
+  @SimpleFunction(description = "Updates the timezone in which the instant is saved at")
+  public static Calendar ChangeTimezone(Calendar instant, String timezone) {
+    if (timezones.contains(timezone))
```
kkashi01: Suggestion: trim spaces and ignore case in comparison

barreeeiroo: image

TimeZone is actually case sensitive, so "america/detroit" will not be a valid timezone. We should not allow users to input it lowercase even though we know it is "America/Detroit", as it's not officially supported by Java.

barreeeiroo:
If a user wants, for example, _america/detroit_, they should use the following procedure:
1. Get all available timezones using the `Timezones` property
2. Look on each element and compare both strings using the `lowercase` text block
3. If they match, get the index of the element or the value of the element
4. Use that index to get the value on the `Timezones` list, or use the fetched value (not the user-given one)

That is how a normal developer would do, as the block is just calling the native method, we're not making conversions into it, as we would be making suppositions of what a user might want.

appinventor/components/src/com/google/appinventor/components/runtime/Clock.java
kkashi01: It is case-sensitive, that's why I mentioned:
> If found, then use the actual value that is in the list (not the passed-in parameter) and set the timezone to the found-value

ewpatton:
> The more direct control we give users, the better. Because we are giving them a higher range of opportunities and chances to learn.
> If a user sees a procedure of "getTimezoneCaseInsensitive" made with blocks, they will gain the logic of how it can be done. Instead, if we give them already done, we are restricting the possibilities (and such procedure can be done quite easily with blocks, without needing us to make assumptions).

Well, they could always learn Java. That gives them much more opportunity than App Inventor. 😄 

One of the goals with App Inventor is to make hard stuff easier. I think allowing one to set the timezone using a case insensitive string fits within that directive. I expect that in other apps students will have plenty of opportunity to learn how to use the upper/lower case blocks. I'm not convinced there isn't any reason to make a component purposefully "dumber" about something. Timezones are already a [challenging enough problem](http://www.creativedeletion.com/2015/01/28/falsehoods-programmers-date-time-zones.html) without also introducing things like case sensitivity into it.

barreeeiroo:
Should be now fixed. It now first checks if it matches the equal timezone, and if not it checks if it matches the case-insensitive one. If it does, it re-runs the method but with the correct timezone.

ewpatton:
@barreeeiroo It looks like you missed an import when you did the merge.

barreeeiroo:
> @barreeeiroo It looks like you missed an import when you did the merge.

My bad. 🤦‍♂
It should be now fixed.

PR #1968, opened 2019-12-28
Fix String comparisons

kkashi01:
Fixing String comparisons

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/widgets/properties/SubsetJSONPropertyEditor.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/widgets/properties/SubsetJSONPropertyEditor.java
@@ -330,13 +330,13 @@ public class SubsetJSONPropertyEditor  extends PropertyEditor
               for (int k = 0; k < jsonComponentBlockProps.size(); ++k) {
                 JSONObject jsonComponentBlockType = jsonComponentBlockProps.get(k).isObject();
                 String componentBlockType = jsonComponentBlockType.get("type").isString().stringValue();
-                if (componentBlockType.equals("component_set_get")) {
+                if ("component_set_get".equals(componentBlockType)) {
                   componentPropHash.put(jsonComponentBlockType.get("mutatorNameToValue").isObject().get("property_name").isString().stringValue(), "PROP");
-                } else if (componentBlockType.equals("component_event")) {
+                } else if ("component_event".equals(componentBlockType)) {
                   JSONValue mutatorValue = jsonComponentBlockType.get("mutatorNameToValue");
                   JSONValue event_name = mutatorValue.isObject().get("event_name");
                   componentPropHash.put(event_name.isString().stringValue(), "EVENT");
-                } else if (componentBlockType.equals("component_method")) {
+                } else if ("component_method".equals(componentBlockType)) {
                   componentPropHash.put(jsonComponentBlockType.get("mutatorNameToValue").isObject().get("method_name").isString().stringValue(), "METHOD");
                 }
               }
@@ -591,7 +591,7 @@ public class SubsetJSONPropertyEditor  extends PropertyEditor
   }
 
   protected void updateValue() {
-    if (property.getValue().equals("")) {
+    if ("".equals(property.getValue())) {
       dropDownButton.setCaption("All");
       dropDownButton.setWidth("");
     } else {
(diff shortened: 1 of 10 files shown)


AppInventorWorkerBee:

ewpatton:

appinventor/appengine/src/com/google/appinventor/client/widgets/properties/SubsetJSONPropertyEditor.java
```diff
@@ -591,7 +591,7 @@ public void onValueChange(ValueChangeEvent<Boolean> valueChangeEvent) {
   }
 
   protected void updateValue() {
-    if (property.getValue() == "") {
+    if ("".equals(property.getValue())) {
```
ewpatton: Consider making this `property.getValue() == null || property.getValue().isEmpty()`

appinventor/appengine/src/com/google/appinventor/server/project/youngandroid/YoungAndroidProjectService.java
```diff
@@ -831,7 +831,7 @@ private String getBuildServerUrlStr(String userName, String userId,
 
   private String getCurrentHost() {
     if (Server.isProductionServer()) {
-      if (appengineHost.get()=="") {
+      if ("".equals(appengineHost.get())) {
```
ewpatton: I believe that this will always return a not-null value, so you could use `isEmpty()` instead of checking equality to `""`.

appinventor/appengine/src/com/google/appinventor/client/widgets/properties/SubsetJSONPropertyEditor.java
kkashi01: I was actually thinking about updating the com.google.appinventor.common.utils.StringUtils to include a method like Apache commons lang3 (IsEmpty) that checks for both null and length 0. Would that be better?

ewpatton: @kkashi01 That sounds acceptable.

kkashi01: @ewpatton Done. Added StringUtils.isEmpty and now using it

appinventor/common/src/com/google/appinventor/common/utils/StringUtils.java
```diff
@@ -329,4 +329,25 @@ public static int countMatches(String str, String sub) {
...
+   * </pre>
+   *
+   * <p>NOTE: This method changed in Lang version 2.0.
+   * It no longer trims the CharSequence.
+   * That functionality is available in isBlank().</p>
+   *
+   * @param cs  the CharSequence to check, may be null
+   * @return {@code true} if the CharSequence is empty or null
+   */
+  public static boolean isEmpty(final CharSequence cs) {
```
ewpatton: After having thought about it a bit, I'd like if you renamed this to `isNullOrEmpty` so it's more explicit. Imagine in some future context we have a scenario where null and empty string have different meanings; the fact that `isEmpty(null) == true` is not obvious from the name of this method (versus `isEmpty("")`).

appinventor/common/src/com/google/appinventor/common/utils/StringUtils.java
```diff
@@ -329,4 +329,25 @@ public static int countMatches(String str, String sub) {
...
+   *
+   * <pre>
+   * StringUtils.isEmpty(null)      = true
+   * StringUtils.isEmpty("")        = true
+   * StringUtils.isEmpty(" ")       = false
+   * StringUtils.isEmpty("bob")     = false
+   * StringUtils.isEmpty("  bob  ") = false
+   * </pre>
+   *
+   * <p>NOTE: This method changed in Lang version 2.0.
```
ewpatton: Where is version 1? What's the license of the code where you took this from?

appinventor/common/src/com/google/appinventor/common/utils/StringUtils.java
kkashi01: @ewpatton Renamed and updated the 1-line statement. FYI, if string contains all spaces, method will return false. If you like, I can trim and check

PR #2003, opened 2020-01-26
Add the EmailAddresses property to activity starter.  For the UCR branch

halatmit:


jisqyv:
Instead of adding a new property, could we just update the “Extras”
property to be able to accept a YailList as the value of a pair. So
you would provide a list like:

```lisp
((“android.intent.extra.EMAIL” (“user1@example.com” “user2@example.com)))
```

PR #2004, opened 2020-01-29
Handle special cases of trigonometric functions.

jsuyash1514:
This PR adds the results for the corner cases of trigonometric functions sin, cos, and tan.
Solves #1816 
@ewpatton Please have a look.


AppInventorWorkerBee:

ewpatton:

jsuyash1514:
@ewpatton I've added some negative values in the tests. Please have a look.

halatmit:
I'm not following this carefully, but it seems that the tests are testing
only integer arguments.

Are we also testing that identities are obeyed, e.g., is tan(x)*cos(x)=
sin(x) for all x?


== Hal

Hal Abelson
hal@mit.edu
Prof. of Comp. Sci. and Eng.

jsuyash1514:
@ewpatton @halatmit I've added a test for a trigonometric identity tan(x)*cos(x) = sin(x). Please let me know if the way I've implemented this test is correct. If so, I would like to add some more trigonometric identities and equations.

appinventor/buildserver/tests/com/google/appinventor/buildserver/YailEvalTest.java
```diff
@@ -1123,6 +1142,24 @@ public void testAtan() throws Throwable {
      testUnaryDoubleFunction("atan-degrees", args, vals);
    }
 
+   public void testTrigonometricIdentities() throws Throwable {
+       Random random = new Random();
```
ewpatton: Indentation is off here. Should be 4 spaces, not 6.

appinventor/buildserver/tests/com/google/appinventor/buildserver/YailEvalTest.java
```diff
@@ -13,6 +13,9 @@
 import junit.framework.TestCase;
 import kawa.standard.Scheme;
 
+import java.util.Random;
+
+
```
ewpatton: Google style guide has the imports placed alphabetically, so the java import should come after gnu and before junit. You also don't need the extra blank lines.

appinventor/buildserver/src/com/google/appinventor/buildserver/resources/runtime.scm
```diff
@@ -1793,13 +1793,31 @@
           360))
 
 (define (sin-degrees degrees)
-  (sin (degrees->radians-internal degrees)))
+  (if(= (modulo degrees 90) 0)
```
ewpatton: Minor code style thing: I think generally it's accepted to put a space before the first `(` in a sequence of `(`, so there should be a space between the `if` and the `(` that comes after it, e.g., `(if (=`. I recommend this change for everywhere it occurs in your change.

jsuyash1514:
@ewpatton I've made the requested changes and resolved the conflicts with the base branch. Kindly review.

jsuyash1514:
@ewpatton I've added tests for some more trigonometric identities and equations. Please have a look and let me know if anything remains to be done.

PR #2026, opened 2020-02-11
Add for each in dictionary block

### LINKED ISSUE #2019: Request foreach block for dictionary
Iterate over key-value pair, as foreach for List do.

BeksOmega:
### Resolves

#2019 

### Description

Adds a "for each key with value in dictionary" block that works like a normal for each block, except instead of providing an "item" variable it provides a "key" variable and a "value" variable.

### Problems

I did run into problems with generating the code (as @ewpatton suspected I might hehe).

My strategy was to try and generate the same code as these blocks:
blocks (2)
Which is:
```
(foreach $item (begin   (let ( ($key (call-yail-primitive yail-list-get-item (*list-for-runtime* (lexical-value $item) 1) '(list number) "select list item")) ($value (call-yail-primitive yail-list-get-item (*list-for-runtime* (lexical-value $item) 2) '(list number) "select list item"))  ) #f )) (call-yail-primitive make-yail-list (*list-for-runtime* ) '() "make a list"))
```
I got the generator to generate this code:
```
(foreach $item (begin   (let ( ($key (call-yail-primitive yail-list-get-item (*list-for-runtime* (lexical-value $item) 1) '(list number) "select list item"))  ($value (call-yail-primitive yail-list-get-item (*list-for-runtime* (lexical-value $item) 2) '(list number) "select list item")) #f)) )
```

But it doesn't seem to be working :/

I tested using this project: [test.zip](https://github.com/mit-cml/appinventor-sources/files/4183719/test.zip)

There were no messages in the js console. But I did find something using logcat. When I loaded the project to my phone this error would occure: [AppInvError.txt](https://github.com/mit-cml/appinventor-sources/files/4183725/AppInvError.txt)

Sadly I could not interpret it :/

Any help is very much appreciated!

### Testing
Besides the obvious problem of the generator, I believe everything is working.

#### Break Statement
Tested that the break statement knows it is valid when inside "for each key in dictionary" blocks.
blocks
Observe how there is no warning on the break block.

#### Variable Scope
Tested that variables know when they are in scope and out of scope.
blocks (2)
Observe how there are warnings on the variables when they are outside the scope of the for each loop, but there are no warnings when they are inside the scope of the for each loop.

#### Variable Renaming
Tested that variables get properly renamed.
VarRenaming
Observe how the renaming obeys the scope of the variable.
Note that the warning upon renaming the loop-var also occures with foreach loops. So I think that would be better saved for a different PR.
RenameError

### Additional Information

This is my first PR into this repo so please tell me if there's anything I can improve on! I tried to write my commit messages following the style guide, but it's still pretty new to me.

Also if you want unit tests I can totally write some, I just couldn't find any so I didn't write any.


AppInventorWorkerBee:

appinventor/blocklyeditor/src/blocks/control.js
```diff
@@ -416,6 +416,140 @@ Blockly.Blocks['controls_forEach'] = {
...
 
+Blockly.Blocks['controls_for_each_dict'] = {
+  category: 'Control',
+
+  helpUrl: Blockly.Msg.LANG_CONTROLS_FOREACH_DICT_HELPURL,
+
+  init: function() {
+    console.log('init');
+    this.setColour(Blockly.CONTROL_CATEGORY_HUE);
+    this.appendValueInput('DICT')
```
ewpatton: While this is technically correct (the best kind of correct), I'd like us to move more toward using `interpolateMsg` as a means of constructing blocks since I think it helps reduce the number of strings that need to be internationalized (we're at over 30k strings so it's a bit daunting). Please rewrite the block definition to use that approach.

appinventor/blocklyeditor/src/generators/yail/control.js
```diff
@@ -93,6 +93,50 @@ Blockly.Yail['controls_forEach'] = function() {
          + listCode + Blockly.Yail.YAIL_CLOSE_COMBINATION;
 };
 
+Blockly.Yail['controls_for_each_dict'] = function() {
+  var yail = Blockly.Yail;
+  var prefix = Blockly.usePrefixInYail ? 'local_' : '';
+  var keyName = yail.YAIL_LOCAL_VAR_TAG + prefix + this.getFieldValue('KEY');
+  var valueName = yail.YAIL_LOCAL_VAR_TAG + prefix + this.getFieldValue('VALUE');
+
+  var dictionaryCode = yail.valueToCode(this, 'DICT', yail.ORDER_NONE);
```
ewpatton: It would make sense to have a fallback here in case the input is empty. In `controls_forEach` we construct an empty dictionary and logically or the result of this call with that (see line 89). You could add a new constant to Blockly.Yail, e.g., `EMPTY_DICT` and define it as `(YailDictionary:makeDictionary)`. Alternatively you should be able to use the existing empty list code because an empty list or empty dictionary should behave the same when iterated over.

appinventor/blocklyeditor/src/warningHandler.js
```diff
@@ -386,10 +386,20 @@ Blockly.WarningHandler.prototype["checkIsNotInLoop"] = function(block) {
   }
 };
 
-Blockly_loopBlockTypes =
-  // add more later
-  ["controls_forEach", "controls_forRange", "controls_while"] ;
-
+// TODO: Maybe change this to a property of the block, instead of maintaining
```
ewpatton: This seems like it would be a straightforward thing to add, i.e. have a `{?boolean=}` property of the block `isLoopBlock`. It would be faster than looking up the block's type in an array. Maybe in another PR?

appinventor/blocklyeditor/src/warningHandler.js
```diff
@@ -386,10 +386,20 @@ Blockly.WarningHandler.prototype["checkIsNotInLoop"] = function(block) {
...
+//   a list. Check how this interacts with extensions first.
+Blockly_loopBlockTypes = [
+    "controls_forEach",
+    "controls_for_each_dict",
+    "controls_forRange",
+    "controls_while"
+];
+
+// TODO: Maybe place this on an object. Options:
```
ewpatton: I could see it being a property of a block, with the recursive call being made on the block's parent, if one exists. Potentially at that point we can memoize the result since most operations on the workspace won't change that (only moves that affect the ancestors of the current block).

appinventor/blocklyeditor/src/generators/yail/control.js
```diff
@@ -93,6 +93,50 @@ Blockly.Yail['controls_forEach'] = function() {
...
+  var bodyCode = yail.statementToCode(this, 'DO') || yail.YAIL_FALSE;
+
+  var loopIndexName = 'item';
+  var loopIndexCommandAndName = yail.getVariableCommandAndName(loopIndexName);
+  loopIndexName = loopIndexCommandAndName[1];
+  var getListCode = loopIndexCommandAndName[0] + loopIndexName +
+      yail.YAIL_CLOSE_COMBINATION;
+  var getKeyCode = yail.YAIL_CALL_YAIL_PRIMITIVE + 'yail-list-get-item' + yail.YAIL_SPACER
```
ewpatton: For the code generation, you can combine the two variable lets into a single let and get rid of the begin, for example:

```scheme
(foreach $item (let (($key ...) ($value ...)) ...body...) ...dict...)
```

In your example code generation, I don't see the value of the `dict` in the code, which makes me think that your test is failing because it has nothing to iterate over. I'd have to test more. You can also write some unit tests in Java for testing the expressions to make sure they work. See buildserver/tests/.../YailEvalTest.java. In particular, see the test starting around line 546.

appinventor/blocklyeditor/src/generators/yail/control.js
```diff
@@ -93,6 +93,50 @@ Blockly.Yail['controls_forEach'] = function() {
...
+      + yail.YAIL_SPACER + getKeyCode
+      + yail.YAIL_CLOSE_COMBINATION + yail.YAIL_SPACER;
+  var getValueCode = yail.YAIL_CALL_YAIL_PRIMITIVE + 'yail-list-get-item' + yail.YAIL_SPACER
+      + yail.YAIL_OPEN_COMBINATION + yail.YAIL_LIST_CONSTRUCTOR + yail.YAIL_SPACER
+      + getListCode
+      + yail.YAIL_SPACER + '2' + yail.YAIL_CLOSE_COMBINATION
+      + yail.YAIL_SPACER + yail.YAIL_QUOTE + yail.YAIL_OPEN_COMBINATION
+      + 'list number' + yail.YAIL_CLOSE_COMBINATION + yail.YAIL_SPACER
+      + yail.YAIL_DOUBLE_QUOTE + 'select list item' + yail.YAIL_DOUBLE_QUOTE + yail.YAIL_CLOSE_COMBINATION;
+  var setValueCode = yail.YAIL_SPACER
```
ewpatton: Ok. So on further inspection of your log I think I've found the potential problem. In the `setKeyCode` string you open the `(let (` but you never close the variable definition. You'll want to have two `YAIL_CLOSE_COMBINATION` here, one to close the `($value ...)` and a second to close the `(let (` started in `setKeyCode`. The generic form of the let is:

```scheme
(let ((var1 value1) (var2 value2) ...) code ...)
```

but since you don't have the closing parens after the variable definitions you end up with unbalanced parens, resulting in the EOF error.

BeksOmega: Awesome! Thank you for the help :D I'll take another shot at it later today!

ewpatton:
@BeksOmega Two other things:

1. Bump the BLOCKS_LANGUAGE_VERSION in YaVersion.java
2. Add an entry for the new version under the "Language" key in versioning.js

appinventor/blocklyeditor/src/blocks/control.js
BeksOmega: Q: I've seen other organizations (ex: [scratch](https://github.com/LLK/scratch-blocks/blob/b7088bb4046c3ffa4400d381238b260bce16d12e/blocks_vertical/motion.js#L37)) pass the block "shape" as json, but wrap inside a JS block definition for easy extension. I think it's a cool idea, but I'm not sure if there are downsides/what they may be. Is that something you'd be interested in?

ewpatton: I'm not really sure that it is a level of indirection that we need at this time. It's also the case that because of things like lexical scoping of variables, etc. App Inventor blocks have more complex semantics than we might be able to capture in a JSON structure (I'd need to think about that more), so while this may work for most blocks it may not work for all, and it's not clear that a complete rewrite to the JSON form would bring much benefit.

BeksOmega:
Ok I believe this is ready for another look!

Notes:
- I moved some of the generation code into functions to avoid duplication. If you don't like it I can totally revert it back.

- I think I wrote the YailEvalTest correctly, but I'm not sure. I got the following output related to it when I ran `ant tests`
```
[junit] Running com.google.appinventor.buildserver.YailEvalTest
[junit] Tests run: 82, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 12.944 sec
```

- I changed the block to use interpolateMsg, but I had to append the 'do' input separately because interpolateMsg doesn't support statement inputs.

appinventor/buildserver/tests/com/google/appinventor/buildserver/YailEvalTest.java
```diff
@@ -605,6 +605,47 @@ public void testListGroup5() throws Throwable {
...
+        "   (set-var! x " +
+        "    (call-yail-primitive + " +
+        "     (*list-for-runtime* (get-var x) (lexical-value $key) (lexical-value $value) )" +
+        "     '(number number number ) \"+\"" +
+        "    ) " +
+        "   ) " +
+        "  ) " +
+        "  (call-yail-primitive make-yail-dictionary " +
+        "   (*list-for-runtime* " +
+        "    (call-yail-primitive make-ke-dictionary-pair " +
```
ewpatton: Bad symbol here: `make-ke-dictionary-pair`

appinventor/buildserver/tests/com/google/appinventor/buildserver/YailEvalTest.java
```diff
@@ -605,6 +605,47 @@ public void testListGroup5() throws Throwable {
...
+        "     (*list-for-runtime* 1 2 ) '(key any)  \"make a pair\" " +
+        "    ) " +
+        "    (call-yail-primitive make-ke-dictionary-pair " +
```
ewpatton: Ditto.

appinventor/buildserver/tests/com/google/appinventor/buildserver/YailEvalTest.java
```diff
@@ -605,6 +605,47 @@ public void testListGroup5() throws Throwable {
...
+        "     (*list-for-runtime* 3 4 ) '(key any)  \"make a pair\" " +
+        "    ) " +
+        "   )" +
+        "  ) '(pair pair ) \"make a dictionary\"" +
```
ewpatton: This line needs to be up one. Right now it is outside of the `(let)` which is one reason why the test fails (the error is foreach expects 3 args but got 5).

appinventor/buildserver/tests/com/google/appinventor/buildserver/YailEvalTest.java
```diff
@@ -605,6 +605,47 @@ public void testListGroup5() throws Throwable {
...
+        " ) " +
+        ") ";
```
ewpatton: Before this last parens you're going to want to call `(get-var x)` so that it returns the value of the test to be checked on line 646.

ewpatton:
Also, I note that with the interpolateMsg change the value is inlined. Can we change it so that it is an external input similar to your original graphic?

ewpatton:
These are the blocks I used to test:

blocks (35)

BeksOmega:
Sorry I totally missed that the test was error-ing! Now `ant tests` outputs:
```
[junit] Running com.google.appinventor.buildserver.YailEvalTest
[junit] Tests run: 82, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 12.559 sec
```

ewpatton:

appinventor/blocklyeditor/src/field_lexical_variable.js
```diff
@@ -675,17 +675,20 @@ Blockly.LexicalVariable.renameParamWithoutRenamingCapturables = function (source
...
+    if (type == "procedures_mutatorarg"
+        || type == "procedures_defnoreturn"
+        || type == "procedures_defreturn") {
       sourcePrefix = Blockly.procedureParameterPrefix;
-    } else if (sourceBlock.type == "controls_forEach") {
+    } else if (type == "controls_forEach") {
       sourcePrefix = Blockly.loopParameterPrefix;
-    } else if ( sourceBlock.type == "controls_forRange") {
+    } else if (type == "controls_for_each_dict") {
+      sourcePrefix = Blockly.loopKeyParameterPrefix;
```
BeksOmega: Should I handle this by combining the loopKeyParameterPrefix and loopValueParameterPrefix into one loopDictParameter prefix? Or should I try and use the sourceBlock to figure out which one is being renamed?

ewpatton: I would say keep them separate for now. An earlier version of App inventor differentiated different types of variables based on their scopes, which is why the distinction is made here. Today, the only variable type that is differentiated is global variables. If we were ever to go back to differentiating different variable types then having this here would save us the work.

appinventor/blocklyeditor/src/field_lexical_variable.js
```diff
@@ -1049,9 +1083,10 @@ Blockly.LexicalVariable.getEventParam = function (block) {
     var name = prefixPair[1];
     var child = block;
     var parent = block.getParent();
+    var type = parent.type;
```
ewpatton: This code is walking up the block tree trying to find the appropriate lexical scope. This evaluation of `parent.type` needs to happen within the while loop otherwise `type` is always the immediate parent's type rather than the parent of the currently evaluated block within the loop.

appinventor/blocklyeditor/src/field_lexical_variable.js
```diff
@@ -1064,21 +1103,25 @@ Blockly.LexicalVariable.getEventParam = function (block) {
...
+           && (parent.getInputTargetBlock('DO') == child) ) { // Only DO is in scope, not other inputs!
          var loopName = parent.getFieldValue('VAR');
-           if (loopName == name) {
-             return null; // Name is locally bound, not an event parameter.
+         if (loopName == name) {
+           return null; // Name is locally bound, not an event parameter.
          }
+       } else if (type == 'controls_for_each_dict'
+           && parent.getInputTargetBlock('DO') == child
+           && parent.getVars().indexOf(name) != -1) {
```
ewpatton: I just want to confirm I understand why this check is here, which is that its purpose is to make sure that the key and value names don't collide. Is that correct? In either case, would you leave a comment for future developers why there's a check versus the other for constructs?

appinventor/docs/html/reference/blocks/control.html
```diff
@@ -168,6 +168,12 @@ <h3 id="foreach">for each item in list</h3>
 
 <p>Runs the blocks in the do section for each item in the list. Use the given variable name, <code class="variable block highlighter-rouge">item</code>, to refer to the current list item. You can change the name <code class="variable block highlighter-rouge">item</code> to something else if you wish.</p>
 
+<h3 id="foreachdict">for each key with value in dictionary</h3>
+
+<p><img src="images/control/foreachdict.png" alt="" /></p>
+
+<p>Runs the blocks in the do section for each key-value entry in the dictionary. Use the given variables, <code class="variable block highlighter-rouge">key</code> and <code class="variable block highlighter-rouge">value</code>, to refer to the key and value of the current dictionary entry. You can change the names <code class="variable block highlighter-rouge">key</code> and <code class="variable block highlighter-rouge">value</code> to something else if you wish. </p>
```
ewpatton: When I build this PR there's a whitespace difference at the end of this paragraph.

ewpatton:
@BeksOmega I think if you address my last two comments then this is ready for merging.

appinventor/blocklyeditor/src/field_lexical_variable.js
BeksOmega: So my understanding of this function is that it is used to check if a lexical get/set is accessing an event parameter, so that it can be tagged.

`parent.getInputTargetBlock('DO') == child`
Checks that the block is within the scope of vars defined by the for_each_dict block.

`parent.getVars()indexOf(name) != -1`
Checks that the variable being accessed is one of the vars defined by the for_each_dict block.

So if the `child` is within the scope && it is accessing a loop var, we return null to say that it is not accessing an event parameter.

Will add comments!

ewpatton:
Thanks @BeksOmega. This was a great contribution.

PR #2034, opened 2020-02-16
Upgrade to new D8 compiler for dexing

pavi2410:
TODO:
- [x] Test- Compare compilation speeds and APK size
- [ ] Pre-dexing - is it needed? if yes, there's a `--intermediate` flag built in d8
- [x] Use D8 for compiling extensions https://github.com/mit-cml/appinventor-sources/pull/2034/commits/16d8330371d38fa9f18022fab35cbc76d82acca1
- [x] https://github.com/mit-cml/appinventor-sources/pull/2034#issuecomment-586680787

This PR builds upon #1435


pavi2410:
This fails on Windows at the moment because the command generated for D8 exceeds the path length limit of Windows. I'm looking for solutions.

pavi2410:
Size comparison (in MiB)
---

APK | dx | d8 | diff
--- | --- | --- | ---
Companion | 15.4 | 14.7 | -4.55%
Blank App | 3.3 | 2.98 | -9.7%

ewpatton:
@pavi2410 Any thoughts on why there's a difference in app size? I expect all of the JARs we use are already compatible with dex (given that the system compiles), so the idea that d8 would make them smaller suggests it is rewriting them in a nontrivial way that is separate from the Java 8 features (which again, we don't have otherwise compilation would fail).

ewpatton:
A few more things...
* What's the purpose of the .draft version of DexExecTask?
* Have you tested this at all with SDK 7 to ensure it doesn't break the build on older versions of Android?

pavi2410:
According to this [blog post](https://android-developers.googleblog.com/2017/08/next-generation-dex-compiler-now-in.html) by Google, they claim that d8 produces smaller dex files

> When comparing with the current DX compiler, D8 compiles faster and **outputs smaller .dex files**, while having the same or better app runtime performance. 





This is maybe because they learned from their experience and tried to optimize the generated bytecode.

Also, another reason maybe that previously, two dex files were generated, whereas this time, dex files are generated on demand, utilizing most of the available method count limit.

pavi2410:
> What's the purpose of the .draft version of DexExecTask?

That was just to ignore the compilation of that file. I kept that in case we need to pre-dex files. But, I'll remove it.

> Have you tested this at all with SDK 7 to ensure it doesn't break the build on older versions of Android?

I have tested this on my phone (Samsung, Android 9). @Vishwas-Adiga has also tested this on his Android 10 device. I'm yet to test this on API < 19 when multi-dex was not introduced.

pavi2410:
@ewpatton, @Sander0542 says it fails on API 7

ewpatton:
@pavi2410 What happens if you don't include the `--release` flag? We'd want to keep the debug information for App Inventor (I realize you might want to turn it off for Kodular).

pavi2410:
Option | Description
--- | ---
--debug | Compiles DEX bytecode to include debug information, such as debug symbol       tables.       This option is enabled by default. To include debug information in your       DEX bytecode, d8 expects that the input Java bytecode       includes that information. For example, if you're using javac       to compile your code, you need to pass the -g flag to include       debug information in the output Java bytecode.       When compiling DEX files for the release version of your app or       library, instead use the --release flag described below.
--release | Compiles DEX bytecode without debug information. However, d8       includes some information that's used when generating stacktraces and       logging exceptions.                You should pass this flag when compiling bytecode for a public release.

ewpatton:
Sorry, I should have been more clear. I assume that your comparison on file sizes is using the release flag, which strips out the debugging info. I'd be interested in a comparison where the debugging info is left in.

pavi2410:
i can try that now

ewpatton:
FYI, I just tested a bunch of SDK versions and it seems like SDK 21 is the earliest version that will run this. SDKs 11-19 crash on starting the companion, and SDKs prior to 11 fail to perform dex optimization with a crash (install doesn't complete). We might be able to fix the SDK 11-19 crash, but I doubt there's much that can be done for SDK <= 10.

ewpatton:
@pavi2410 Yeah, we can always check out the older version from git if we need to, so no need to keep a copy around in the source tree.

pavi2410:
> 
> 
> Sorry, I should have been more clear. I assume that your comparison on file sizes is using the release flag, which strips out the debugging info. I'd be interested in a comparison where the debugging info is left in.

I have updated the table above https://github.com/mit-cml/appinventor-sources/pull/2034#issuecomment-587645926

pavi2410:
Seeing the results, there seems no improvement in APK size

AKBIROCK3:
but how we can configurate D8 ? let's say i need to use a feature how should i use it ?
and will you add support for R8 ?
https://developer.android.com/studio/build/shrink-code

pavi2410:
@ewpatton with almost the same change in Kodular, we are seeing interesting results in build times and apk size of the Companion app. This was run in Github action.

|   | dx | d8
--- | --- | ---
build time | 3.5 min | 2 min
apk size | 26 mb | 22 mb

pavi2410:
Xamarin also observed a significant reduction in build times and apk size with d8

https://devblogs.microsoft.com/xamarin/androids-d8-dexer-and-r8-shrinker/

ewpatton:
Since we've moved to multidex we are no longer predexing either. My two big concerns regarding d8 is that we need to make sure that 1) the companion app compiled with d8 will work in SDK 7 as we are not ready to bump the min SDK and 2) nothing is being removed from the final APK that breaks behavior due to how Kawa uses reflection for method invocations. I'm rather surprised by the fact that it is so much smaller (4 MB after compression) and would like to better understand exactly what is going away in that 4 MB.

ewpatton:
@pavi2410  Can you reconcile this branch with the AAB changes @barreeeiroo made?

pavi2410:
Yes, I will be doing that during this weekend

pavi2410:
> 2) nothing is being removed from the final APK that breaks behavior due to how Kawa uses reflection for method invocations. I'm rather surprised by the fact that it is so much smaller (4 MB after compression) and would like to better understand exactly what is going away in that 4 MB.

I would compare the generated dex files soon and share the results here.

Regarding Kawa using reflection, I would use a proguard config file to ensure that they are not tree-shaken away. But, I doubt D8 does tree-shaking and minfication because that's the role of R8.

pavi2410:
@ewpatton I'll be redoing this PR because the rebase is too complex

pavi2410:
@ewpatton This [PR](https://github.com/mit-cml/appinventor-sources/pull/2172) brings AndroidX (v1.0.0) libraries. As reported [here](https://androidx.tech/artifacts/core/core/1.0.0), the `minSdkVersion` for any AndroidX should be 14. I fail to understand how compatibility with SDK 7+ is being maintained. Can you please shed some light on this?

joseluu:
Note: this PR is missing collecting all the classes?.dex files as it is done in runMultiDex

      // Aggregate all of the classes.dex files output by dx
      File[] files = new File(dexedClassesDir).listFiles(
        new FilenameFilter() {
          @Override
          public boolean accept(File dir, String name) {
          return name.endsWith(".dex");
          }
        }
      );
      Collections.addAll(dexFiles, files);
      // Log status
      out.println(d8TimeMessage);
      LOG.info(d8TimeMessage);

PR #2036, opened 2020-02-16
Translate global variable typeblock names

### LINKED ISSUE #2035: Typeblock for variables not translated
**Describe the bug**

Typeblock for global variables does not seem to be translated. I can type in the translated name for "get" and the translated "get global name" block will not appear:
TypeblockTranslation

**Affects**

- [ ] Designer
- [X] Blocks editor
- [ ] Companion
- [ ] Compiled apps
- [ ] Buildserver

**Expected behavior**

Global variable typeblock suggestions should only appear when you type in the translated version of get/set.

**Steps to reproduce**

1. Open the blocks editor.
2. Create an "initialize global" block.
3. Switch the language to something not-english (I picked german).
4. Type in the translated version of get/set.
5. Observe how the global variable is not suggested.

BeksOmega:
### Resolves

Closes #2035 

### Description

Changes the loadGlobalVariables function so that it returns "get/set global" in the correct language.

### Testing

Repeated the reproduction steps from #2035 and was not able to replicate the bug:
TypeblockTranslationFixed

As you can see the "get/set" text is properly translated and the "global" text is properly translated.

### DIFF
--- a/appinventor/blocklyeditor/src/typeblock.js
+++ b/appinventor/blocklyeditor/src/typeblock.js
@@ -401,55 +401,38 @@ Blockly.TypeBlock.prototype.loadProcedures_ = function(){
  * example of how to call this function.
  */
 Blockly.TypeBlock.prototype.loadGlobalVariables_ = function () {
-  //clean up any previous procedures in the list
-  this.TBOptions_ = goog.object.filter(this.TBOptions_,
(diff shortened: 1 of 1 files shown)


AppInventorWorkerBee:

ewpatton:

PR #2052, opened 2020-02-21
Add AutoscaleImages property to screen1

jsuyash1514:
I've added the Autoscale images property to the Form component. Now the users can use the high-resolution images in their app.
Solves #2015 
@ewpatton Please have a look.

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/Form.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/Form.java
@@ -249,6 +249,8 @@ public class Form extends AppInventorCompatActivity
   private ProgressDialog progress;
   private static boolean _initialized = false;
 
+  private boolean autoscaleImages;
+
   // It should be changed from 100000 to 65535 if the functionality to extend
   // FragmentActivity is added in future.
   public static final int MAX_PERMISSION_NONCE = 100000;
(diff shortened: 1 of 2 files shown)


AppInventorWorkerBee:

ewpatton:
@jsuyash1514 This is a good start. I was thinking in the issue that this would be a Screen1-only property and have it govern the behavior of the entire app. Right now it would allow people to mix and match behaviors on different screens, potentially resulting in inconsistent/unexpected appearance. Take a look in MockForm.java to see how it is handled for the `ShowListsAsJson` property.

jsuyash1514:
@ewpatton I've made the required changes. Kindly have a look.

While testing the same, I've encountered the following error: 
`java.lang.IllegalStateException: no such property: AutoscaleImages`

bug

Can you please help me out?

Thank You.

ewpatton:
@jsuyash1514 You'll need to add the new property to YoungAndroidSettings.java so that it is defined.

jsuyash1514:
@ewpatton Thanks for pointing me out to the required file. I've made the necessary changes. Kindly review.

jsuyash1514:
@ewpatton I've made the required changes. Kindly review.
For overcoming the new issue, can we improve the way, we calculate the sample size? For example, we can apply either a linear search or a binary search to find the best fit sample size instead of calculating it as a power of 2.

PR #2065, opened 2020-02-25
Support to load App Inventor over https

jisqyv:
This depends on the Rendezvous server working over https, which it now does.

Change-Id: I84631fb429828a1cd7bb8f4a95707dcaa4700776


jisqyv:
@ewpatton Good idea, I'll figure out how to make the change.

jisqyv:
@ewpatton Change made...

jisqyv:
Note: This change does not force the use of https, it just provides the needed changes to serve App Inventor over https. If loaded without https, all behavior will be as it is today. The only difference is that the browser *will* always use https when communicating with the Rendezvous server (which is why the Rendezvous server has to support https as a pre-requisite). MIT's Rendezvous server already does.

PR #2075, opened 2020-03-02
Add radix number block

BeksOmega:
### Resolves

https://github.com/mit-cml/appinventor-sources/pull/2068#issuecomment-591982878

### Description

Adds a number block that has a dropdown to support input in different radix. The number automatically converts to the new radix when the number changes. For code it always outputs a decimal number.

### Testing

RadixSwitching

I also tested generating the code and it always generates the decimal version (as expected).

### Additional Information

I'm kind of unsure if decimal should be included on this block. My concern is that the other input options only allow you to input whole numbers, while decimal allows decimals and negatives. It seems like that might be confusing.

Also I tried to look into the gnu.math.IntNum that you mentioned @ewpatton, but I couldn't figure out where it was actually declared. So this version of the block does not impliment it.

Also also I haven't bumped version numbers yet, because that would conflict with #1998.


jisqyv:
@BeksOmega @ewpatton I'm not sure whether or not this needs to have the blocks language bumped. The only reason I can think of is to warn people if they load a project that contains this new block into an older version of MIT App Inventor. However they will get that warning in the next release anyway because of other changes. So I'm going to merge this change as is.

PR #2079, opened 2020-03-05
Add external issue link

barreeeiroo:
Github introduced the ability to link external sites in the issue templates. It may be worthy to give it a try, so Community could be linked to this repository.

image

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,5 @@
+blank_issues_enabled: true
+contact_links:
+  - name: MIT App Inventor Community
+    url: https://community.appinventor.mit.edu/
+    about: Looking for help while developing your apps? Ask in the Community your doubts and questions!
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -3,3 +3,6 @@ contact_links:
   - name: MIT App Inventor Community
     url: https://community.appinventor.mit.edu/c/mit-app-inventor-help/5
     about: Looking for help while developing your apps? Ask in the Community your doubts and questions!
+  - name: Extension Development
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,5 @@
+blank_issues_enabled: true
+contact_links:
+  - name: MIT App Inventor Community
+    url: https://community.appinventor.mit.edu/
```
ewpatton: I would have it point specifically to https://community.appinventor.mit.edu/c/mit-app-inventor-help/5

barreeeiroo: Fixed, it now points to that URL

ewpatton:
@barreeeiroo I think we should also have a pointer for extension developers going to the corresponding category in the community as well.

https://community.appinventor.mit.edu/c/open-source-development/extension-development/25

barreeeiroo:
> @barreeeiroo I think we should also have a pointer for extension developers going to the corresponding category in the community as well.
> 
> https://community.appinventor.mit.edu/c/open-source-development/extension-development/25

Should be done now. I thought now about adding an extra link too: https://community.appinventor.mit.edu/c/open-source-development/10
However, I also thought that it was going to be a bit odd to point issues while contributing here to an external site.
What do you think?

ewpatton:
If it becomes a problem we can add it. Most people “contribute” by making extensions.

barreeeiroo:
> If it becomes a problem we can add it. Most people “contribute” by making extensions.

Humm, so... What if the first link gets changed? Instead of directly point the root of the community, why not point it directly to https://community.appinventor.mit.edu/c/open-source-development/10?
Now that I think about it, in my opinion, it makes more sense to point it to the OS category rather than to the root. I thought that people who reach this repository is because they want to know more or contribute to the sources, and not because they want to create apps.

PR #2085, opened 2020-03-06
Replace All Mappings block

### LINKED ISSUE #2029: Implement a replace all in text using mapping block

Implement a new text block `replace all in text using mapping` that replaces substrings in a text with one or more mappings provided by a dictionary or associative list.

Currently we can use the replace all block to replace a single string with another string. However, if you want to provide potentially many mappings, you would have to implement this in a way that is not commutative with respect to the application of replacements. For example, the string "this is a string" with the mappings `{"i": "x", "xs": "y"}` should produce "thxs xs a strxng" but with a replace all within a loop approach would result in "thy y a strxng".

One can implement such a function with a relatively complicated procedure, but this is the type of exercise one would normally do in a AP CS or CS1 college level course, and therefore isn't necessarily aligned with our goal of computational action:

blocks (34)

People might often want to make multiple substitutions simultaenously and in a predictable way based on the input. A loop using a replace all block doesn't always provide the most accurate way of doing this. Further, if the replacements are cyclical, a naive implementation could result in an infinite loop (e.g., `{"I": "you", "you": "I"}`). By providing a robust abstraction we can potentially reduce/eliminate these types of issues.

elatoskinas:
# Replace All Mappings
Resolves #2029 by introduced the new replace all mappings block that applies all specified mappings to the given text with the given order.

## Block
image

`mappings` accept dictionary blocks (and list as a result)
The order specifies which order to replace mappings in, in the case of one mapping being a substring of another.

## Ordering
* **Longest string first:** replace longest first, then replace the shorter substrings in order of string length
* **Dictionary order:** replace by element order in dictionary
* **Earliest occurrence:** replace by first index of occurrence of a substring. If the indices are the same, replace by longest first.

## Example
blocks

image

## TODOs
Some work left in blocks.js (not the generator):
- [x] Update help urls
- [x] Update tooltips
- [x] Update hardcoded string constants


appinventor/blocklyeditor/src/msg/en/_messages.js
```diff
@@ -758,6 +758,14 @@ Blockly.Msg.en.switch_language_to_english = {
     Blockly.Msg.LANG_TEXT_TEXT_IS_STRING_INPUT_THING = 'thing';
     Blockly.Msg.LANG_TEXT_TEXT_IS_STRING_TOOLTIP = 'Returns true if <code>thing</code> is a string.';
 
+    Blockly.Msg.LANG_TEXT_REPLACE_WITH_MAPPINGS_HELPURL = '/reference/blocks/text.html#replacemappings';
+    Blockly.Msg.LANG_TEXT_REPLACE_WITH_MAPPINGS_INPUT_TEXT = 'text';
```
elatoskinas: TODO: Update data here

appinventor/blocklyeditor/src/generators/yail/text.js
```diff
@@ -318,3 +318,30 @@ Blockly.Yail['text_is_string'] = function() {
...
+    code = code + Blockly.Yail.YAIL_SPACER + Blockly.Yail.YAIL_QUOTE;
+    code = code + Blockly.Yail.YAIL_OPEN_COMBINATION + "text" + Blockly.Yail.YAIL_SPACER;
+    code = code + "dictionary" + Blockly.Yail.YAIL_SPACER;
+    code = code + "number" + Blockly.Yail.YAIL_CLOSE_COMBINATION;
+    code = code + Blockly.Yail.YAIL_SPACER + Blockly.Yail.YAIL_DOUBLE_QUOTE + "replace with mappings" + Blockly.Yail.YAIL_DOUBLE_QUOTE + Blockly.Yail.YAIL_CLOSE_COMBINATION;
+
+    return [ code, Blockly.Yail.ORDER_ATOMIC ];
+  };
+
+  Blockly.Yail.text_replace_mappings.OPERATORS = {
```
elatoskinas: @ewpatton Right now what I did was I took the order as a String parameter, and mapped it to an integer, which I pass in to the corresponding Java function as a parameter later on. Would this be a viable approach, or should we define a function for each mode in Scheme, and then call in to the appropriate function?

appinventor/blocklyeditor/src/blocks/text.js
```diff
@@ -497,3 +497,79 @@ Blockly.Blocks['text_is_string'] = {
...
+  return [
+    ['longest string', 'LONGEST_STRING_FIRST'],
+    ['dictionary', 'DICTIONARY_ORDER'],
+    ['earliest occurrence', 'EARLIEST_OCCURRENCE']
+  ]
+};
+
+Blockly.Blocks.text_replace_mappings.TOOLTIPS = function () {
+  return {
+    LONGEST_STRING_FIRST : "tooltip",
```
elatoskinas: TODO: Update tooltips

appinventor/blocklyeditor/src/blocks/text.js
```diff
@@ -497,3 +497,79 @@ Blockly.Blocks['text_is_string'] = {
...
+    DICTIONARY_ORDER : "tooltip",
+    EARLIEST_OCCURRENCE : "tooltip"
+  }
+};
+
+Blockly.Blocks.text_replace_mappings.HELPURLS = function () {
+  return {
+    LONGEST_STRING_FIRST : "help",
```
elatoskinas: TODO: Update help urls

appinventor/blocklyeditor/src/blocks/text.js
```diff
@@ -497,3 +497,79 @@ Blockly.Blocks['text_is_string'] = {
...
+        .appendField(new Blockly.FieldDropdown(this.OPERATORS, Blockly.Blocks.text_split.dropdown_onchange), 'OP')
+        .setAlign(Blockly.ALIGN_RIGHT)
+
+
+    this.setTooltip(Blockly.Msg.LANG_TEXT_REPLACE_WITH_MAPPINGS_TOOLTIP);
+    this.setInputsInline(false);
+  },
+  //typeblock: [{translatedName: Blockly.Msg.LANG_TEXT_REPLACE_WITH_MAPPINGS_TITLE_REPLACE_WITH_MAPPINGS}]
+  typeblock: [{
+    translatedName: Blockly.Msg.LANG_TEXT_SPLIT_OPERATOR_SPLIT,
```
elatoskinas: TODO: Update translatedName

appinventor/blocklyeditor/src/generators/yail/text.js
```diff
@@ -318,3 +318,30 @@ Blockly.Yail['text_is_string'] = function() {
   code = code + Blockly.Yail.YAIL_DOUBLE_QUOTE + "is a string?" + Blockly.Yail.YAIL_DOUBLE_QUOTE + Blockly.Yail.YAIL_CLOSE_COMBINATION;
   return [ code, Blockly.Yail.ORDER_ATOMIC ];
 };
+
+Blockly.Yail['text_replace_mappings'] = function() {
+    // String replace with segment
```
elatoskinas: TODO: Update description

appinventor/blocklyeditor/src/generators/yail/text.js
ewpatton: I think it ultimately depends on how much code overlap there is. If it's saner to keep each as a different function you could map to three Scheme functions that call through to the Java implementation. If it's easier to pass an argument, then I would say pass the String itself and not bother mapping to an integer internally.

appinventor/components/src/com/google/appinventor/components/common/YaVersion.java
```diff
@@ -499,8 +499,10 @@ private YaVersion() {
   // - FORM_COMPONENT_VERSION was incremented to 27.
   // For YOUNG_ANDROID_VERSION 201:
   // - CANVAS_COMPONENT_VERSION was incremented to 13
+  // For YOUNG_ANDROID_VERSION 202:
+  // - BLOCKS_LANGUAGE_VERSION was incremented to 30
```
ewpatton: I would recommend against bumping YOUNG_ANDROID_VERSION since there's no guarantee that this will be the next PR to be merged that affects YOUNG_ANDROID_VERSION. We will push a commit to your branch when we are ready to merge it to bump the version number here.

appinventor/components/src/com/google/appinventor/components/runtime/util/JavaReplaceWithMappings.java
```diff
@@ -0,0 +1,189 @@
+// -*- mode: java; c-basic-offset: 2; -*-
```
ewpatton: I would consider merging the logic of JavaJoinListOfStrings and JavaReplaceWithMappings into a single utility class, e.g., JavaStringUtils. I think it's better than introducing a number of very small, targeted classes that hold only one or two methods each.

appinventor/components/tests/com/google/appinventor/components/runtime/util/JavaReplaceWithMappingsTest.java
```diff
@@ -0,0 +1,162 @@
...
+    mappings.put("a", "d");
+    mappings.put("ab", "bc");
+    mappings.put("ba", "cb");
+
+    final String result = JavaReplaceWithMappings.replaceWithMappings(text, mappings, 2);
+    final String expected = "bc ed d e";
+
+    assertEquals(expected, result);
+  }
+}
```
ewpatton: Missing newline at end of file.

appinventor/components/src/com/google/appinventor/components/runtime/util/JavaReplaceWithMappings.java
```diff
@@ -0,0 +1,189 @@
...
+    Map<String, String> stringMappings = new HashMap<>();
+
+    // Construct a new List to store the Map's keys
+    List<String> keys = new ArrayList<>();
+
+    while (it.hasNext()) {
+      Map.Entry<Object, Object> current = it.next();
+
+      // Get Key & Value, and update string mappings map
+      String key = current.getKey().toString();
```
ewpatton: We need to think about what to do if the map contains both the string "1" and the number 1.

appinventor/components/src/com/google/appinventor/components/runtime/util/JavaReplaceWithMappings.java
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    // We will construct a union regex pattern from the keys
+    StringBuilder patternBuilder = new StringBuilder();
+
+    for (int i = 0; i < keys.size(); ++i) {
+      String key = keys.get(i);
+
+      // Append mapping that we want to replace to the regex pattern
+      patternBuilder.append(key);
```
ewpatton: Probably need to escape any `|` in key as `\\|`, otherwise they will be incorporated into the regex.

appinventor/components/src/com/google/appinventor/components/runtime/util/JavaReplaceWithMappings.java
```diff
@@ -0,0 +1,189 @@
...
+
+      // If there is still another mapping, then we append the union (OR) operator
+      if ((i+1) < keys.size()) {
+        patternBuilder.append("|");
```
ewpatton: Another approach to this is to have a `String sep = ""` outside the loop and then set it to `"|"` inside the loop. This way you never need to have the conditional.

ewpatton:
@elatoskinas I gave you some initial comments to go along with the TODO items. Please ping me when you think this is ready for a formal review.

appinventor/components/src/com/google/appinventor/components/runtime/util/JavaReplaceWithMappings.java
elatoskinas: For the replacement itself, it should not affect the mapping itself, since we check for the key's presence before we apply the mapping when we encounter a regex match. However, that is a very fair point, since it can potentially make the regex slower (and not entirely correct). I will update to escape the keys.

appinventor/components/src/com/google/appinventor/components/runtime/util/JavaReplaceWithMappings.java
elatoskinas: @ewpatton I'm not sure I follow? The conditional is needed so that the union/OR operator does not get appended to the very last string, for example:
a | b | c

If we did not have the conditional, it would be:
a | b | c |

I am not really sure how this can be accomplished otherwise apart from looping backwards or changing the logic to another equivalent one (with more or less equal complexity)

appinventor/components/src/com/google/appinventor/components/runtime/util/JavaReplaceWithMappings.java
elatoskinas: @ewpatton The default solution in place (according to LinkedHashMap order) is to take the most recent (last) entry in the map (so effectively, new values override older ones). I think this is fairly reasonable, since we are replacing values in text, and there is no distinction in text between a number literal, and a number character. It might be worth adding in the description of the block or in the MD documentation that this would be the expected behavior.

elatoskinas: Also might be worthwhile detecting this case explicitly, since both keys will be added to the List. Good point!

appinventor/components/src/com/google/appinventor/components/runtime/util/JavaReplaceWithMappings.java
elatoskinas: Agreed; Done!

appinventor/components/src/com/google/appinventor/components/runtime/util/JavaStringUtils.java
```diff
@@ -0,0 +1,240 @@
...
+ * Used for ease of reasoning about the solution or to address issues
+ * with Kawa (e.g. memory problems).
+ * See runtime.scm
+ */
+public class JavaStringUtils {
+  public static final String LOG_TAG_JOIN_STRINGS = "JavaJoinListOfStrings";
+  private static final boolean DEBUG = false;
+
+
+  // Implements the following operation
```
elatoskinas: @ewpatton Should we move these comments elsewhere? Since the JavaJoinWithStrings class had some loose comments around, I moved them under the method. Let me know if you'd like to re-organize this a bit. Also, I moved the majority to the corresponding Javadoc.

appinventor/blocklyeditor/src/blocks/text.js
```diff
@@ -497,3 +497,79 @@ Blockly.Blocks['text_is_string'] = {
   },
   typeblock: [{translatedName: Blockly.Msg.LANG_TEXT_TEXT_IS_STRING_TITLE}]
 };
+
+Blockly.Blocks['text_replace_mappings'] = {
+  // Replace all occurrences in mappings with their corresponding replacement
+  category: 'Text',
+  helpUrl: Blockly.Msg.LANG_TEXT_REPLACE_WITH_MAPPINGS_HELPURL,
```
elatoskinas: @ewpatton Is there any new functionality for auto-generating docs for blocks (like with the component markdown), or does it the same as before? (have to edit an .md/.html manually?)

appinventor/blocklyeditor/src/generators/yail/text.js
elatoskinas: @ewpatton I figure keeping separate functions is a bit cleaner (since we do some form of mapping in the generator anyway). I did some refactoring; Let me know if you think it's overkill (I introduce a class for ordering in JavaStringUtils)

ewpatton:
@stezelMIT I've deployed this at http://pr2085.ai2-ewpatton-temp.appspot.com if you want to try it out.

stezelMIT:
Oh wow!  Thank you! Thank you all!

appinventor/components/src/com/google/appinventor/components/runtime/CloudDB.java
```diff
@@ -1328,10 +1328,10 @@ private synchronized void ensureSslSockFactory() {
           TrustManagerFactory.getDefaultAlgorithm());
         tmf.init(keyStore);
         // // DEBUG
-        // Log.d(LOG_TAG, "And now for something completely different...");
+        // Log.d(LOG_TAG_JOIN_STRINGS, "And now for something completely different...");
```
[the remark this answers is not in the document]
elatoskinas: Ah, seems like refactor all in my IDE bit back on me a bit, and I overlooked this. Thanks for pointing it out, I will revert it.

appinventor/components/src/com/google/appinventor/components/runtime/util/AssetFetcher.java
```diff
@@ -101,7 +101,7 @@ public static void upgradeCompanion(final String cookieValue, final String input
     //           intent.setFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION);
     //           form.startActivity(intent);
     //         } catch (Exception e) {
-    //           Log.e(LOG_TAG, "ERROR_UNABLE_TO_GET", e);
+    //           Log.e(LOG_TAG_JOIN_STRINGS, "ERROR_UNABLE_TO_GET", e);
```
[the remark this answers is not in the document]
elatoskinas: Reverted

appinventor/components/src/com/google/appinventor/components/runtime/TinyWebDB.java
```diff
@@ -181,14 +181,14 @@ private void postStoreValue(String tag, Object valueToStore) {
     // commands to aid in debugging while the component is being
     // developed, and then commenting them out when the component is
     // deployed.
-    // Log.w(LOG_TAG, "postStoreValue: sending tag = " +
+    // Log.w(LOG_TAG_JOIN_STRINGS, "postStoreValue: sending tag = " +
```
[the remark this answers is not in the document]
elatoskinas: Reverted

elatoskinas:
@ewpatton That should cover the review changes above; We probably still want to add some documentation to the blocks to the website. Should this be done manually via the corresponding .md files?

ewpatton:
@elatoskinas Yes, you'll need to edit the markdown file to add new documentation. When you build the system (or run `ant docs`) it should update the HTML documentation.

elatoskinas:
> @elatoskinas Yes, you'll need to edit the markdown file to add new documentation. When you build the system (or run `ant docs`) it should update the HTML documentation.

@ewpatton Done; There is still the consideration left of how we should handle "1" and 1 cases if they are both present in the mappings (in one of the review comments left unresolved)

stezelMIT:
Screen Shot 2020-03-12 at 12 53 34 PM
Thanks to @elatoskinas and everyone working on this.
I am running into errors of sorts.  Can someone explain to me why this returns "that me were good" whereas I would expect "that I was good" as "you were" is a longer string than "at you".
Also a question in general:  When we perform a replacement by longest string first, is the replacement made immediately before going to the next longest string replacement or are the replacements done at the very end?

elatoskinas:
@stezelMIT Ah, that is an interesting edge case there. The reason why the output is as you got is because the implementation I made as of now follows regex semantics, so in fact 'at you' is found first, hence it is also replaced first. The way the replacement happens is that we find a string by an union constructed by the keys in the specified order, and then we replace them as we go; So this all happens in a single pass (as it appears to us; but the regex matcher in use might be doing more work internally). So the replacement, in a sense, is made immediately.

So for the result you got, whether it's a bug or not depends on what semantics we view this as. If we put the regexp into a matcher, this is the result:
image

But I believe for our use cases it would make far more sense to actually replace the longest one first. I'll take a look at this.

stezelMIT:
Hi @elatoskinas,
Thank you again so much for your effort in this.
From the user's point of view when they select "longest string first" order (or the "dictionary" order) there is no ambiguity about what they need, so I am not sure how the following can be maintained as part of the replacement all algorithm:

  **"underlying order of matching is basically leftmost first, meaning that irregardless of the string length, we will always match by the key that appears first, and replace that first."**

I would assume first in the algorithm the strings would be sorted in the order requested by the user and then the reg exp procedure would be applied in that order.

@ewpatton if you have any wisdom to share (after your vacation is over, not before) kindly give us some guidance.
Thank you all

elatoskinas:
@stezelMIT I agree on the points you've mentioned. For the majority of use cases, it would make sense to replace in the order as specified in the block. However, it is also worth noting that this has the trade-off that now the order is completely altered, and we no longer have the ability to replace in the order of occurrence in the original string (which I believe has quite a few use cases as well, namely if the user wants to indeed apply a direct regex union operation as currently implemented). If we do consider the previously discussed "earliest occurrence" option, then the block becomes complicated in the sense that we again need 2 options for it ("longest string first" and "dictionary order").

Thus, for the sake of ambiguity, I would agree that perhaps changing it to the behavior you mentioned would make more sense. However, we should also take the trade-off into account, and perhaps it might be worth thinking about (any) implications caused by choosing one option over the other (here we will lose use cases in either way, unless we make a more complex block, but maybe one method is better than the other)

stezelMIT:
Dear @elatoskinas,
I believe the general sentiment here at App Inventor, as indicated by Hal's earlier comments, is to keep things simple as our main audience are K-12 students and teachers.  There are no plans at this point to implement "earliest occurrence" or other complex possibilities.  The replace all block just needs to do what it promises: make replacements according to either the "longest string first" preference or "dictionary order first" preference.  
@ewpatton Evan is due Wednesday and might be able to give more technical advice on how to make the implementations within the reg ex framework.  Thank you again so much for your interest in helping us out with this.

elatoskinas:
@stezelMIT Fair points indeed. I agree with you fully -- we should go to the suggested route to keep things simple, and implement a more simplified version (favoring the specified order) in the block. My initial confusion was in the sense that the pure regex implementation will do earliest occurrence first replacement by default (and actually, I don't think there's order redefinition), so there is some tweaking to do there to come to a more adapted solution to fit the needs. I will come back to a solution here soon.

@ewpatton 
For a more technical description, right now (internally) I'm thinking of the following:
1. Use regular expressions to find all matches, in the form of string index intervals [start, end]
2. Order the intervals by the specified ordering (dictionary/earliest occurrence first)
3. Replace the strings in the intervals by the mappings, preserving conflicts (if any character in interval is already replaced, we do not apply replacement)

Perhaps I'm overlooking into this and there is a more straightforward solution; But I feel like we need a manual adaptation here so as to preserve the replacements not being applied on top of each other.

stezelMIT:
1.  Order the intervals by the specified ordering (dictionary/earliest occurrence* first)
I think you mean * “longest string first” here.
Thank you.
Selim

elatoskinas:
@stezelMIT Ah yes, you are right. I seem to be mixing up the terms in that regard!

I have implemented a new solution that no longer uses the standard regex order, but rather the redefined order to fit our needs. So now in the block you showed above, the result will be: "that I was good" indeed, since we first apply the "you were" mapping, and then no more mappings remain to be applied, since no more parts of text remain where the other mappings fit.

@ewpatton Interestingly enough, the change was quite complex. The idea that I did now is as follows:
**Issue:** The issue with regex order is that we cannot really redefine it. E.g. if I have the string `abcd`, and I have the union `bc|ab`, `ab` will be found first, no matter the order, because regex evaluates it from start to end (maybe there's a clever trick with lookaheads that I'm not aware of, but from what I could tell, regexp order cannot really be redefined due to the nature of it)

So the solution that I had in mind and implemented is to use, more or less, an interval tree data structure (I used `TreeSet` from Guava, since we already seem to have google commons in the codebase). The idea is as follows:
1. All the steps to constructing the post-processed map of mappings remains the same. We also re-order the keys in the same order (in the List). Then, we call in a function with our text, mappings and ordered key list to apply the mappings.

2. The first step is to process each key individually. We do this using the regex patterns & matchers, but now, instead of taking a union, we do it on a key-by-key basis (we process one key, then the next one, and so on). We do this because of the regex issue specified above.

3. In this key processing step, we can use the matcher to find the next occurrence. From this occurrence, we can extract the start & end indices. We construct a `Range` object from these indices (to use in our `RangeSet`), and we check whether the range overlaps with all the intervals in the `RangeSet` (if the RangeSet is implemented as per the Interval tree data structure, this should be `O(log n)` time. Provided the range does not overlap, we add it to the `RangeSet`

4.In addition to the above, we will keep a Map (let's say `ReplacementMap` from `Range` to `String`. This essentially tells us what range of the string we want to replace with which `String` for ease of use. This map will be ordered in **descending** order of the range endpoint (it will be a `TreeMap`)

5.After processing all the keys, we iterate the `ReplacementMap`'s entries, and apply the replacement in order. We use the combination: ` L + M + R `, where ` L ` is the string of text up until the current range's lower bound index, `M` is the replacement string (the value of the mapping) and `R` is the remainder of the string after `M`.

There is one key observation to note, specifically the map's chosen order in step 4. If we do ordering in a different way, we will run into issues that the ranges will no longer be accurate (we will need to update their indices). For example, consider:
Ranges: `[0 4], [5, 6]`. Suppose we replace `[0, 4]` with a substring of size 7. Well, now all the other ranges have shifted by `3`, and this becomes harder to keep track of.
However, if we replaced starting from `[5,6]`, the indices of `[0,4]` do not change, because we only replaced the right end of the string.
And because the ranges don't overlap, this ordering should be correct.

That was quite lengthy, but this solution seemed to work for the test cases that we've had so far. I've tried to document this as thoroughly as possible in the current code.

ewpatton:
@stezelMIT I've pushed the new version to the test server.

@elatoskinas I'm not able to compile apps with this version. Companion works okay though.

elatoskinas:
> @elatoskinas I'm not able to compile apps with this version. Companion works okay though.
@ewpatton This  seems to be an extremely weird problem, one that might seem to cause some indirect crashes.. I identified the crash to be caused in the `applyMappings` method that does all the heavy-lifting, as commenting out did not crash the app building.

However, I cannot seem to understand the following phenomenon. This does not crash:
```java
    // Create a set of ranges to keep track of which index ranges in the
    // original text string are already set for replacement.
    RangeSet<Integer> ranges = TreeRangeSet.create();

    // Map to map Range to String to replace with.
    // E.g. [1, 3) -> 'abcd' indicates that the substring from 1 to 3 exclusive should
    // be replaced with the string 'ab'.
    // Note that the length of the string does not matter.
    Map<Range<Integer>, String> replacements = new TreeMap<>(new Comparator<Range<Integer>>() {
      @Override
      public int compare(Range<Integer> r1, Range<Integer> r2) {
        // Sort in descending order
        return Integer.compare(r2.upperEndpoint(), r1.upperEndpoint());
      }
    });

    // Range construction step: Iterate through all the keys,
    // and fill in ranges set & replacements map.
//    for (String key : keys) {
//      // omitted commented out code
//    }
 
     // more commented out code here ommited

    return text;
```

Yet, this crashes:
```java
    // Create a set of ranges to keep track of which index ranges in the
    // original text string are already set for replacement.
    RangeSet<Integer> ranges = TreeRangeSet.create();

    // Map to map Range to String to replace with.
    // E.g. [1, 3) -> 'abcd' indicates that the substring from 1 to 3 exclusive should
    // be replaced with the string 'ab'.
    // Note that the length of the string does not matter.
    Map<Range<Integer>, String> replacements = new TreeMap<>(new Comparator<Range<Integer>>() {
      @Override
      public int compare(Range<Integer> r1, Range<Integer> r2) {
        // Sort in descending order
        return Integer.compare(r2.upperEndpoint(), r1.upperEndpoint());
      }
    });

    // Range construction step: Iterate through all the keys,
    // and fill in ranges set & replacements map.
    for (String key : keys) {
      // omitted commented out code
    }
 
     // more commented out code here ommited

    return text;
```

The only difference is that I uncommented out the for-loop there, which has an empty body. Is this some strange memory issue that we are running into? (although, would assigning one String of memory really break it in such a way?)

I will try to look at it tomorrow again with fresh eyes, but this seems very strange.

ewpatton:
@elatoskinas It's not even to that point. I mean that when I run the dev server and build server, even with a blank project I get an error in the Kawa stage at compiling.

appinventor/components/tests/com/google/appinventor/components/runtime/util/JavaStringUtilsTest.java
```diff
@@ -172,9 +173,102 @@ public void testEscapeUnionCharacter() {
...
+
+    assertEquals(expected, result);
+  }
+
+  /**
+   * Test case to verify that replacing mappings in a non-empty
+   * String with a mapping that maps an empty string to some String
+   * replaces all gaps in the String.
+   *
+   * TODO: Is this expected behavior?
```
ewpatton: This is expected behavior as at each index in the string the empty string trivially matches.

appinventor/components/src/com/google/appinventor/components/runtime/util/JavaStringUtils.java
```diff
@@ -0,0 +1,431 @@
...
+
+  /**
+   * Auxiliary function to apply the mappings provided to the given
+   * matcher that has some regex pattern loaded.
+   *
+   * @param matcher  Matcher representing matcher of Pattern on text
+   * @param mappings String -> String mappings to replace from Key to Value
+   * @return String with mappings applied
+   */
+  /*private static String applyMappings(Matcher matcher, Map<String, String> mappings) {
```
ewpatton: Please remove unused code if it is no longer necessary.

appinventor/components/src/com/google/appinventor/components/runtime/util/JavaStringUtils.java
```diff
@@ -0,0 +1,431 @@
...
+
+  /**
+   * Auxiliary function to build a String of keys combined via
+   * regex union operators.
+   * The keys are escaped to prevent them being interpreted as regex.
+   *
+   * @param keys  List of keys to union together
+   * @return  Single Regex String of keys combined by unions
+   */
+  /*private static String buildUnionFromStrings(List<String> keys) {
```
ewpatton: Ditto

elatoskinas:
> @elatoskinas It's not even to that point. I mean that when I run the dev server and build server, even with a blank project I get an error in the Kawa stage at compiling.

@ewpatton Yes; that's correct. But what I meant is, commenting the code as above will no longer make it crash at that stage. Meanwhile, uncommenting just one for-loop with a blank body will crash prior to the compilation stage, which is very odd.

ewpatton:
@elatoskinas Let me see if I can repro that on my end.

elatoskinas:
@ewpatton I found the underlying cause. When I did the following instantiation:
`RangeSet<Integer> ranges = TreeRangeSet.create()`

`RangeSet` being a superclass of `TreeRangeSet`, for some reason this made the apk builder act up and crash. Weirdly enough, replacing with:
`TreeRangeSet<Integer> ranges = TreeRangeSet.create()`

Solved the problem.

This is a very weird subtle bug.. At this point I'm really confused whether this is a Kawa, Guava or a build issue in AI locally

ewpatton:
@elatoskinas Thanks for the sleuthing. Confirmed that your most recent change fixes the issue. Can you go ahead and merge ucr to resolve the conflicts?

elatoskinas:
> @elatoskinas Thanks for the sleuthing. Confirmed that your most recent change fixes the issue. Can you go ahead and merge ucr to resolve the conflicts?

Done. However, we still need to resolve the EDIT in the comment above, which occurs when using the new block in compiled .apks:
image

I see that Guava is in the build.xml for components, but somewhat the Guava class does not get resolved. Is this somehow related to the way runtime.scm calls into Java methods, and do we also need to update a build.xml in that regard?

ewpatton:
@elatoskinas Ah, right. The guava library is conditionally included for certain components. It's not included in all builds, which is probably why builds for simple apps failed but it worked for the companion. You may be able to copy the TreeRangeSet source into the com.google.appinventor.components.runtime.collect package. I don't know how tightly coupled it is to the rest of the Guava system though.

elatoskinas:
@ewpatton It seems to be quite coupled in this regard (at least ~9 classes, although some of them are annotations, but the major problem is the classes that are subclassed from), so copy pasting it directly is not as straight forward it seems.

Do we have other alternatives before we resort to this? Also dropping Guava to implement an interval tree is an option (although will end up being tedious, complex and hard to maintain), or we might look for other solutions if this doesn't work out.

ewpatton:
I think it wouldn't be that difficult to implement the logic with TreeSet and a custom comparator that sorts the ranges by start index.

elatoskinas:
> I think it wouldn't be that difficult to implement the logic with TreeSet and a custom comparator that sorts the ranges by start index.

@ewpatton It depends. If we are aiming for a naive solution that does the job, then yes, it's probably not too difficult (although could lead to range re-combination solutions that would ultimately lead to beyond quadratic, cubic possibly time in the usage of the block).

If we want the solution to be a bit more scalable (although, understandably, that typically is not a concern for the majority of the use cases), the interval tree is quite complex to implement from scracth, in the sense that it will be hard to maintain (e.g. the [TreeRangeSet](https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/TreeRangeSet.java) in Guava has 44 lines of code roughly for the add method alone. When we add everything we need, we have 200-300 more for a supplementary data structure alone).

We could always go with the more naive approach, but if possible, I am always up for re-using solutions that already exist, are stable and fully tested. I will try to look through the Guava class again and see if maybe we can somehow extract only the needed bits while dropping some functionality, provided that conditionally including this class straight form Guava is not an option

ewpatton:
One problem is that we don't have a way of conditionally including a library for a specific block. Another is that Guava is pretty big to provide support for just a single block. Unless I'm mistaken, it looks like you only need the TreeRangeSet to test whether there is overlap between the current match and existing matches. This should be really quite doable within the existing Java constructs like TreeSet and its ability to return iterators to elements before/after a given value.

elatoskinas:
@ewpatton Agreed on Guava; If conditional inclusion is not an option, then it is indeed to big for our sake. Also, you had a very good point with TreeSet; For some reason I read TreeRangeSet initially and thought you meant re-implementing the logic posed there somehow. But using overlap checks in the comparator, the TreeSet indeed suffices. I have pushed the rewritten changes to no longer use Guava.

appinventor/components/src/com/google/appinventor/components/runtime/util/JavaStringUtils.java
```diff
@@ -290,18 +310,28 @@ public static String replaceAllMappings(String text, Map<Object, Object> mapping
...
-
-    // Map to map Range to String to replace with.
-    // E.g. [1, 3) -> 'abcd' indicates that the substring from 1 to 3 exclusive should
-    // be replaced with the string 'ab'.
-    // Note that the length of the string does not matter.
-    Map<Range<Integer>, String> replacements = new TreeMap<>(new Comparator<Range<Integer>>() {
+    // original text string are already set for replacement, together with the
+    // text to replace with. We sort this TreeSet in descending order to preserve
+    // indices of all ranges after replacement.
+    TreeSet<Range> ranges = new TreeSet<Range>(new Comparator<Range>() {
```
ewpatton: Minor nit: You could make this `static final` on the class so we only instantiate one.

elatoskinas: You mean the comparator, right?

ewpatton: Yes, the comparator only needs to be constructed once.

ewpatton:
I tested this in both companion and compiled apps. The approach seems to work as expected.

ewpatton:
Just waiting on the change w.r.t. the comparator and then I think we can merge it.

appinventor/components/src/com/google/appinventor/components/runtime/util/JavaStringUtils.java
elatoskinas: Done!

appinventor/components/src/com/google/appinventor/components/runtime/util/JavaStringUtils.java
```diff
@@ -140,11 +165,13 @@ public Range(int start, int end, String text) {
   private static final MappingOrder mappingOrderDictionary;
   private static final MappingOrder mappingOrderLongestStringFirst;
   private static final MappingOrder mappingOrderEarliestOccurrence;
+  private static final Comparator<Range> rangeComparator;
 
   static {
     mappingOrderDictionary = new MappingOrder();
```
ewpatton: Is there any reason you need these to be initialized in a `static` block rather than just assigning them when the variable is declared?

elatoskinas: Good point; There's no reason to have it anymore, before that I used it when the code was structured in such a way that required adding things to a data structure. I've removed it.

PR #2104, opened 2020-03-19
Fix: OnDestroyListener doesn't fire

pavi2410:
https://community.appinventor.mit.edu/t/ondestroy-not-fired/5341


jisqyv:
@pavi2410 I just rebased onto ucr

PR #2113, opened 2020-03-23
Permit Companion to install from https

jisqyv:
When the Companion is used to scan a bar code, if it finds a string that
begins with http:// it calls the package installer (or in the case of
the Play Store version, it hands it off to the on-phone browser). This
functionality was added a while ago to deal with the case where people
used the Companion to scan the barcode resulting from a package
build. Basically people got confused between scanning a QR Code for
connection to the browser and scanning the QR code of a packaged app.

However it turns out that this was a useful feature because it lets
people use the Companion to install their created apk’s so they do not
have to have a separate bar code scanning app installed. This is
particularly useful in a school setting in that it reduces the amount of
work that staff have to do to setup devices.

This commit updates the code to also call the installer when the string
begins with https as well as http. Given that we will soon serve MIT App
Inventor over https, packaged URL’s will soon begin with https.

Change-Id: I5797f998749b1a309c9504088fa03d5aaf4eb06e


jisqyv:
Note: This PR was quasi-mechanically created by editing the Companion itself in the blocks editor and then exporting the resulting change to the .bky and .yail files.

PR #2116, opened 2020-03-26
Move to Trash menu item changes to Delete from Trash

### LINKED ISSUE #2107: Move to Trash menu item should change to Delete from Trash
The drop-down menu item My Projects->Move to Trash should change to Delete from Trash when viewing the Trashcan.

- [X] Designer
- [ ] Blocks editor
- [ ] Companion
- [ ] Compiled apps
- [ ] Buildserver

beinganonymous:
- [x] My code follows the:

For all other changes:

What does this PR accomplish?

``Move To Trash`` menu item will change to ``Delete From Trash`` Button when ``View Trash`` Button is Clicked 

Fixes #2107  

mit2


AppInventorWorkerBee:

SusanRatiLane:

appinventor/appengine/src/com/google/appinventor/client/TopToolbar.java
```diff
@@ -180,16 +180,31 @@ public TopToolbar() {
...
+	   fileDropDown.replaceSpecificItem(item, 6);
+    }
+  } 
 
   public void updateMenuState(int numSelectedProjects, int numProjects) {
     boolean allowDelete = !isReadOnly && numSelectedProjects > 0;
     boolean allowExport = numSelectedProjects > 0;
     boolean allowExportAll = numProjects > 0;
-    fileDropDown.setItemEnabled(MESSAGES.trashProjectMenuItem(), allowDelete);
+    fileDropDown.setItemEnabled(MESSAGES.trashProjectMenuItem(), true);
```
SusanRatiLane: Did you intend to remove the allowDelete check here? AllowDelete is set to true only if the user has permissions to trash projects, which is not always the case.

appinventor/appengine/src/com/google/appinventor/client/TopToolbar.java
```diff
@@ -439,6 +454,8 @@ public void execute() {
   private static class SwitchToProjectAction implements Command {
     @Override
     public void execute() {
+	  int flag = 0;
+	  Ode.getInstance().getTopToolbar().updateMoveToTrash(flag);
```
SusanRatiLane: Again, no need for the flag here.

appinventor/appengine/src/com/google/appinventor/client/widgets/DropDownButton.java
```diff
@@ -246,6 +246,20 @@ public void replaceLastItem(DropDownItem item) {
     items.remove(items.size()-1);
     items.add(menu.addItem(item.caption, true, item.command));
   }
+  
+  public void replaceSpecificItem(DropDownItem item, int num) {
```
SusanRatiLane: Rather than change the Text and action of a menu item, we'd prefer to create both menu items and then make them enabled/visible as appropriate.

beinganonymous: @SusanRatiLane Sorry for late replies, I was busy with my GSoC Proposal.
So will it be okay to add the ``Delete From Trash`` button to the ``My Projects`` file DropDown?

SusanRatiLane: @beinganonymous Yes. Just make it invisible and disabled when not in use.

beinganonymous:
@SusanRatiLane following changes implemented in recent commit
 - [x] Rather than changing the Text and action of a menu item, created both menu items and then make them enabled/visible.
 - [x]  Added ``allowDelete`` Check
 - [x]  Removed all toggle flags

beinganonymous:
@SusanRatiLane Please do review whenever you have the time.

SusanRatiLane:
@beinganonymous Sorry for the delay. I was looking at it yesterday, and I'm testing today.

SusanRatiLane:
@beinganonymous I note two things in testing:

1. After I delete a project from trash and return to the My Projects view, the menu item does not always switch back to Move to Trash. 

2. More concerning, I can use the menu item to delete a project from the My Projects view.

It's probably worth adding a check to be sure that the project is actually in the trash before deleting it.

A third thing -- When a project is open, the Move to Trash menu option is present but disabled. We should be able to send the active project to trash and then return to the My Projects view. If this is straightforward to do, we should add it to this PR. Otherwise, I can open a separate issue to add this feature.

beinganonymous:
@SusanRatiLane No need to open a separate issue to send the active project to trash, I'll cover this feature in this PR.

beinganonymous:
@SusanRatiLane Implemented following changes
 - [x] Menu Item will switch back to ``Move To Trash`` after returning to My Project View.
 - [x] ``Move To Trash`` menu item will be enabled when the project is active.

``Delete From Trash`` menu item will not be used from My Project view, hence no need to check that the project is actually in the trash before deleting it.

beinganonymous:
@SusanRatiLane please do review whenever you have the time.

SusanRatiLane:
One last detail: The Move To Trash option grays out when there are no projects selected, but the Delete From Trash option stays enabled and shows a message to the user if it is clicked with no projects selected. Delete From Trash should behave like Move To Trash.

Otherwise, this checks out.

beinganonymous:
@SusanRatiLane Updated ``Delete From Trash`` button, it will be gray out if the project is not selected

beinganonymous:
@SusanRatiLane please do review the changes whenever you have time.

SusanRatiLane:
@beinganonymous I've been chatting with the team about the DeleteForeverProjectAction. We already have one on ProjectToolbar, but it's in a completely different scope from this one. We don't want to duplicate the code and are probably going to make this action its own class so that it can be accessed both places. I'm going to take care of that, but it will delay getting the PR merged.

Other than that, the final thing I see is that all strings shown to the users should be internationalized. I think all that we need to do here is change "Move to Trash" and "Delete From Trash" to MESSAGES.trashProjectMenuItem() and MESSAGES.deleteFromTrashButton(). These aren't buttons, but the strings should match.

beinganonymous:
@SusanRatiLane  Sounds great. I'll make strings internationalized

SusanRatiLane:
@jisqyv This should be ready to merge, but I'd appreciate your eyes on it, since I had to resolve a conflict with the new Gallery code.

SusanRatiLane (approved):

PR #2118, opened 2020-03-26
Remove unnecessary string append calls

pavi2410:


pavi2410:
I found this while reviewing [MIT App Inventor Extensions](https://github.com/mit-cml/appinventor-sources/commit/ededd5ae9733866dc3da88ddb9e3de655dbf11a1) commit. 

https://github.com/mit-cml/appinventor-sources/commit/ededd5ae9733866dc3da88ddb9e3de655dbf11a1#r38353830

I looked to me that the author replaced the previous values with empty strings and forgot to clean it up.

PR #2170, opened 2020-04-20
ListView Adapter Refactor (multi-highlight fix & preserve programatic highlight)

elatoskinas:
Refactored ListView to use custom Adapter to draw views based on selection.

Addresses issue described in: https://www.google.com/url?q=https://community.appinventor.mit.edu/t/listview-problem-with-selection-mark/7368?u%3Dewpatton&sa=D&source=hangouts&ust=1587499474271000&usg=AFQjCNEaCBPQSRdsuHJTrsWnf7hUnWbMjg

Additionally addressed another special case of ListView selection color to change immediately after programatically updating it.

## Files
.aia for testing:
[listview-test.zip](https://github.com/mit-cml/appinventor-sources/files/4506599/listview-test.zip)

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/ListView.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/ListView.java
@@ -6,11 +6,11 @@
 
 package com.google.appinventor.components.runtime;
 
+import android.content.Context;
 import android.graphics.Color;
 import android.graphics.drawable.ColorDrawable;
 import android.graphics.drawable.Drawable;
 import android.graphics.drawable.GradientDrawable;
-import android.graphics.drawable.StateListDrawable;
 import android.text.Editable;
 import android.text.Spannable;
 import android.text.SpannableString;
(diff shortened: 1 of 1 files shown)


elatoskinas:
This change introduces an interesting side effect.. Basically whenever a ListView item is clicked, it gets highlighted for a bit (as in, it became a bit more responsive); Seems like the color is always orange. We might want to keep this feature, and implement a property for `SelectionHighlightColor` or similar.

elatoskinas:
> I found an issue but I'm not sure the best way to replicate it. I created a listview with the 26 letter alphabet as elements and had a Clock.Timer that randomly picked a cell once per second and selected it. If a cell ended up being selected off screen, it's background was exactly as wide as the text, and then when the cell is reused for a visible cell, it does not fill the entirety of the listview's width.
> 
> Edit: Maybe after setting the background drawable we need to invalidate the view in the adapter?

@ewpatton 
Sorry for the late late reply, there's quite a bit going on my end for the past months, so I missed this PR

I tried building once more and revalidating on different setups, but I can't seem to reproduce this, I've tried on the following platforms, none of which yielded the described behavior:
* Android 7.0 physical device (API level 24) [LG G6]
* API29 emualated AVD [Pixel 2]
* API14 emulated AVD


So it's a bit of guess work on my end trying to pinpoint the issue, it does seem like it might be a lack of invalidation somewhere, but there are multiple things that we'd have to try to invalidate; Were there any other conditions/variables that could've came into play to cause/reproduce this error?

By the way, this was the block that I added to the `aia` linked in the PR to test this:
image

.apk for reference, if needed: [ListViewTest.zip](https://github.com/mit-cml/appinventor-sources/files/5105639/ListViewTest.zip)

PR #2181, opened 2020-05-04
Improve "Do It" comments

BeksOmega:
### Description

Previously triggering a "do it" on a block would move the comment and scroll to the bottom of the text. This wasn't a great UX, especially when moving the text meant your "do it" result wasn't visible.

Now the comment stays in the same place:
DoItAgain

I also added i18n so that the label is in the user's chosen language:
MacheEs
DoIt

But as always I'm happy to revert.


BeksOmega:
@ewpatton Added and retested!

PR #2182, opened 2020-05-07
Fix HELPURL for radix number block

conorshipp:
The URL was pointing to the block tooltip instead.


BeksOmega:
Thanks for the fix @conorshipp !

PR #2207, opened 2020-06-09
Make sure autoload property propagates

jisqyv:
Make sure the “autoload” query parameter is passed through all of the
various login paths.

Change-Id: I445355d325f2e5d1964bda270b52c4ef6d43e0dd


jisqyv:
Note: This is only part of the fix. The loginserver integration occurs on the branding branch, so there is another part that needs to be applied there (though it follows the same pattern as you find here).

PR #2248, opened 2020-07-19
Removed redundant </center> tag from login.jsp

xuwenbin:
Removed orphan </center> tag.


AppInventorWorkerBee:

PR #2261, opened 2020-07-27
Google sheets component rebased

tomiyee:
A Google Sheets Component to replace the (now discontinued) FusionTables Component.

Read Blocks are credential optional. 
Write Blocks require that the developer create a Google Service Account and provide the credentials.json file to App Inventor.


AppInventorWorkerBee:

jisqyv:

halatmit:
Let's make sure to test this thoroughly

ewpatton:
The latest version of this PR is available here: http://sheets.ai2-ewpatton-temp.appspot.com

jisqyv:
Note: Because this PR is replacing libraries that are used by multiple components. Those components need to be identified and completely tested.

jisqyv:
This PR contains an older and incompatible version of Evan's Multidex change. 8969d1e4cf should be removed from this PR.

jisqyv (dismissed):
8969d1e4cf should be removed from this PR. This can be done via an interactive rebase (or using git rebase with the appropriate obscure arguments). I can also do this for you, provided I can push to your branch. Emacs "magit" has tools that makes this easy.

ewpatton:
IIRC Tommy and I did this earlier and the only two components it seemed to potentially touch were the FusiontablesControl (obsolete) and Texting (for GV auth). It might be good to double check that we didn't miss any.

jisqyv:
Not to add work, but...

All the Read and Write routines have a common pattern. The work to be done is set up, and then an asynchronous task is triggered to do the work that then calls a completion function on the UI thread.

Couldn't the asynchronous task be generalized and then used from all the various Read/Write calls (well, there may be one generalized task for reading and another for writing)? I believe in Java you can pass a method to call, so there can still be separate completion methods for each operation.

Just thinking...

fabform:
Handy to [find google sheet id](https://fabform.io/a/how-to-find-google-sheet-id)

jisqyv:
So I'm going to do a detailed review and test of this component this weekend!

jisqyv:
So, I rebased this change (again), squashed it down to one commit (signed). I also updated YOUNG_ANDROID_VERSION, I reverted the changes to Compiler.java, they were only whitespace. I fixed a document reference (removed the "ai2.appinventor.mit.edu part) so it works. I also checked in the updated documentation that results from a build. I also fixed the one merge conflict, which was trivial. After the unit tests run, and some last minute tests I want to run, I'll approve and merge this change!

jisqyv (approved):
LGTM

PR #2289, opened 2020-08-22
Removed unnecessary request of Read/Write permissions

vknow360:
There were some permission related issues in the File component so I have tried to fix them.
Let me know if something is still wrong.


wxbit:
You should not apply for permissions in the constructor,only apply before use.

vknow360:
Hi @wxbit 
Yes, but I am updating them after getting permissions.
I forgot that AI2's File and Java's File may conflict so sorry for that.I will fix it as soon as possible.

vknow360:
Fixed previous issue.
But it seems LegacyMode() method was added after this PR so I will have to fix that too. 😅

PR #2335, opened 2020-10-23
AvailableCountries to AvailableCountriesAndRegions

puneetkathar1:
My first open source contibution.
Kindly merge it if you are satisfied !!


AppInventorWorkerBee:

conorshipp (changes requested):
You cannot just change the name of a block without writing an 'upgrader' otherwise existing projects that use the block will not load.

The documentation for upgraders can be found [here](https://docs.google.com/document/d/1ukpZXcd0fIn1fV0cNp4J00wrfwwA1daswU1yY6dqqfs/pub#h.2g0jyctaq6go). You will need to make use of the `changePropertyName` function in `versioning.js`.

PR #2341, opened 2020-11-22
Add Gallery Readonly Flag

jisqyv:
When set we do not display the “Send to Gallery” Button, only the “Login
to Gallery” button. The Gallery itself enforces read only status, so if
the button is enabled but the gallery is in read only (for this
instance) then the user will receive an error when they attempt to send
a project to the gallery. But it is better if we can just avoid the
issue by not providing the button in the first place.

Another implementation here could be to have the server side actually
check in with the gallery itself to determine read-only status, but for
now we provide it as a config flag so we do not have to work out the
details of how and when the server would communicate with the Gallery as
well as how to handle errors and time-outs etc.

Change-Id: Ib6ca63ab2ca63636d7451762f782050dbffd9b3c


ewpatton (approved):
LGTM

PR #2353, opened 2020-12-19
Fix: Load Tutorial URLs with https

pavi2410:


ewpatton:
I don't think this is sufficient as it assumes that existing URLs already use https in projects, which they don't. What I think we need to do instead is compare the current page scheme to the scheme of the tutorial URL and then make the URL HTTPS if the current page is HTTPS (or if the original tutorial url was, but this is a no-op). This way if we roll out legacy.appinventor over http tutorials will continue to work without having to provide two different AIAs.

pavi2410:
Would protocol-independent URLs work? `//:appinventor.mit.edu`

appinventor/appengine/src/com/google/appinventor/client/Ode.java
```diff
@@ -2431,7 +2431,7 @@ public boolean isTutorialVisible() {
   }
 
   public void setTutorialURL(String newURL) {
-    if (newURL.isEmpty() || (!newURL.startsWith("http://appinventor.mit.edu/")
+    if (newURL.isEmpty() || (!newURL.startsWith("https://appinventor.mit.edu/")
```
pavi2410, line 2434:
```suggestion
    if (newURL.isEmpty() || (!newURL.startsWith("http://appinventor.mit.edu/")
            && !newURL.startsWith("https://appinventor.mit.edu/"))) {
```

@ewpatton Will this work?

ewpatton:
I think what we want is to use HTTPS if either the current page is HTTPS or the TutorialURL is HTTPS. If App Inventor is served over HTTPS, the tutorials will also have to be served over HTTPS else the browser won't load them. Like wise, if the tutorial URL is HTTPS then we can't necessarily count (and shouldn't) downgrade to HTTP just because App Inventor is served over HTTP.

I think your logic in the proposed change to the if statement is good, but we will want to apply the above logic to adjust the URL to compensate for the presence of HTTPS on the host page.

pavi2410:
If I am not mistaken, then this should be the truth table for the behaviour you asked

Current page | Tutorial URL | Effective URL
--- | --- | ---
HTTPS | HTTPS | HTTPS
HTTPS | HTTP | HTTPS
HTTP | HTTPS | HTTPS
HTTP | HTTP | HTTP

ewpatton:
Right. It's effectively an inclusive or with HTTPS==true and HTTP==false

appinventor/appengine/src/com/google/appinventor/client/Ode.java
```diff
@@ -2431,16 +2431,23 @@ public boolean isTutorialVisible() {
   }
 
   public void setTutorialURL(String newURL) {
-    if (newURL.isEmpty() || (!newURL.startsWith("http://appinventor.mit.edu/")
-        && !newURL.startsWith("http://appinv.us/"))) {
+    boolean isUrlAllowed = !newURL.isEmpty()
+      && (newURL.startsWith("http://appinventor.mit.edu/")
```
ewpatton, line 2435: Code style: indent should be +4 here.

ewpatton (commented):
I've tested this and it works fine. Once you fix the one style issue we can merge this.

ewpatton (approved):
LGTM

PR #2406, opened 2021-02-01
Update CloudDB auth token

jisqyv:
Add two boolean fields. The “monall” field (in production for a while)
permits the monitoring of all key changes. This is used for statistics
gathering and research.

The “privileged” flag permits commands that manage the overall server. At
the moment the only such command is one to manually close specified
client connections.

Note: MIT App Inventor itself does not make use of these flags, but the
MIT App Inventor source tree is the canonical authoritative location for
the token format definition.

Change-Id: If7cd4ae7cf2fd49718607aeac51d8a56f76df877


ewpatton (approved):
LGTM. I'll file an issue for updating TokenAuth.

PR #2420, opened 2021-02-28
Fix show status bar

jisqyv:
Do not attempt to resize the MockForm when Width and Height are
zero. This happens during initialization when a project is loading, so
just don’t do it.

The bug here is harmless, but results in errors in the browser console
during project load. This fix eliminates those error.

Change-Id: I5953b9e184548c3d004c89dd8573817f77bbd912


jisqyv:
I'm closing this PR because I merged it into #2227 which I am doing my final testing on (and merged in master etc.) I hope to merge #2227 within a few days

PR #2427, opened 2021-03-05
Update CloudDB LUA Script

jisqyv:
The POP_FIRST_SCRIPT updated so that when the list is empty, it is set
to the empty list instead of the empty object (which appears to be the
Redis default when you empty a “table”).

Change-Id: I15002a883eead373b3d23e85867c0b1ab103c0fd


jisqyv:
Fixes #2427

jisqyv:
This should not go into production until the SHA1 hash of the new script is added to the allowed scripts on the clouddb server. Note: The CloudDB server will also recognize the old version and turn it into this new version so we get the benefit of this fix even before we deploy a new Companion.

Updated CloudDB version has been coded and is now being tested.

jisqyv:
The updates to the CloudDB server are in place so this can be tested, merged and put into production.

Note: The kludge to replace the script on the fly on the server is also in place.

ewpatton (approved):
LGTM

PR #2469, opened 2021-04-12
Add hostname to buildserver/vars

jisqyv:
We run 9-15 buildservers simultaneously and have a status page that
reports the /buildserver/vars foreach server. However it is hard to know
which status is for which buildserver. So we add the buildserver
hostname here.

Change-Id: I46fc59626cccf668ab57a5f81581273f0358fd1e


ewpatton (approved):
LGTM

PR #2500, opened 2021-06-21
[GSoC] RadioButton & RadioGroup Component

Hardik5247:
As part of my proposal, I have added a RadioButton component with tests. Next goal is to work on RadioGroup component.

Output of tests:
121785012-f04cc380-cbd4-11eb-8561-5959f2514d82


AppInventorWorkerBee:

Hardik5247:
@ellelili2025 It's ready for review. Code for Component and MockComponent for Horizontal and Vertical RadioGroup and RadioButton is complete along with appropriate test cases.

SusanRatiLane:
I think there are some design issues to be looked at here. The radio button is a user interface component, and the group is a layout, which makes sense at first glance. However, one important feature of a radio button group is that you can easily determine which item is selected in the group, and it doesn't look like the group has a block for that. It looks like the radio buttons are treated like checkboxes, and the only way to determine what is selected in a group is by checking the individual radio buttons.

It might make sense for a radio button group to be more like a listview, where the group itself is the user interface component, and the items are defined inside it.

I also triggered a browser error while working with the blocks.

wxbit:
It is no necessary to add a layout to group RadioButtons, add a grouping attribute is better.

PR #2527, opened 2021-07-30
Refactor: Handle default values cleanly

pavi2410:
- Refactors to handle default values in one place and in a consistent way.
- Changes the default app name to a non-empty string ("AI2 App" for now; any suggestions?) // Read the comment on why it was necessary

Should I handle type conversions too in `Project.java`?


SusanRatiLane (approved):
Looks good.

SusanRatiLane:
This works fine on its own. It should have automated tests, but the attempt to write them has uncovered places where we need the underlying infrastructure to be more robust. 

We're splitting off the automated test effort and merging this on its own.

PR #2552, opened 2021-09-03
Move build tools to its proper dir "/tools/"

pavi2410:
and also prevent leaking these tools into the component classpath


ewpatton (approved):
LGTM

PR #2553, opened 2021-09-03
Support Whitespaces for the Sources' File Paths

MohamedTamer94:
Windows directories, unlike from the UNIX directories, can have whitespaces in its name, these whitespaces aren't recognized correctly in the appinventor build system, which causes builds, which have the sources in a path that contains spaces, to fail.
The failure is mostly caused because of the fact paths are mostly passed as a line arguments when executing a JAR file, the second part of the path ( after the whitespace ), would be considered as a separate argument, I have noticed this behavior exactly with the targets : `Keystore`, `YaClientApp`, and `BlocklyCompile` .
This PR, passes the arguments which could potentially contain a path with spaces using the `arg` tag's `value` attribute, which would preserve paths with spaces as a one argument by quoting it.
I tested this, having  appinventor in the following path: `C:\AppInventor Sources`, and it has compiled correctly.
Please check this the forum [post](https://community.appinventor.mit.edu/t/trying-to-make-own-app-inventor/40254/236?u=mohamed_tamer).


AppInventorWorkerBee:

ewpatton (approved):
LGTM

XomaDev:
👏

PR #2554, opened 2021-09-06
Vagrantfile: Add support for libvirt images.

peanutbutterandcrackers:
Add support for libvirt images in the Vagrantfile. Tested and found to be working.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -22,6 +22,14 @@ Vagrant.configure("2") do |config|
     v.customize ["modifyvm", :id, "--nictype1", "virtio"]
   end
 
+  config.vm.provider :libvirt do |libvirt|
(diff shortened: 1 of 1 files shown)


AppInventorWorkerBee:

peanutbutterandcrackers:
A quick way to test this would be to get a debian-family vm (w/ GUI) running in virtualbox (seems like that is what most of the devs use), and then inside that VM:
`apt install virt-manager vagrant vagrant-libvirt`
...
`cd appinventor-sources`
...
`vagrant up`
and things should work.

If one has both virtualbox and libvirt installed on the host, then `vagrant up --provider=libvirt` should do the trick (once `vagrant-libvirt` is installed)

PR #2557, opened 2021-09-07
Fix built in blocks translation in traditional chinese

ellelili2025:
Those appeared as simplified Chinese.


SusanRatiLane (approved):
LGTM

PR #2600, opened 2021-11-03
Fix ListView handling of filters and recyled elements

SusanRatiLane:
Add handling of elements filtered out and then restored when the filter is changed before they are cycled back into memory. Previously, elements were not restored to visible, resulting in items that disappeared permanently as the filter changed.


ewpatton (approved):
Tested with a list of 10k items on a Google Pixel and it seems to work fine with these changes.

PR #2601, opened 2021-11-04
Fix ListView programmatic selection changes not highlighting

SusanRatiLane:
When selection is changed programmatically with the new ListView, the selection property was set but the component did not highlight the new selection.

This fixes this problem.


jisqyv (approved):
LGTM

PR #2602, opened 2021-11-05
ListView: Fix out of bounds exception

SusanRatiLane:
When the select property is set in blocks to a value not present or against an empty list, an out of bounds exception was thrown. This fixes the problem. 

Now, if the ListView attempts to select an invalid entry, any existing selection will be cleared.


jisqyv:
@SusanRatiLane Unfortunately, if you select the first item, it doesn't get highlighted and the highlight remains on where it was.

jisqyv:
@SusanRatiLane This looks like it works!

jisqyv (approved):
LGTM

PR #2608, opened 2021-11-15
Set ListView SelectionIndex before loading of elements is complete

SusanRatiLane:
Reference: https://community.appinventor.mit.edu/t/listview-selection-error/45502

When the Screen.Initialize event executes, the Cardview objects for each ListView element have not loaded. If the SelectionIndex is set using blocks in this event, the object will be null when the component tries to change the background color of the selected element.

This probably would also happen if the element was not loaded in the Recyclerview.


jisqyv (approved):
LGTM

Issue #2594, opened 2021-10-26
ListView right/center text justification


SusanRatiLane:
**Describe the desired feature**

Requested from the community forum: Ability to right justify text in ListView entries. 

The feature should also include the ability to center and explicitly left-justify.

preetvadaliya:
@SusanRatiLane can I work on this issue?

SusanRatiLane:
> @SusanRatiLane can I work on this issue?

Definitely.

ANUKULJHA1506:
can i work on this issue

preetvadaliya:
@ANUKULJHA1506 thanks for your interest but I have already opened a PR on it please see #2609

prabhasg5:
@SusanRatiLane 
Looking at the existing PR #2609, it added **TextAlignmentMain** and **TextAlignmentDetail** properties but was blocked by two issues:

In the "Main Text, Detail Text (Horizontal)" layout, setting center/right alignment on the main text causes the detail text to disappear, because the main TextView has **WRAP_CONTENT** width and takes all horizontal space.
**Designer preview for the alignment properties was never implemented.**

The codebase has also been substantially refactored since that PR in 2021. ListView now uses dedicated per-layout adapter classes (ListViewSingleTextAdapter, ListViewTwoTextLinearAdapter, etc.), so the old PR won't apply cleanly without conflicts.

My plan is to add TextAlignmentMain and TextAlignmentDetail properties to ListView.java, update each adapter to accept and apply the gravity values, and fix the horizontal layout bug by using weighted LinearLayout distribution so both text views share the row properly when alignment is non-default.

Happy to hear any feedback on this approach before I start!

SusanRatiLane:
> [@SusanRatiLane](https://github.com/SusanRatiLane) Looking at the existing PR [#2609](https://github.com/mit-cml/appinventor-sources/pull/2609), it added **TextAlignmentMain** and **TextAlignmentDetail** properties but was blocked by two issues:
> 
> In the "Main Text, Detail Text (Horizontal)" layout, setting center/right alignment on the main text causes the detail text to disappear, because the main TextView has **WRAP_CONTENT** width and takes all horizontal space. **Designer preview for the alignment properties was never implemented.**
> 
[...]

PR #2609, opened 2021-11-23
ListView Component's Text Alignment Properties

preetvadaliya:
### About PR

1. This PR adds two new designer properties into the ListView component for setting TextAlignment of Main Text and Detail Text.
1. Fixes Issue #2594 

### Note for Future Work

- [ ] Designer Preview for TextAlignemrt Properties.

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/youngandroid/YoungAndroidFormUpgrader.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/youngandroid/YoungAndroidFormUpgrader.java
@@ -1355,6 +1355,10 @@ public final class YoungAndroidFormUpgrader {
       // Added ...
       srcCompVersion = 6;
     }
+    if (srcCompVersion < 7) {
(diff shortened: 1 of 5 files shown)


AppInventorWorkerBee:

SusanRatiLane:

SusanRatiLane (changes requested):
This looks good and tests well for the Companion, but it will need to display in the Designer before we can release it.

SusanRatiLane (changes requested):
This works pretty well overall, but it breaks the Main Text, Detail Text (Horizontal) layout when you set the main text alignment to anything other than default.

SusanRatiLane (changes requested):
@preetvadaliya This overall looks good, but this feature does not work correctly:
center and right-justify for TextAlignment (main text) for layout Main Text, Detail Text (Horizontal).

In the two text horizontal layouts, the detail text does not display at all when the main text is set to center or right-justify.

Issue #2597, opened 2021-10-27
ListView separator line between items.


SusanRatiLane:
**Describe the bug**

See community thread: https://community.appinventor.mit.edu/t/listview-items-missing-visible-line-separator-between-items/44384

The old ListView had separator lines between elements at least some of the time. The current ListView does not. Separator lines between elements would be cleaner than what we currently have.

preetvadaliya:
We can use [DividerItemDecoration](https://developer.android.com/reference/androidx/recyclerview/widget/DividerItemDecoration) to set item separator line.

SusanRatiLane:
> We can use [DividerItemDecoration](https://developer.android.com/reference/androidx/recyclerview/widget/DividerItemDecoration) to set item separator line.

This seems like a good approach.

prabhasg5:
.

prabhasg5:
Hi @SusanRatiLane, I looked at the current state of this on both Android and iOS.

Android: **DividerThickness** and **DividerColor** properties already exist, but the default thickness is 0 so no separator shows unless the user explicitly sets it. The underlying drawing logic works fine.
 and with deeper examination there is also a small bug in **setDivider()** where **removeLayoutChangeListener()** is called on the newly created decoration object instead of the old one, which **can cause a listener leak** a OnLayoutChangeListener when horizontal orientation + margins are used.

iOS: **UITableView** shows separators by default (UIKit behavior), but there's no code to hide them when **DividerThickness** is 0 — so the separator is always visible on iOS regardless of what the user sets.

Android hides separators by default, iOS always shows them.

Should I make a PR to align the behaviour, or is the current state acceptable to close this?

SusanRatiLane:
@prabhasg5 It sounds like current state is fine for Android. A PR to fix the bug you discovered would be nice.

Do you have an idea of how much work it might take to bring iOS in line with Android? I know that @cindyloo has also done work on this on the iOS side. I want to be sure PRs for her work have been reviewed and merged before any more work is done on that.

prabhasg5:
@SusanRatiLane  Thanks for the clarification. From my initial look, bringing iOS behavior in line with Android for this issue seems relatively small, since it would mainly involve conditionally hiding/showing **UITableView** separators based on **DividerThickness**. I’ll wait for the existing iOS PRs to be reviewed and merged first to avoid overlap.

For now, I’ll focus on opening a PR for the Android listener leak bug.

[...]

PR #2610, opened 2021-11-24
ListView component Line Separator properties 

preetvadaliya:
### About PR

1. This PR adds support of **Line Separator** properties (HasSeparator, SeparatorColor, SeparatorSize) into the ListView component.
2. Fixes issue #2597 

### HasSeparator
Description:- Sets whether the ListView has a line separator or not.
Editor Type:- Boolean (Default value **false**)
Property Type:- Designer Only

### SeparatorColor
Description:- Sets the color of the ListView's line separator if ListView has a line separator.
Editor Type:- Color (Default value **pink**)
Property Type:- Designer Only

### SeparatorSize
Description:- Sets the size of the ListView's line separator it would be a height of the line separator in the case of vertical orientation and width of the line separator in the case of horizontal orientation.
Editor Type:- Non-Negative Integer (Default value **1DP**)
Property Type:- Designer Only

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/ListView.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/ListView.java
@@ -6,12 +6,14 @@
 
 package com.google.appinventor.components.runtime;
 
+import android.graphics.drawable.GradientDrawable;
 import android.text.Editable;
 import android.text.TextWatcher;
 import android.util.Log;
 import android.view.View;
 import android.widget.AdapterView;
 import android.widget.EditText;
+import androidx.recyclerview.widget.DividerItemDecoration;
 import androidx.recyclerview.widget.RecyclerView;
 import androidx.recyclerview.widget.LinearLayoutManager;
 import androidx.recyclerview.widget.RecyclerView.LayoutParams;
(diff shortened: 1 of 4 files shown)


AppInventorWorkerBee:

appinventor/components/src/com/google/appinventor/components/runtime/ListView.java
```diff
@@ -922,6 +930,91 @@ public void ListData(String propertyValue) {
...
+   *
+   * @return the int value of line separator size
+   */
+  @SimpleProperty(userVisible = false)
+  public int SeparatorSize()  {
+    return this.separatorSize;
+  }
+
+  private void setSeparator() {
+    if(hasSeparator)  {
```
SusanRatiLane, line 1002: The HasSeparator property does not appear to work. I see a separator in my test app whether it is checked or unchecked in the Designer. It looks like your if statement in SetSeparator needs an else clause to remove the separator.

appinventor/components/src/com/google/appinventor/components/runtime/ListView.java
```diff
@@ -922,6 +930,91 @@ public void ListData(String propertyValue) {
...
+      GradientDrawable gradientDrawable = new GradientDrawable(GradientDrawable.Orientation.BOTTOM_TOP, new int[]{separatorColor, separatorColor});
+      DividerItemDecoration dividerItemDecoration;
+      if(orientation == ComponentConstants.LAYOUT_ORIENTATION_VERTICAL) {
```
SusanRatiLane, line 1005: When Orientation is set to horizontal, I see a vertical separator, which seems correct. However, I also see a horizontal separator below the listview element that is set to default size and color.

SusanRatiLane (changes requested):
When I use the filter bar, separators appear for elements that have been filtered out. Filtering sets the visibility property of the elements to false if they don't match the filter. This should also happen for the separator.

You should look at ListAdapterWithRecyclerView.java. That's where the CardView is defined. I think it might make more sense to make the separator part of the CardView. That should make separators disappear when the rest of the element is filtered out.

PR #2617, opened 2021-12-10
Cache Asset Database 

jingmiao-z:
Previously, we need to upload all the assets again every time we reconnect to the AI companion. 
The logic we implement in this pull request will check/compare in the "assets" folder on the device.
If items are already there, we will only upload/change differences. If there is no change, no need to download again.


AppInventorWorkerBee:

jisqyv:
This PR is superceded by #2664 which contains the changes here as well as additional work

PR #2621, opened 2021-12-17
"is string" block tooltip fixed.

preetvadaliya:
This PR remove not useful `<code></code>` tag from built-in block "is string" tool-tip.


AppInventorWorkerBee:

SusanRatiLane (approved):

PR #2629, opened 2022-01-12
Fix problems that have not been translated into tooltips.

### LINKED ISSUE #2624: Problems in which descriptions of component blocks are not translated.
**Describe the bug**

image
As shown in the picture above, the description of the component's events, procedures, and property blocks is displayed in English regardless of the set language.
This appears only in blocks of components(buttons, etc.) and screen, and does not appear in built-in blocks such as control, logic, etc.

**Affects**

- [ ] Designer
- [x] Blocks editor
- [ ] Companion
- [ ] Compiled apps
- [ ] Buildserver
- [ ] Debugging

**Expected behavior**

It can be difficult for users unfamiliar with English to understand the blocks.

**Steps to reproduce**

You can just access the appinventer and check it right away.

I did an experiment to see if this problem actually does not apply to translation files (OdeMessages_xx.properties).
image
For example, I changed the description of the Timer event to "Description in Clock.java" in file Clock.java, the source code of the clock component, and built it to check.
image
Then, as in the picture above, the revised description regardless of language was displayed as it was.

I don't think the translation file is applied normally. But I don't know where the problem is.

semicolon3253:
Fixes #2624 
Fixed a problem in which tooltips were not translated in the component block.

### DIFF
--- a/appinventor/blocklyeditor/src/component_database.js
+++ b/appinventor/blocklyeditor/src/component_database.js
@@ -786,7 +786,7 @@ Blockly.ComponentDatabase.prototype.getInternationalizedEventName = function(nam
  * @returns {string} The localized string if available, otherwise the unlocalized name.
  */
 Blockly.ComponentDatabase.prototype.getInternationalizedEventDescription = function(component, name, opt_default) {
-  return this.i18nEventDescriptions_[component + '.' + name] || this.i18nEventDescriptions_[name] || opt_default || name;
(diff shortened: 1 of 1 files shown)


AppInventorWorkerBee:

devarsh-mavani-19 (approved):
nice

SusanRatiLane (approved):
This works, though I'm pretty sure there are more problems to be solved in component tooltip translation.

PR #2630, opened 2022-01-20
Add some detail to logging from ProjectServiceImpl

jisqyv:
Add the name of the template attempting to be read to the log when an
I/O Exception occurs. The I/O Exception in question would only happen
if there was an internal error inside of App Engine. And there was, so
this code was added, so we could determine which template json file
was corrupted.

This change also includes adding an innocuous space to
TranslateApp.json, which was the corrupted file. The additional space
forced App Engine to replace it when we uploaded a new version, thus
mitigating the corruption.

It isn't clear how long the corrupted version will live inside of App
Engine or how long it should be before we remove the extra space.

BTW. This corruption appears to only be on the hardware hosting
ai2. So other versions, such as ai2-test were unaffected.

Change-Id: I1106c2abf9cfa6b06b4b57adebebc2ca9fcec0b5


ewpatton (approved):
LGTM. Do you want to rebase on master to keep the signature intact?

PR #2632, opened 2022-01-24
Updated Wrong Documentation issue#2631

devarsh-mavani-19:
There were small problems in documentation
1. There was wrong image of Square root Block 
2. There was wrong image of Absolute block
So I replaced it with new block

See issue no. 2631


AppInventorWorkerBee:

PR #2634, opened 2022-01-24
Bug in Documentation Fixed Issue#2631

devarsh-mavani-19:
There was this 2 mistakes in documentation which I have fixed

1. Wrong image for Square root block
2. Wrong image for Absolute block
3. Square root being referred as Sqrt in 2 places


AppInventorWorkerBee:

devarsh-mavani-19:
@ewpatton I made my changes to a separate branch as per your suggestion thank you.

preetvadaliya:
@ewpatton it looks good, I checked it.
1. all block images are changed from **sqrt** to **square root**.
2. no broken links.
3. gifs are also updated to **square root**.

preetvadaliya:
If we don't consider the height and width of the block images then it looks good otherwise the size of the new images is larger than the previous one.

@devarsh-mavani-19 I think you had downloaded block as png from block editor then used in docs while other block images are just screenshots.

PR #2638, opened 2022-02-09
Clear unused blocks Issue#2582

devarsh-mavani-19:
I have added a function to remove unused blocks.
a confirmation dialog will be shown and when user press okay unused blocks will be deleted

see this issue for more idea
https://github.com/mit-cml/appinventor-sources/issues/2582

### DIFF
--- a/appinventor/blocklyeditor/src/workspace_svg.js
+++ b/appinventor/blocklyeditor/src/workspace_svg.js
@@ -1012,6 +1012,28 @@ Blockly.WorkspaceSvg.prototype.customContextMenu = function(menuOptions) {
   };
   menuOptions.push(enableAll);
 
+  // Clear unused blocks
+  var clearUnusedBlocks = {enabled: true};
+  clearUnusedBlocks.text = Blockly.Msg.REMOVE_UNUSED_BLOCKS;
+  clearUnusedBlocks.callback = function() {
+    var allBlocks = Blockly.mainWorkspace.getAllBlocks();
+    for (var x = 0, block; block = allBlocks[x]; x++) {
+      // block.setDisabled(false);
+      if(block.category == 'Procedures' && block.type.search('procedures_def') != -1) {
+        var name = block.getProcedureDef()[0];
+        //if this procedure is not called by any other block then remove it
+        if(Blockly.Procedures.getCallers(name, Blockly.mainWorkspace) == 0 && block.getChildren().length == 0 && block.getParent() == null) {
+          block.dispose(false);
+        }
+      } else {
+        if(block.getChildren().length == 0 && block.getParent() == null) {
+          block.dispose(false);
+        }
+      }
+    }
+  };
+  menuOptions.push(clearUnusedBlocks);
+
   // Disable all blocks
   var disableAll = {enabled: true};
   disableAll.text = Blockly.Msg.DISABLE_ALL_BLOCKS;
--- a/appinventor/blocklyeditor/src/workspace_svg.js
+++ b/appinventor/blocklyeditor/src/workspace_svg.js
@@ -1016,6 +1016,8 @@ Blockly.WorkspaceSvg.prototype.customContextMenu = function(menuOptions) {
   var clearUnusedBlocks = {enabled: true};
   clearUnusedBlocks.text = Blockly.Msg.REMOVE_UNUSED_BLOCKS;
   clearUnusedBlocks.callback = function() {
+    var response = confirm(Blockly.Msg.CONFIRM_DELETE)
+    if(!response) return;
     var allBlocks = Blockly.mainWorkspace.getAllBlocks();
     for (var x = 0, block; block = allBlocks[x]; x++) {
       // block.setDisabled(false);
--- a/appinventor/blocklyeditor/src/msg/ai_blockly/messages.json
+++ b/appinventor/blocklyeditor/src/msg/ai_blockly/messages.json
@@ -31,6 +31,7 @@
 	"Blockly.Msg.EXPAND_ALL": "Expand Blocks",
 	"Blockly.Msg.ARRANGE_H": "Arrange Blocks Horizontally",
 	"Blockly.Msg.ARRANGE_V": "Arrange Blocks Vertically",
+	"Blockly.Msg.REMOVE_UNUSED_BLOCKS": "Remove Unused Blocks",
 	"Blockly.Msg.ARRANGE_S": "Arrange Blocks Diagonally",
 	"Blockly.Msg.SORT_W": "Sort Blocks by Width",
 	"Blockly.Msg.SORT_H": "Sort Blocks by Height",


AppInventorWorkerBee:

SusanRatiLane:

preetvadaliya:
@devarsh-mavani-19 you need to change the base branch from `ucr` to `master`.

appinventor/blocklyeditor/src/workspace_svg.js
```diff
@@ -1012,6 +1012,30 @@ Blockly.WorkspaceSvg.prototype.customContextMenu = function(menuOptions) {
   };
   menuOptions.push(enableAll);
 
+  // Clear unused blocks
+  var clearUnusedBlocks = {enabled: true};
+  clearUnusedBlocks.text = Blockly.Msg.REMOVE_UNUSED_BLOCKS;
+  clearUnusedBlocks.callback = function() {
+    var response = confirm(Blockly.Msg.CONFIRM_DELETE)
```
ewpatton, line 1019: I'm not sure that this is the logic we want here.I think what we should do is compute the list of non-declaration top level blocks and present the dialog with "Are you sure you want to remove N blocks?" so the user is aware how many blocks will be trashed first. The deletion operations should be wrapped in an event group so if the user presses ctrl+z they will get them all back.

Note that the only valid declaration blocks are global declarations, procedure definitions, and event handler blocks

ewpatton (changes requested):

appinventor/blocklyeditor/src/workspace_svg.js
devarsh-mavani-19: @ewpatton sir just to summarize  we have ro remove non declaration blocks (for example: if-else block) and top level blocks (those blocks that does not have any parent block and no child blocks ) right?

devarsh-mavani-19: @ewpatton sir I have made changes as you said. here is what i have done 
`if(block.previousConnection != null || (block.outputConnection && block.outputConnection != null)) {
          removeList.push(block)
      }`
this makes sure that the block is of non-declarative type. after that i have opened a dialog with a message 
"Are you sure you want to delete all X of these blocks?" and if user clicks ok it will delete those.

In addition to that i have also wrapped as event group so that undo functionality works

devarsh-mavani-19: @ewpatton sir i have made changes as you said can you please review it so that i can work on other issues?

devarsh-mavani-19: @ewpatton Can you please check the latest commit I have made? I did changes as per your inputs

appinventor/blocklyeditor/src/workspace_svg.js
```diff
@@ -1012,6 +1012,38 @@ Blockly.WorkspaceSvg.prototype.customContextMenu = function(menuOptions) {
   };
   menuOptions.push(enableAll);
 
+  // Clear unused blocks
+  var clearUnusedBlocks = {enabled: true};
+  clearUnusedBlocks.text = Blockly.Msg.REMOVE_UNUSED_BLOCKS;
+  clearUnusedBlocks.callback = function() {
+    Blockly.Events.setGroup(true);
```
ewpatton, line 1019: The start of the event group should only happen when the user confirms the deletion on line 1036. You'll also want to wrap the changes in a try-finally block, e.g.:

```
try {
  Blockly.Events.setGroup(true);
  // manipulate the workspace
} finally {
  // clean up the group regardless of whether an error occurs
  Blockly.Events.setGroup(false);
}
```

This prevents the event group from being left open if for some reason deleting a block causes an exception.

appinventor/blocklyeditor/src/workspace_svg.js
```diff
@@ -1012,6 +1012,38 @@ Blockly.WorkspaceSvg.prototype.customContextMenu = function(menuOptions) {
...
+    var allBlocks = Blockly.getMainWorkspace().getTopBlocks()
+    var removeList = []
+    for (var x = 0, block; block = allBlocks[x]; x++) {
+      if(block.previousConnection != null || (block.outputConnection && block.outputConnection != null)) {
+          removeList.push(block)
```
ewpatton, line 1024: Too much indentation here.

appinventor/blocklyeditor/src/workspace_svg.js
```diff
@@ -1012,6 +1012,38 @@ Blockly.WorkspaceSvg.prototype.customContextMenu = function(menuOptions) {
...
+      }
+    }
+    if(removeList.length == 0) {
+      alert('No Orphaned Blocks');
```
ewpatton, line 1028: I'm not sure the alert is necessary.

appinventor/blocklyeditor/src/workspace_svg.js
```diff
@@ -1012,6 +1012,38 @@ Blockly.WorkspaceSvg.prototype.customContextMenu = function(menuOptions) {
...
 
+  // Clear unused blocks
+  var clearUnusedBlocks = {enabled: true};
+  clearUnusedBlocks.text = Blockly.Msg.REMOVE_UNUSED_BLOCKS;
+  clearUnusedBlocks.callback = function() {
+    Blockly.Events.setGroup(true);
+    var allBlocks = Blockly.getMainWorkspace().getTopBlocks()
+    var removeList = []
+    for (var x = 0, block; block = allBlocks[x]; x++) {
+      if(block.previousConnection != null || (block.outputConnection && block.outputConnection != null)) {
```
ewpatton, line 1023: It seems like `if (block.previousConnection || block.outputConnection)` would be a sufficient check here because these ought to only be set for blocks that have one of those (which the valid top-level blocks do not).

appinventor/blocklyeditor/src/workspace_svg.js
```diff
@@ -1012,6 +1012,38 @@ Blockly.WorkspaceSvg.prototype.customContextMenu = function(menuOptions) {
...
+      return;
+    }
+    var msg = Blockly.Msg.WARNING_DELETE_X_BLOCKS.replace('%1', String(removeList.length));
+    var cancelButton = top.BlocklyPanel_getOdeMessage('cancelButton');
+    var deleteButton = top.BlocklyPanel_getOdeMessage('deleteButton');
+    var dialog = new Blockly.Util.Dialog(Blockly.Msg.CONFIRM_DELETE, msg, deleteButton, true, cancelButton, 0, function(button) {
+        dialog.hide();
+        if (button == deleteButton) {
+          Blockly.mainWorkspace.playAudio('delete');
+          for(var x = 0;x < removeList.length;x++) {
```
ewpatton, line 1038: Code style: spacing after `for` keyword and semicolons.

ewpatton (changes requested):
@devarsh-mavani-19 A few small things but they should be relatively quick to address.

devarsh-mavani-19:
@ewpatton fixed the above mentioned changes as well. Can you please check the latest commit I have made? Thanks

appinventor/blocklyeditor/src/workspace_svg.js
```diff
@@ -1012,6 +1012,41 @@ Blockly.WorkspaceSvg.prototype.customContextMenu = function(menuOptions) {
...
+    var dialog = new Blockly.Util.Dialog(Blockly.Msg.CONFIRM_DELETE, msg, deleteButton, true, cancelButton, 0, function(button) {
+      dialog.hide();
+      if (button == deleteButton) {
+        try {
+          Blockly.Events.setGroup(true);
+          Blockly.mainWorkspace.playAudio('delete');
+          for (var x = 0; x < removeList.length; x++) {
+            removeList[x].dispose(false);
+          }
+        } catch (e) {
```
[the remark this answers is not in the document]
devarsh-mavani-19: ok so i should throw exception right? @ewpatton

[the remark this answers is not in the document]
ewpatton: You can just omit the catch clause. It's not necessary.

devarsh-mavani-19:
@ewpatton  removed catch block. Please check it

ewpatton:
@devarsh-mavani-19 The behavior looks good, but I question placing the "Remove all orphan blocks" between the Enable all blocks and Disable all blocks options since it might make it easy to accidentally hit one of those two. It might be better to place it below the Clean up blocks item to group it with another "cleaning" function.

devarsh-mavani-19:
@ewpatton  I have placed the item along with clean block item. Can you please check the latest commit I have made? Thanks

PR #2643, opened 2022-02-21
Fix reading images out of CloudDB

jisqyv:
Broken as of nb188. The problem is that Form.getDefaultPath returns a
URI, not a file path so cannot be simply handed to "File". Instead we
have to parse the URI and extract the path component.

Change-Id: I2f7659219bba144ce5f89a26aec54f39ddfd4ca0


jisqyv:
Fixes #2642

jisqyv:
There are some other bugs associated with this that I want to fix and include here... thus a draft

ewpatton (approved):

PR #2644, opened 2022-02-23
change language name to code (yandex translate) 

daki7711:
Capable of entering the Language/Country name and will automatically convert it to the corresponding language code. next steps is to improve the interface of the block

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/YandexTranslate.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/YandexTranslate.java
@@ -28,6 +28,10 @@ import com.google.appinventor.components.runtime.util.ErrorMessages;
 import java.io.IOException;
 import java.io.InputStreamReader;
 
+import java.util.Dictionary;
+import java.util.Enumeration;
+import java.util.Hashtable;
+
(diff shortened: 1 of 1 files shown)


AppInventorWorkerBee:

halatmit:
Reminder to be careful about using "country" as opposed to "country/region"

== Hal

Hal Abelson
***@***.***
Prof. of Comp. Sci. and Eng.

daki7711:
> Reminder to be careful about using "country" as opposed to "country/region" == Hal Hal Abelson ***@***.*** Prof. of Comp. Sci. and Eng. MIT Dept. of Elec. Eng. and Comp. Sci.

Got it just changed to language name as of now

halatmit:
Good.  Thanks.

== Hal

Hal Abelson
***@***.***
Prof. of Comp. Sci. and Eng.

jisqyv:
@daki7711 I'm putting this change on hold as we are deprecating YandexTranslate in favor of the new Translate Component (see #2656). This new component uses a translation proxy run at MIT, which currently uses Amazon's translation service. I'm thinking that this mapping can be done in the proxy, so updating it would not require a component release.

PR #2661, opened 2022-03-18
Typeblocking "1" should prioritize number 1

devarsh-mavani-19:
Entering 1 in the typeblocking window includes the number 1 as the last entry rather than the first. The results should probably be sorted most-least specific but it seems to be opposite. It's much more likely for a program to contain the number 1 rather than the various base conversion blocks that appear earlier in the list.

see issue #2660


AppInventorWorkerBee:

devarsh-mavani-19:
> @AppInventorWorkerBee ok to test

I found another bug while fixing this one.

when we typeblock the word "text" in any combination of lower & uppercase letter except "text" and "Text", we are not able to see the text block on top.Probably the expected behaviour is: the text block should be on top of list (ex: on top of button.text) no matter casing of characters. here is a screenshot showing that

Screenshot from 2022-03-18 23-23-03

If you can confirm this is not an expected behaviour, I can fix that as well in same PR as it's just a small change.

ewpatton:
@devarsh-mavani-19 I will need to think about that. Usually the way I do text is to do an open double quote, e.g. `"` and then start typing the text. The thing I usually trip up on after this is putting the close quote, which then ends up in the text. We may want to detect if the user has typed the closing " and ignore it when generating the block if this is the case.

devarsh-mavani-19:
> @devarsh-mavani-19 I will need to think about that. Usually the way I do text is to do an open double quote, e.g. `"` and then start typing the text. The thing I usually trip up on after this is putting the close quote, which then ends up in the text. We may want to detect if the user has typed the closing " and ignore it when generating the block if this is the case.

so I can do something like this?
```
if more than 1 " found 
then
        ignore the first and last " and add rest as text
```

devarsh-mavani-19:
> @devarsh-mavani-19 I will need to think about that. Usually the way I do text is to do an open double quote, e.g. `"` and then start typing the text. The thing I usually trip up on after this is putting the close quote, which then ends up in the text. We may want to detect if the user has typed the closing " and ignore it when generating the block if this is the case.


latest commit should solve it. can you please check this video
https://user-images.githubusercontent.com/55136047/159063509-15625b57-2b44-41da-9c38-7bfce24f4a82.mp4

ewpatton:

appinventor/blocklyeditor/src/typeblock.js
```diff
@@ -547,7 +547,7 @@ Blockly.TypeBlock.prototype.createAutoComplete_ = function(inputText){
             canonicName: 'text',
             dropDown: {
               titleName: 'TEXT',
-              value: blockName.substring(1)
+              value: blockName.endsWith('"') ? blockName.substring(1, blockName.length - 1) : blockName.substring(1)
```
ewpatton, line 550: Since this would require us to bump the compatibility matrix, we may want to use `goog.string.endsWith` instead.

devarsh-mavani-19: @ewpatton sorry for the long delay. I changed that can you please review it? Thanks

ewpatton (approved):
LGTM

PR #2668, opened 2022-04-05
Enable non-default event parameter names in imported and exported .aia files

mark-friedman:
This PR allow you to optionally specify the names of event parameters in the XML of imported (and exported) aia files.  This helps support Aptly.  Please see [this Aptly issue](https://github.com/mit-cml/aptly/issues/80) for further insight.

I've also attached an aia file (renamed as a zip file, so that GitHub will allow me to uploaded it) that you can use to help test this:
[EventParams.zip](https://github.com/mit-cml/appinventor-sources/files/8422242/EventParams.zip)

### DIFF
--- a/appinventor/blocklyeditor/src/blocks/components.js
+++ b/appinventor/blocklyeditor/src/blocks/components.js
@@ -602,6 +602,20 @@ Blockly.Blocks.component_event = {
       if (varList.length != params.length) {
         return false; // parameters have changed
       }
+      if ("true" === componentType.external) {
+        for (var x = 0; x < varList.length; ++x) {
+          var found = false;
+          for (var i = 0, param; param = params[i]; ++i) {
+            if (componentDb.getInternationalizedParameterName(param.name) == varList[x]) {
+              found = true;
+              break;
+            }
+          }
+          if (!found)  {
+            return false; // parameter name changed
+          }
+        }
+      }
       // No need to check event return type, events do not return.
       return true; // passed all our tests! block is defined!
     };
--- a/appinventor/blocklyeditor/src/blocks/components.js
+++ b/appinventor/blocklyeditor/src/blocks/components.js
@@ -240,10 +240,11 @@ Blockly.Blocks.component_event = {
                                                              // The absence of this attribute means horizontal.
     }
 
+    // Note that this.parameterNames only contains parameter names that have
+    // overridden the default event parameter names specified in the component
+    // DB
     for (var i = 0; i < this.parameterNames.length; i++) {
-      var parameter = document.createElement('param');
-      parameter.setAttribute('name', this.parameterNames[i]);
-      container.appendChild(parameter);
+      container.setAttribute('param_name' + i, this.parameterNames[i]);
     }
 
     return container;
@@ -276,14 +277,18 @@ Blockly.Blocks.component_event = {
       delete this.instanceName;
     }
 
+    // this.parameterNames will be set to a list of names that will override the
+    // default names specified in the component DB. Note that some parameter
+    // names may be overridden while others may remain their defaults
     this.parameterNames = [];
-    var paramElements = xmlElement.getElementsByTagName('param');
-    if (paramElements) {
-      for (var i = 0; i < paramElements.length; i++) {
-        var paramElement = paramElements[i];
-        var paramName = paramElement.getAttribute('name');
-        this.parameterNames.push(paramName);
-      }
+    var numParams = this.getDefaultParameters_().length
+    for (var i = 0; i < numParams; i++) {
+      var paramName = xmlElement.getAttribute('param_name' + i);
+      // For now, we only allow explicit parameter names starting at the beginning
+      // of the parameter list.  Some day we may allow an arbitrary subset of the
+      // event params to be explicitly specified.
+      if (!paramName) break;
+      this.parameterNames.push(paramName);
     }
 
     // Orient parameters horizontally by default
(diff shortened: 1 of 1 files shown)


AppInventorWorkerBee:

appinventor/blocklyeditor/src/blocks/components.js
```diff
@@ -572,18 +602,6 @@ Blockly.Blocks.component_event = {
       if (varList.length != params.length) {
         return false; // parameters have changed
       }
-      for (var x = 0; x < varList.length; ++x) {
-        var found = false;
-        for (var i = 0, param; param = params[i]; ++i) {
-          if (componentDb.getInternationalizedParameterName(param.name) == varList[x]) {
```
ewpatton, line 578: It seems like this causes the internationalization of parameter names to be dropped.

ewpatton:

appinventor/blocklyeditor/src/blocks/components.js
mark-friedman: I thought that it was just verifying that all of the parameter names matched the defaults, but maybe I misunderstood.  I don't see how that could happen in ordinary App Inventor programs, so it seemed safe to drop it, especially since it interferes with the purpose of this PR.

ewpatton: Based on my review of the logs, I think it was to do with extension upgrading, i.e., if the parameter names changed then the block was likely no longer valid (or at least should be inspected). Prior to extensions, this code was not here.

mark-friedman: So, do you think it's worth keeping that check if it prevents Aptly users (including Codex) from using the non-default event parameter names?  Or maybe you know of a way to only do this check for extensions.

halatmit:
Shouldn't Aptly users be able to use non-default parameter names?
Is there a reason why not?

==Hal

mark-friedman:
> Shouldn't Aptly users be able to use non-default parameter names? Is there a reason why not? ==Hal

Yes, I think they should.  That's what this PR enables.  However there is some existing code (that I removed in this PR) that does a verification check that would prevent Aptly users being able to use non-default parameter names.  The thread with Evan is about why that check was there, how important is it, and whether there might be a way to alter the check so that it works for it's original intent but doesn't prevent the Aptly use case (or a future use case where we allowed App Inventor users to rename event parameters).

halatmit:
I'm wondering why the check was there in the first place.  Is there a use
case for restricting the parameter names?

==Hal

mark-friedman:
> I'm wondering why the check was there in the first place. Is there a use case for restricting the parameter names? ==Hal

Apparently, it was there to ensure that when extensions were upgraded that they kept the same event parameters.  I've adjusted the code to only do the check for extensions.

appinventor/blocklyeditor/src/blocks/components.js
mark-friedman: As we discussed, I've adjusted the code to only do the check for extensions.

mark-friedman:
@ewpatton PTAL

mark-friedman:
I'm going to change the code from using XML child elements of `mutation` for the params to using attributes of the `mutation` element.  This will make it possible for aiatools to process it and is also more Blockly-y.

mark-friedman:
@ewpatton 
I think this is ready again.  Attached is a new aia (disguised as a zip) that can be used to test.  The previous one will not work.

[EventParamsAsAttrs.zip](https://github.com/mit-cml/appinventor-sources/files/8433778/EventParamsAsAttrs.zip)

halatmit:
I'm confused about how this is supposed to work.  If I go to the event when
block, I should be able to click on the
parameter name and edit it -- similar to changing the local variable  name
in the "initialize local" block.
But I don't see that ability to change.

==Hal

mark-friedman:
@halatmit 

This PR does **not** change the current App Inventor UI to allow renaming the event parameters. It just allows for blocks with non-default parameter names to be imported and exported.  This fixes the problem described in [this Aptly issue](https://github.com/mit-cml/aptly/issues/80#issuecomment-1086962994). 

I've changed the title of this PR to hopelully make this clearer.

halatmit:
That makes sense.  But how does one get a block with a non-default event
parameter name?
==Hal

mark-friedman:
> That makes sense. But how does one get a block with a non-default event parameter name?

Aptly

halatmit:
Ah!

thanks.

I'd still like to  change App Inventor so non-default parameters are OK.
But maybe it doesn't matter for
beginning users.

==Hal

mark-friedman:
> I'd still like to change App Inventor so non-default parameters are OK.

That's perfectly reasonable, just not what this PR is for :-)

appinventor/blocklyeditor/src/blocks/components.js
```diff
@@ -405,7 +427,17 @@ Blockly.Blocks.component_event = {
...
-    /** @type {EventDescriptor} */
+   /** @type {EventDescriptor} */
+   var defaultParameters = this.getDefaultParameters_();
+   var explicitParameterNames = this.getExplicitParameterNames_();
+   var params = [];
+   for (var i = 0; i < defaultParameters.length; i++) {
+     var paramName = explicitParameterNames[i] || defaultParameters[i].name;
+     params.push({name: paramName, type: defaultParameters[i].type});
+   }
+   return params;
```
ewpatton, line 438: Indentation looks like it's only +1 here rather than +2.

ewpatton (changes requested):
One minor code style issue to address but otherwise LGTM. I was actually surprised to see that this not only handles the blocks but that the YAIL generator seems to do the right thing as well, which is nice.

appinventor/blocklyeditor/src/blocks/components.js
mark-friedman: Fixed.

mark-friedman:
PTAL

ewpatton (approved):
LGTM

PR #2690, opened 2022-06-05
Add Introduction Message on the home screen of Companion

jisqyv:
Change-Id: I1fdfcbaac6b88d7db36bd9b3af9499e8b37191f3


ewpatton (approved):
@jisqyv I'm approving this, but we may want to think about the text. In the iOS version, we previously had the following text centered on the screen before @SusanRatiLane implemented the onboarding slides:

```
New to MIT App Inventor?
Try our Hour of Code tutorial by opening
http://bit.ly/MITAppInventorHoC
on your computer
```

PR #2698, opened 2022-06-15
Fix "Doit" in Legacy Mode

jisqyv:
Also fix Legacy Connection when the first App opened does not have any assets.

Change-Id: I47df823f8a67b95fe3338277e31cf47564f58968


jisqyv:
Fixes #2695 and #2697

SusanRatiLane (approved):
LGTM

PR #2708, opened 2022-07-28
Sdk31

jisqyv:
First crack at SDK 31 (Android 12)

### DIFF
--- a/appinventor/buildserver/src/com/google/appinventor/buildserver/Compiler.java
+++ b/appinventor/buildserver/src/com/google/appinventor/buildserver/Compiler.java
@@ -1217,6 +1217,8 @@ public final class Compiler {
         } else if (isMain && isForCompanion) {
           out.write("android:launchMode=\"singleTop\" ");
         }
+        // The line below is required for Android 12+
+        out.write("android:exported=\"true\" ");
 
         out.write("android:windowSoftInputMode=\"stateHidden\" ");
 
@@ -1273,7 +1275,7 @@ public final class Compiler {
 
         // Companion display a splash screen... define it's activity here
         if (isMain && isForCompanion) {
-          out.write("    <activity android:name=\"com.google.appinventor.components.runtime.SplashActivity\" android:screenOrientation=\"behind\" android:configChanges=\"keyboardHidden|orientation\">\n");
+          out.write("    <activity android:name=\"com.google.appinventor.components.runtime.SplashActivity\" android:exported=\"false\" android:screenOrientation=\"behind\" android:configChanges=\"keyboardHidden|orientation\">\n");
           out.write("      <intent-filter>\n");
           out.write("        <action android:name=\"android.intent.action.MAIN\" />\n");
           out.write("      </intent-filter>\n");
@@ -1290,8 +1292,8 @@ public final class Compiler {
       subelements.addAll(contentProvidersNeeded.entrySet());
 
 
-      // If any component needs to register additional activities, 
-      // broadcast receivers, services or content providers, insert 
+      // If any component needs to register additional activities,
+      // broadcast receivers, services or content providers, insert
       // them into the manifest here.
       if (!subelements.isEmpty()) {
         for (Map.Entry<String, Set<String>> componentSubElSetPair : subelements) {
@@ -1342,7 +1344,7 @@ public final class Compiler {
           }
         }
         out.write(
-            "<receiver android:name=\"" + brNameAndActions[0] + "\" >\n");
+            "<receiver android:name=\"" + brNameAndActions[0] + "\" android:exported=\"true\">\n");
         if (brNameAndActions.length > 1) {
           out.write("  <intent-filter>\n");
           for (int i = 1; i < brNameAndActions.length; i++) {
--- a/appinventor/components/build.xml
+++ b/appinventor/components/build.xml
@@ -12,7 +12,7 @@
     App Inventor component libraries
   </description>
 
-  <property name="android.sdk.version" value="30"/>
+  <property name="android.sdk.version" value="31"/>
 
   <target name="all"
           depends="CommonConstants,AndroidRuntime">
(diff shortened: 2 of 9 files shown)


jisqyv:
There is a known problem with Android 12. I should have a fix later tonight or tomrrow

appinventor/buildserver/src/com/google/appinventor/buildserver/Compiler.java
```diff
@@ -1273,7 +1275,7 @@ private boolean writeAndroidManifest(File manifestFile) {
 
         // Companion display a splash screen... define it's activity here
         if (isMain && isForCompanion) {
-          out.write("    <activity android:name=\"com.google.appinventor.components.runtime.SplashActivity\" android:screenOrientation=\"behind\" android:configChanges=\"keyboardHidden|orientation\">\n");
+          out.write("    <activity android:name=\"com.google.appinventor.components.runtime.SplashActivity\" android:exported=\"true\" android:screenOrientation=\"behind\" android:configChanges=\"keyboardHidden|orientation\">\n");
```
ewpatton, line 1278: I wonder if export="true" make sense here. I think we would only want the companion app itself to launch this activity, in which case false would be more approprirate.

appinventor/buildserver/src/com/google/appinventor/buildserver/Compiler.java
```diff
@@ -1217,6 +1217,8 @@ private boolean writeAndroidManifest(File manifestFile) {
         } else if (isMain && isForCompanion) {
           out.write("android:launchMode=\"singleTop\" ");
         }
+        // The line below is required for Android 12+
+        out.write("android:exported=\"true\" ");
```
ewpatton, line 1221: I wonder if it makes sense to only export Screen1. On the other hand, people might have programs that deep link into App Inventor projects in which case every Screen would need to be exported.

appinventor/components/src/com/google/appinventor/components/scripts/ComponentProcessor.java
```diff
@@ -2229,7 +2229,7 @@ private static String activityElementToString(ActivityElement element)
       throws IllegalAccessException, InvocationTargetException {
     // First, we build the <activity> element's opening tag including any
     // receiver element attributes.
-    StringBuilder elementString = new StringBuilder("    <activity ");
+    StringBuilder elementString = new StringBuilder("    <activity android:exported=\\\"true\\\" ");
```
ewpatton, line 2232: I'm not sure this is the correct behavior. In particular, the `@ActivityElement` does provide a `exported` field, which defaults to the empty strong (so it's not written). If someone explicitly sets that field, I think this causes issues in the XML parsing (but I haven't tested, so maybe not). We might want to adjust the annotation definition to have the default be "true" rather than "", or maybe even "false" so that it's more conservative, i.e., making a public activity requires taking action on the developer's part.

appinventor/components/src/com/google/appinventor/components/scripts/ComponentProcessor.java
```diff
@@ -2247,7 +2247,7 @@ private static String receiverElementToString(ReceiverElement element)
       throws IllegalAccessException, InvocationTargetException {
     // First, we build the <receiver> element's opening tag including any
     // receiver element attributes.
-    StringBuilder elementString = new StringBuilder("    <receiver ");
+    StringBuilder elementString = new StringBuilder("    <receiver android:exported=\\\"true\\\" ");
```
ewpatton, line 2250: Ditto previous comment.

ewpatton (changes requested):
One note is that we should check whether services also need an explicit exported statement. If so, we'll need to adjust the logic of ServiceElement as well (see my comments on ComponentProcessor.java).

ewpatton:
I also noticed this in the documentation. I wonder if this will impact our Chromebook connectivity at all:

https://developer.android.com/about/versions/12/behavior-changes-12#android-app-links-verification-changes

ewpatton:
Re. service `android:exported`: https://developer.android.com/about/versions/12/behavior-changes-12#exported

josmas:
Hey guys, quick note: there is one more change that will probably have an affect if you start building for 31 as a target. Bluetooth permissions have been changed largely: https://developer.android.com/guide/topics/connectivity/bluetooth/permissions.

Somehow unrelated, I see that you are requesting both FINE and COARSE for location, just wondering if you need that in other places like the BLE extension? It used to be the case that location was needed but it is now optional for BLE. Maps seems to use both already.

appinventor/buildserver/src/com/google/appinventor/buildserver/Compiler.java
ewpatton: In order to not break backwards compatibility, let's leave this for now.

appinventor/buildserver/src/com/google/appinventor/buildserver/Compiler.java
ewpatton: @jisqyv Please review my comment and let me know what you'd prefer to do here. From my point of view, other apps should not be able to launch this activity so setting exported to `false` makes more sense to me.

ewpatton (commented):
@jisqyv We should also delete android-30/android.jar since it is no longer needed once we switch over to SDK 31.

ewpatton:
@jisqyv Per @josmas's earlier comment, maybe we should add BLUETOOTH_SCAN and BLUETOOTH_CONNECT to BluetoothClient and BLUETOOTH_ADVERTISE to BluetoothServer.

josmas:
The new permissions are `runtime` and need explicit user approval, so  so you might have to `askPermission` from the bluetooth components.

ewpatton:
@josmas Ugh.. that's particularly gnarly given that these permissions are only Android 12+

josmas:
hey @ewpatton yeah, you need finer control, which is a pain. You will gain flexibility though, cause location perms are not necessary anymore for Android 12+ when using bluetooth, but abstracting all this out for users will need some work.

ewpatton (approved):

ewpatton:
Hi Jose,

I had seen that and was aware of it. My understanding is this is specifically for BLE, which is not part of our main repo. The challenge is I still believe the coarse/fine constraint is true of Android < 12 but not starting with Android 12 (plus I think at some point only coarse was needed argh!). Unfortunately our system for modeling permissions I don't think can handle such fine grained rules as these and I'm worried about pushing these constraints onto App Inventor users.

PR #2733, opened 2022-09-16
Use AssetFetcher to update the Companion

jisqyv:
Instead of the client browser fetching the updated Companion and sending it down to the running Companion, we request the running Companion to perform the download and then call on the package manager to install it.

Note: Prior to building a Companion version for the Google Play Store, this change should be reverted because the store's static analysis will see the call to the package manager and Google will reject the update.

Change-Id: I8fb8518db66099d19936a46b1d779fcceba83e5b


ewpatton (approved):
LGTM

PR #2777, opened 2022-11-29
Fixes Issue #969

vivek3210:
Basically just adds a variable for the volume and returns it.


AppInventorWorkerBee:

vivek3210:
fixes issue #969

vivek3210:
.

vivek3210:
@ewpatton fixes issue #969

ewpatton:
Unfortunately, this does not address the underlying issue of reading the volume back from the underlying system API. It is the equivalent of returning the value of 50, which may not be ground truth.

PR #2789, opened 2022-12-26
Add "RemoveItemAtIndex" block to ListView 

devarsh-mavani-19:
See issue #2782
I have created a `@SimpleFunction` in listview component.


AppInventorWorkerBee:

appinventor/components/src/com/google/appinventor/components/runtime/ListView.java
```diff
@@ -99,8 +99,8 @@ public final class ListView extends AndroidViewComponent implements AdapterView.
 
   private float fontSizeMain;
   private float fontSizeDetail;
-  private String fontTypeface;
-  private String fontTypeDetail;
+  private int fontTypeface;
+  private int fontTypeDetail;
```
ewpatton, line 103: I'm not clear why this change is necessary to satisfy the request.

appinventor/components/src/com/google/appinventor/components/runtime/ListView.java
```diff
@@ -423,6 +423,19 @@ public void SelectionIndex(int index) {
     }
   }
 
+  /**
+   */
```
ewpatton, line 427: You can move the description below into the Javadoc and the documentation generator should take care of the rest.

appinventor/components/src/com/google/appinventor/components/runtime/ListView.java
```diff
@@ -423,6 +423,19 @@ public void SelectionIndex(int index) {
...
+   @SimpleFunction(
+      description = "Removes Item from list at a given idnex")
+    public void RemoveItemAtIndex(int index) {
+      if (!dictItems.isEmpty()) {
+        dictItems.remove(index);
+      } else {
+        stringItems.remove(index);
+      }
```
ewpatton, line 435: Make sure that this logic keeps the two lists in sync. We don't want to remove an element at one index if it has a corresponding representation at the same index in the other list.

appinventor/components/src/com/google/appinventor/components/runtime/ListView.java
```diff
@@ -724,7 +737,7 @@ public void FontSizeDetail(float fontSize) {
   @SimpleProperty(
           category = PropertyCategory.APPEARANCE,
           userVisible = false)
-  public String FontTypeface() {
+  public int FontTypeface() {
```
ewpatton, line 740: Unrelated change.

appinventor/components/src/com/google/appinventor/components/runtime/ListView.java
```diff
@@ -741,7 +754,7 @@ public String FontTypeface() {
           defaultValue = Component.TYPEFACE_DEFAULT + "")
   @SimpleProperty(
           userVisible = false)
-  public void FontTypeface(String typeface) {
+  public void FontTypeface(int typeface) {
```
ewpatton, line 757: Unrelated change.

appinventor/components/src/com/google/appinventor/components/runtime/ListView.java
```diff
@@ -758,7 +771,7 @@ public void FontTypeface(String typeface) {
   @SimpleProperty(
           category = PropertyCategory.APPEARANCE,
           userVisible = false)
-  public String FontTypefaceDetail() {
+  public int FontTypefaceDetail() {
```
ewpatton, line 774: Unrelated change.

appinventor/components/src/com/google/appinventor/components/runtime/ListView.java
```diff
@@ -775,7 +788,7 @@ public String FontTypefaceDetail() {
           defaultValue = Component.TYPEFACE_DEFAULT + "")
   @SimpleProperty(
           userVisible = false)
-  public void FontTypefaceDetail(String typeface) {
+  public void FontTypefaceDetail(int typeface) {
```
ewpatton, line 791: Unrelated change.

ewpatton (changes requested):

ewpatton:
@devarsh-mavani-19 In addition to my comments, you'll also want to bump the version number of the ListView component in YaVersion.java, then implement the (trivial) upgraders in YoungAndroidFormUpgrader.java and versioning.js

ewpatton:

devarsh-mavani-19:
> @ewpatton when i ran this feature in app, it is throwing an exception that "YailList does not implement removePosRange" I checked the super class of YaiList i.e. AbstractSequence.java (in kawa source code). The remove method is calling removePosRange method which throws unsupported exception. So do I need to implement the remove method in YailList from scratch or is there already kawa based function that i can use in java?
> 
> edt: 1 more way i can think of is to convert the Yailist to array using the toArray method and then remove item at index from array and recreate the list using makeList method (of Yailist class). But i think this will not be efficient.

@ewpatton do i have to implement removePosRange function in kawa as well or I just have to implement it directly in YailList.java? And which approach from above comment should I go for?

devarsh-mavani-19:
@ewpatton can you please have look at above comment?

ewpatton:
I think it might make more sense to change the type of `stringItems` from being a `YailList` to a regular `List`. I don't think there's a reason to keep the internal type as YailList, which is mainly used for content manipulated by the blocks.

halatmit:
sounds reasonable

== Hal

Hal Abelson
***@***.***
Prof. of Comp. Sci. and Eng.

devarsh-mavani-19:
> I think it might make more sense to change the type of `stringItems` from being a `YailList` to a regular `List`. I don't think there's a reason to keep the internal type as YailList, which is mainly used for content manipulated by the blocks.

@ewpatton Have converted YailList to List<String> can you please check this. I haven't refactored some existing functions like 
`elements`, `selectionIndex`, `setSelectionFromIndex`, `setSelectedIndexFromValue` because they were used by other components. So instead I just changed it as per my the requirements of ListView in a new method.

devarsh-mavani-19:
I have converted YailList to java list. Please can u review it?
Thanks

appinventor/components/src/com/google/appinventor/components/runtime/ListView.java
```diff
@@ -307,7 +307,7 @@ public void Elements(YailList itemsList) {
         }
       } else {
         // Support legacy single-string ListViews
-        stringItems = ElementsUtil.elements(itemsList, "Listview");
+        stringItems = ElementsUtil.elementsStrings(itemsList, "Listview");
```
ewpatton, line 310: We should fix the misspelling of "Listview" to "ListView"

appinventor/components/src/com/google/appinventor/components/runtime/util/ElementsUtil.java
ewpatton, line 1: There are a number of code style errors in this file. You can run `ant -Dcommit=ucr checkstyle` to have it check the code, but mostly it looks like you've missed putting whitespace before an opening curly brace. For future reference we follow the Google Java Style guide.

ewpatton (changes requested):

ewpatton:
@devarsh-mavani-19 There are a few more things that need to be done to get this ready to merge since you're adding a new block:

1. Update the LISTVIEW_COMPONENT_VERSION field of YaVersion by 1 and note the addition of the new block
2. In YoungAndroidFormUpgrader, add a check for the previous version number and update it to the new version. No change is needed.
3. In versioning.js, add the new version number and indicate that no upgrade to the blocks is necessary.

ewpatton:
@devarsh-mavani-19 We'd like to get this into the next App Inventor release. Do you think you could address the earlier feedback this week, or should one of the core devs adopt it?

ewpatton:
@gordonlu310 Also, after thinking about it I think that the SetCursorAt() function should be 1-indexed rather than 0-indexed, so we will need to subtract 1 from the incoming index. This will make use of this function compatible with the text blocks without requiring the user to do additional computations. For example:

blocks (83)

If in the TextBox1 you put "abcdefg" and then search for "cde" the cursor will end up between the "c" and "d", whereas it seems like it should show up at the start of "cde".

ewpatton (approved):

PR #2804, opened 2023-01-14
Adding-showAsInteger-property-chart2d

adirn26:
Fixes: https://github.com/mit-cml/appinventor-sources/issues/2759
Added a new Property **ShowAsIntegers** to **Chart2D Component** 
When Checked X-Axis Labels and Point labels are rendered as Integers. 
Updated Version numbers


adirn26:
@ewpatton could u please review?

AppInventorWorkerBee:

ewpatton:
Thank you @adirn26 for doing this. However, I'm thinking we want to have the option be a dropdown rather than a checkbox so that we can include other options such as date or time. Would you mind adapting this change to allow for those options as well?

ewpatton:
@adirn26 I think those should be good enough for now. We can always decide to add more options in the future if the use cases arise.

adirn26:
Added a dropdown property called **ChartValueType** which contains options **Decimal, Integer, Date and Time**
image
Created a new PropertyEditor class called **YoungAndroidChartLabelValueTypeChoicePropertyEditor**
@ewpatton Updated According to the changes requested

ewpatton (approved):

PR #2827, opened 2023-02-16
Feature/line of best fit

HanyaElhashemy:
This PR introduces the "Data Science" drawer with "Regression" and "Anomaly Detection" components.
In the future, any data science related operations will be added there as components.

1. The regression component currently has **linear regression** operations but can be extended in the future with different regression types, like quadratic regression.
The results of the linear regression can be visualized using the new "DrawLineOfBestFit" function of the "ChartData2D" component.
2. The anomaly detection component currently has Z-Score anomaly detection but can be extended in the future with other different algorithms, like K-Means Clustering or Isolation Forest.
The detected anomalies can be highlighted on the graph using the new "HighLightDataPoints" function of the "ChartData2D" component.
3. Refactored the current chart component architecture to decouple data processing from data visualization by introducing the new superclasses "DataModel" and  "DataCollection". This UML Class diagram shows a simplified overview of the architecture:
App_Inventor_Data_Science_Software_Architecture-Copy of Page-1 drawio
The new blocks can be tested on the following test server: https://datasci.appinventor.mit.edu/


AppInventorWorkerBee:

ewpatton:
@HanyaElhashemy Why was master merged into ucr? I don't recall anything specifically in master but not ucr that should be needed for your feature.

ewpatton:

appinventor/components/src/com/google/appinventor/components/runtime/AnomalyDetection.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+    variance /= data.size();
+
+    double sd = Math.sqrt(variance);
+
+    // Detect anomalies using Z-score
+    for (int i = 0; i < data.size(); i++) {
+      double zScore = Math.abs((data.get(i) - mean) / sd); // The z-score is a measure of how many standard deviations a data point is away from the mean
+      if (threshold.length == 0) {
+        double defaultThreshold = 1; // Todo: define a default value if not defined by the user
```
daki7711, line 82: z value of 1 seems too harsh for an anomaly, what about 2?

appinventor/components/src/com/google/appinventor/components/runtime/Regression.java
```diff
@@ -0,0 +1,135 @@
...
+      double prediction = slope * x.get(i) + intercept;
+      predictions.add(prediction);
+    }
+    // ToDo: return the user a dictionary and let the user extract the parameters they need themselves instead
+    ArrayList result = new ArrayList<>();
+    result.add(slope);
+    result.add(intercept);
+    result.add(corr);
+    result.add(predictions);
+    return result;
```
daki7711, line 90: Wouldn't kids be confused what number is what? Wouldn't implementing something like this https://algs4.cs.princeton.edu/code/edu/princeton/cs/algs4/LinearRegression.java.html be more helpful? and there are separate blocks that calculate the slope, intercept, ...

mark-friedman:
> blocks (73)

Why not combine the call with the multiple values into a single block, sort of like event handler blocks?

ewpatton:
That's also a possibility. Either option will require making modifications to how the component processor works. I think the destructuring approach is slightly nicer because it could allow any existing blocks to be seamlessly upgraded to be useful, whereas changing a method from return a set of values to having a nested statement input (and presumably shaped like a statement) would be a more complicated upgrade.

mark-friedman:
> That's also a possibility. Either option will require making modifications to how the component processor works.

Agreed.

> I think the destructuring approach is slightly nicer because it could allow any existing blocks to be seamlessly upgraded to be useful ...

True, but I guess I was just thinking that from the user's point of view it really is like the call is returning multiple values, similar to how event handlers can provide multiple values.

ewpatton:
@mark-friedman So I sketched out a prototype of that as well and I think there is a limitation there. With the use of a variable-like structure we could provide for versions that both have a statement body and a value body. However, if we allow the method block to have an nested input, we don't have a way currently to allow for us to change whether its a statement or value, and if one or the other the outer shape of the block would probably need to reflect that as well. If we do allow for the value format, it would also need to be clarified, somehow, that the value being returned from the method block is the return value of the indented input scope, not the value of the method itself.

mark-friedman:
@ewpatton I'm not sure that the situation is so different than the current situation of having two versions of the local variable block.  However, that would mean having two versions of every `call` block the returned multiple values, which might be a lot of extra blocks.  If you envision that there are going to be lots of these then I would go with pragmatic approach that you describe, though I think I would add another, value producing, `"initialize variables from result of"` block, rather than forcing users in that situation to manually extract the values from the JSON.

FWIW, this was exactly why I created the "chameleon" blocks for my Scheme Blocks project, but I wouldn't propose such a radical change to App Inventor.

appinventor/components/src/com/google/appinventor/components/runtime/Regression.java
pavi2410: Looks great! I assume this is modeled similar to destructing in JS.

appinventor/components/tests/com/google/appinventor/components/runtime/ChartData2DTest.java
```diff
@@ -50,7 +50,7 @@ public void setUp() {
     super.setUp();
     chartComponent = new Chart(getForm());
     data = new ChartData2D(chartComponent);
-    model = data.chartDataModel;
+    model = data.dataModel;
```
daki7711, line 53: So I assume this is where Jenken's fails

daki7711:
The linear regression and outlier detection icons needs to be updated in the palette
linear regression
outlier
How do these look?

appinventor/components/src/com/google/appinventor/components/runtime/AnomalyDetection.java
HanyaElhashemy: What do you mean by "too harsh"?
image

daki7711: Well if you are setting any data that is over z-score 1 as outliers, then given that you have enough samples and your data forms a Gaussian Distribution, you are basically saying 31.74 percent of the data are outliers. Z score should at least be over 2 to consider it as outliers and based on your image it should be more than 3.

HanyaElhashemy: My approach was to set a low threshold to detect more outliers, so learners (kids) see the effect of Z-score anomaly detection even if their data is “not perfect”. Then they can critically question the output of anomaly detection and set a higher threshold using the block that takes a threshold value as input:
image

From a learner point of view, it might be very frustrating to use a block and get nothing back as output.
As a compromise to your comment, I could set the default threshold to 2, but we would need to add a very good explanation on block hover to explain this to the learners and not risk frustrating them.

daki7711: Maybe for the workshop, we can set a default z-score of 1 just for the purpose of understanding outliers. Since I doubt someone will come up with a large enough data where z-score less then 2 happens. But for this to be merged and published default z-score should be at least 2 (in fact, more than that more like 2.5~3) since even z-score of 2 means you are regarding around 5% of the data as anomoly.

appinventor/components/src/com/google/appinventor/components/runtime/ChartData2D.java
```diff
@@ -40,6 +50,7 @@ public ChartData2D(Chart chartContainer) {
     dataFileColumns = Arrays.asList("", "");
     sheetsColumns = Arrays.asList("", "");
     webColumns = Arrays.asList("", ""); // Construct default webColumns list with 2 entries
```
ewpatton, line 52: See note on AnomalyDetection about moving this initialization into the DataCollection class.

appinventor/components/src/com/google/appinventor/components/runtime/ChartData2D.java
```diff
@@ -62,10 +73,12 @@ public void AddEntry(final String x, final String y) {
       public void run() {
         // Create a 2-tuple, and add the tuple to the Data Series
         YailList pair = YailList.makeList(Arrays.asList(x, y));
-        chartDataModel.addEntryFromTuple(pair);
+        dataModel.addEntryFromTuple(pair);
 
         // Refresh Chart with new data
         refreshChart();
+        // Set new Entry color to the initial default chart color
+        setEntryDefaultColor(x,y);
```
ewpatton, line 81: Revisit this based on our discussion in the office 2023-06-21

appinventor/components/src/com/google/appinventor/components/runtime/ChartData2D.java
```diff
@@ -145,7 +166,7 @@ public Boolean call() {
   @SuppressWarnings("checkstyle:ParameterName")
   @SimpleFunction()
   public void ImportFromDataFile(final DataFile dataFile, String xValueColumn,
-      String yValueColumn) {
+                                 String yValueColumn) {
```
ewpatton, line 169: Correct the indentation back to what it was. Continuations of parameter lines should be +4 from the start of the previous line.

appinventor/components/src/com/google/appinventor/components/runtime/ChartData2D.java
```diff
@@ -167,7 +188,7 @@ public void ImportFromDataFile(final DataFile dataFile, String xValueColumn,
    */
   @SimpleFunction
   public void ImportFromSpreadsheet(final Spreadsheet spreadsheet, String xColumn, String yColumn,
-      boolean useHeaders) {
+                                    boolean useHeaders) {
```
ewpatton, line 191: Ditto.

appinventor/components/src/com/google/appinventor/components/runtime/Regression.java
```diff
@@ -0,0 +1,125 @@
+// -*- mode: java; c-basic-offset: 2; -*-
+// Copyright 2022-2023 MIT, All rights reserved
+// Released under the Apache License, Version 2.0
+// http://www.apache.org/licenses/LICENSE-2.0
+
+package com.google.appinventor.components.runtime;
+
+import com.google.appinventor.components.annotations.*;
+import com.google.appinventor.components.common.*;
```
ewpatton, line 9: We only use single class imports, not wildcards.

appinventor/components/src/com/google/appinventor/components/runtime/DataCollection.java
```diff
@@ -0,0 +1,1047 @@
...
+      Log.e(this.getClass().getName(), e.getMessage());
+    }
+
+    return new YailList();
+  }
+
+  /**
+   * Imports data from the specified TinyDB component by taking the value
+   * identified by the specified tag value.
+   * <p>
```
ewpatton, line 604: Don't use `<p>` tags in the Javadocs for block methods. Instead, indent the following paragraph with two spaces.

appinventor/components/src/com/google/appinventor/components/runtime/DataModel.java
```diff
@@ -0,0 +1,379 @@
+package com.google.appinventor.components.runtime;
```
ewpatton, line 1: Add preamble comments to this file. See other files for examples.

appinventor/components/tests/com/google/appinventor/components/runtime/AnomalyDetectionTest.java
```diff
@@ -0,0 +1,65 @@
+package com.google.appinventor.components.runtime;
```
ewpatton, line 1: Add preamble comments.

appinventor/components/tests/com/google/appinventor/components/runtime/AnomalyDetectionTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+import com.google.appinventor.components.runtime.util.YailList;
+import gnu.lists.Pair;
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.Assert.*;
```
ewpatton, line 11: Do not use wildcard imports.

appinventor/components/tests/com/google/appinventor/components/runtime/AnomalyDetectionTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+public class AnomalyDetectionTest extends RobolectricTestBase {
+  AnomalyDetection anomalyDetection = new AnomalyDetection(getForm());
+  YailList xList = YailList.makeList(Arrays.asList("1", "2", "3","4","5","6"));
+  YailList yList = YailList.makeList(Arrays.asList("1", "2", "3","2","2","88"));
```
ewpatton, line 16: You should put the initialization into a `setUp()` function and make sure to call `super.setUp()`.

appinventor/components/tests/com/google/appinventor/components/runtime/AnomalyDetectionTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+  @Test
+  public void testDetectAnomalies(){
+    ArrayList anomalies = (ArrayList) anomalyDetection.DetectAnomalies( yList, 2);
+    ArrayList expectedAnomalies = new ArrayList<Pair>();
+    expectedAnomalies.add(Arrays.asList(6, 88f));
+    assertEquals(expectedAnomalies.size(), anomalies.size());
+    // Assert.assertEquals(expectedAnomalies, anomalies);
```
ewpatton, line 24: Uncomment this.

appinventor/components/tests/com/google/appinventor/components/runtime/AnomalyDetectionTest.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+  @Test
```
ewpatton, line 26: Add blank lines between methods.

appinventor/components/tests/com/google/appinventor/components/runtime/AnomalyDetectionTest.java
```diff
@@ -0,0 +1,65 @@
...
+  public void testDetectMultipleAnomalies(){
+    YailList yList = YailList.makeList(Arrays.asList("1", "2", "78","2","2","88"));
```
ewpatton, line 28: Spacing between arguments is a bit off here.

appinventor/components/tests/com/google/appinventor/components/runtime/RegressionTest.java
```diff
@@ -0,0 +1,51 @@
+package com.google.appinventor.components.runtime;
```
ewpatton, line 1: Add preamble comment.

appinventor/components/tests/com/google/appinventor/components/runtime/RegressionTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+
+import com.google.appinventor.components.common.LOBFValues;
+import com.google.appinventor.components.runtime.util.YailList;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
```
ewpatton, line 8: Static imports should come before class imports.

appinventor/components/tests/com/google/appinventor/components/runtime/RegressionTest.java
```diff
@@ -0,0 +1,51 @@
+package com.google.appinventor.components.runtime;
+
+
+import com.google.appinventor.components.common.LOBFValues;
+import com.google.appinventor.components.runtime.util.YailList;
+import org.junit.Test;
```
ewpatton, line 6: Sort imports alphabetically.

appinventor/components/tests/com/google/appinventor/components/runtime/RegressionTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class RegressionTest extends RobolectricTestBase {
+  Regression regression = new Regression(getForm());
+  LOBFValues value;
+  YailList xList = YailList.makeList(Arrays.asList("1", "2", "3","4","5","6"));
+  YailList yList = YailList.makeList(Arrays.asList("1", "2", "3","2","2","88"));
```
ewpatton, line 18: Move the initialization into the `setUp()` method and make sure to call `super.setUp()`.

appinventor/components/tests/com/google/appinventor/components/runtime/RegressionTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+  @Test(expected = Exception.class)
+  public void testCalculateLineOfBestFitWrongInputSize() throws Exception{
+    YailList xListWrongSize = YailList.makeList(Arrays.asList("1", "2", "3","4","5"));
+    regression.CalculateLineOfBestFitValue(xListWrongSize, yList, value.Slope);
+  }
+  @Test
```
ewpatton, line 25: Add blank lines between methods.

appinventor/components/src/com/google/appinventor/components/runtime/Regression.java
```diff
@@ -0,0 +1,125 @@
...
+
+  /**
+   * Creates a new Regression component.
+   */
+  public Regression(ComponentContainer container) {
+    super();
+    // Construct default dataFileColumns list with 2 entries
+    dataFileColumns = Arrays.asList("", "");
+    sheetsColumns = Arrays.asList("", "");
+    webColumns = Arrays.asList("", ""); // Construct default webColumns list with 2 entries
```
ewpatton, line 39: See previous comments on AnomalyDetection and ChartData2D about how these fields are initialized.

appinventor/components/src/com/google/appinventor/components/runtime/Regression.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  /**
+   * Calculates the line of best fit
```
ewpatton, line 43: End Javadoc with period.

appinventor/components/src/com/google/appinventor/components/runtime/Regression.java
```diff
@@ -0,0 +1,125 @@
...
+   */
+  public static YailDictionary ComputeLineOfBestFit(final YailList xEntries, final YailList yEntries) {
+    LList xValues = (LList) xEntries.getCdr();
+    List<Double> x = castToDouble(xValues);
+
+    LList yValues = (LList) yEntries.getCdr();
+    List<Double> y = castToDouble(yValues);
+
+    if (xValues.size() != yValues.size())
+      throw new IllegalStateException("Must have equal X and Y data points");
```
ewpatton, line 57: Need to also check that there are at least 2 points in the dataset as well.
Code style: wrap the body of the if statement in curly braces.

appinventor/components/src/com/google/appinventor/components/runtime/Regression.java
```diff
@@ -0,0 +1,125 @@
...
+
+    YailDictionary resultDic = new YailDictionary();
+    resultDic.put("slope",slope);
+    resultDic.put("Yintercept", intercept);
+    resultDic.put("correlation coefficient", corr);
+    resultDic.put("predictions", predictions);
+
+    return resultDic;
+  }
+  /**
```
ewpatton, line 96: Blank line between method definitions.

appinventor/components/src/com/google/appinventor/components/runtime/Regression.java
```diff
@@ -0,0 +1,125 @@
...
+   * Returns one of the Line of Best Fit values. A value could be "slope", "Yintercept", "correlation coefficient", "predictions" or a dictionary with all values above if nothing specific provided
+   *
+   * @param xList - the list of x values
+   * @param yList - the list of y values
+   * @param value - the string name of the line of best fit property
+   * @return Double slope
+   */
+  @SimpleFunction(description = "Returns one of the Line of Best Fit values. A value could be \"slope\", \"Yintercept\", \"correlation coefficient\"or \"predictions\". The block returns the complete dictionary with all values if no specific value string is provided")
```
ewpatton, line 104: Line wrap at 100 characters.

appinventor/components/src/com/google/appinventor/components/runtime/Regression.java
```diff
@@ -0,0 +1,125 @@
...
+    if (result.containsKey(value.toUnderlyingValue())){
+        return result.get(value.toUnderlyingValue());
+    }else{
+      return result;
+    }
+  }
+
+  @Override
+  public HandlesEventDispatching getDispatchDelegate() {
+    return null;
```
ewpatton, line 116: Do not override this. One of the superclasses should hold a reference to `form` and return that for `getDispatchDelegate`.

appinventor/components/src/com/google/appinventor/components/runtime/Regression.java
```diff
@@ -0,0 +1,125 @@
...
+import com.google.appinventor.components.runtime.util.YailList;
+import gnu.lists.LList;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * A data science component to apply different regression models
+ * The component requires a data source to apply the model on
```
ewpatton, line 20: Since this will be shown to the user in the help bubble, we may want to be more clear about how it is used.

appinventor/components/src/com/google/appinventor/components/runtime/ChartData2D.java
```diff
@@ -199,4 +220,68 @@ public void ImportFromWeb(final Web web, String xValueColumn, String yValueColum
 
     importFromWebAsync(web, columns);
   }
+
+  /**
+   * Draws the line of best fit
```
ewpatton, line 225: End Javadoc with a period.

appinventor/components/src/com/google/appinventor/components/runtime/ChartData2D.java
```diff
@@ -199,4 +220,68 @@ public void ImportFromWeb(final Web web, String xValueColumn, String yValueColum
...
+    }
+    YailList predictionPairsList = YailList.makeList(predictionPairs);
+    dataModel.importFromList(predictionPairsList);
+    if (dataModel.getDataset() instanceof LineDataSet) {
+      ((LineDataSet) dataModel.getDataset()).setDrawCircles(false);
+      ((LineDataSet) dataModel.getDataset()).setDrawValues(false);
+    }
+    refreshChart();
+  }
+  /**
```
ewpatton, line 246: Add a blank line between method definitions.

appinventor/components/src/com/google/appinventor/components/runtime/ChartData2D.java
```diff
@@ -199,4 +220,68 @@ public void ImportFromWeb(final Web web, String xValueColumn, String yValueColum
...
+   * Highlights all given data points on the Chart in the color of choice
+   *
+   * @param dataPoints - the list of data points. A data point inside this list is a pair of point index and point value: [[point index, point value],...,[,]]
+   * @param color - the highlight color chosen by the user
+   */
+  @SimpleFunction(description = "Highlights data points of choice on the Chart in the color of choice. This block expects a list of data points, each data point is an index, value pair")
```
ewpatton, line 252: Line wrap at 100 chars.

appinventor/components/src/com/google/appinventor/components/runtime/DataCollection.java
```diff
@@ -0,0 +1,1047 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+
+/**
+ * Data Source that is given to the Data Science component for analysis or directly to the chart component for visualisation
+ */
+@SuppressWarnings({"TryWithIdenticalCatches", "checkstyle:JavadocParagraph"})
+@SimpleObject
+public abstract class DataCollection<C extends ComponentContainer, E extends Entry, M extends DataModel>
+implements Component, DataSourceChangeListener {
```
ewpatton, line 39: Indent +4

appinventor/components/src/com/google/appinventor/components/runtime/ChartDataBase.java
```diff
@@ -122,18 +61,6 @@ protected ChartDataBase(Chart chartContainer) {
     threadRunner = Executors.newSingleThreadExecutor();
```
ewpatton, line 61: Since the threadRunner is initialized in `DataCollection()` we shouldn't create another one here.

appinventor/components/src/com/google/appinventor/components/runtime/DataCollection.java
```diff
@@ -0,0 +1,1047 @@
...
+
+    importFromDataFileAsync(dataFile, columns);
+  }
+
+  /**
+   * Imports data from the specified Spreadsheet component by taking the specified x column
+   * for the x values, and the specified y column for the y values. Prior to calling this function,
+   * the Spreadsheet component's ReadSheet method has to be called to load the data. The usage of
+   * the GotSheet event in the Spreadsheet component is unnecessary.
+   * <p>
```
ewpatton, line 811: Ditto markdown comment above.

appinventor/components/src/com/google/appinventor/components/runtime/DataCollection.java
```diff
@@ -0,0 +1,1047 @@
...
+
+    importFromSpreadsheetAsync(sheet, columns, useHeaders);
+  }
+
+  /**
+   * Imports data from the specified Web component by taking the specified x column
+   * for the x values, and the specified y column for the y values. Prior to calling this function,
+   * the Web component's Get method has to be called to load the data. The usage of the gotValue
+   * event in the Web component is unnecessary.
+   * <p>
```
ewpatton, line 832: Ditto.

appinventor/components/src/com/google/appinventor/components/runtime/ChartDataModel.java
```diff
@@ -57,15 +58,6 @@ public abstract class ChartDataModel<
   */
   protected int maximumTimeEntries = 200;
```
ewpatton, line 59: Note that `entries` and `maximumTimeEntries` were moved up to DataModel so they should not be declared here.

appinventor/components/src/com/google/appinventor/components/runtime/DataModel.java
```diff
@@ -0,0 +1,379 @@
...
+  protected List<E> entries;
+  /**
+   * Limit the maximum allowed real-time data entries
+   * Since real-time data comes in fast, the case of
+   * multi-data source input is unhandled since it's
+   * better to avoid it.
+   */
+  protected int maximumTimeEntries = 200;
+
+  public abstract void clearEntries();
```
ewpatton, line 28: Move this abstract method down after the constructor, alphabetically sorted between the add and get methods.

ewpatton (changes requested):

appinventor/components/src/com/google/appinventor/components/common/LOBFValues.java
```diff
@@ -0,0 +1,34 @@
...
+
+  LOBFValues(String lobfV) {
+    this.lobfValues = lobfV;
+  }
+
+  public String toUnderlyingValue() {
+    return lobfValues;
+  }
+
+  private static final Map<String, LOBFValues> lookup = new HashMap<>();
```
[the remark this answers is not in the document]
HanyaElhashemy: Sure, I can change it. I followed here the same style as in the other files of common directory (check MapFeature for example) to stay consistent though.

appinventor/components/src/com/google/appinventor/components/runtime/DataCollection.java
```diff
@@ -0,0 +1,1047 @@
...
+  private boolean isKeyValid(String key) {
+    // The key should either be equal to the local key, or null.
+    return (key == null || key.equals(dataSourceKey));
+  }
+
+
+  /**
+   * Casts list items to doubles
+   */
+  public static List<Double> castToDouble(List list) {
```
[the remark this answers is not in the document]
HanyaElhashemy: Could you please explain to me why? The idea of this function is to cast a list to a List of Double elements.

appinventor/components/src/com/google/appinventor/components/runtime/AnomalyDetection.java
```diff
@@ -0,0 +1,142 @@
...
+      throw new IllegalStateException("Must have equal X and Y data points");
+
+    int index = (int) GetAnomalyIndex(anomaly);
+
+    xData.remove(index - 1);
+    yData.remove(index - 1);
+
+    ArrayList cleanData = new ArrayList<Pair>();
+
+    if (xData.size() == yData.size()) {
```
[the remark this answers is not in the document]
HanyaElhashemy: This case is handled in the code above (line 100 -102)

appinventor/components/src/com/google/appinventor/components/common/LOBFValues.java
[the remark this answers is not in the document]
ewpatton: @HanyaElhashemy Ok. let's leave it as is then.

appinventor/components/src/com/google/appinventor/components/runtime/Chart.java
```diff
@@ -221,6 +221,42 @@ public void Type(ChartType type) {
     }
   }
 
+  /**
```
ewpatton, line 224: Since it looks like the code was just moved from line 575 to 224, it seems like we may just want to revert the changes to this file.

appinventor/components/src/com/google/appinventor/components/runtime/ChartView.java
```diff
@@ -9,7 +9,6 @@
 import android.os.AsyncTask;
 import android.os.Handler;
 import android.os.Looper;
-
```
ewpatton, line 12: Since this whitespace-only change is the sole modification in this file, please revert the file to match ucr.

appinventor/components/src/com/google/appinventor/components/runtime/ComponentContainer.java
```diff
@@ -52,4 +52,5 @@ public interface ComponentContainer {
 
   int Height();
 
+
```
ewpatton, line 55: Since this whitespace-only change is the sole modification in this file, please revert the file to match ucr.

appinventor/components/src/com/google/appinventor/components/runtime/AnomalyDetection.java
```diff
@@ -0,0 +1,142 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import gnu.lists.Pair;
+
+
+/**
+ * A data science component to apply different anomaly detection models
+ * The component needs a data source to apply the model on
+ */
```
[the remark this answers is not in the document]
HanyaElhashemy: I modified the description a bit, but I think when people start using it and provide us feedback, it will become clearer which parts need a more detailed description

ewpatton (approved):

daki7711:
#2957 This issue should be addressed for this PR or a future PR

PR #2830, opened 2023-02-17
Hard reset confirmation button should be "Reset"-Resolved

dhruvpatidar359:
Fixes #1228 , this PR has added translation in proper manner , also including the translation for polisky.
Thus changes the ok - reset in all cases except the korean one.

In this the cancel,ok,factory_reset text is also changed for polisky.

PR #2842, opened 2023-03-07
Hard reset confirmation button should be "Reset"-Resolved 

dhruvpatidar359:
Fixes https://github.com/mit-cml/appinventor-sources/pull/1228 , this PR has added translation in proper manner , also including the translation for polisky.
Thus changes the ok - reset in all cases except the korean one.

In this the cancel,ok,factory_reset text is also changed for polisky.

This is referencing the older PR #2830


AppInventorWorkerBee:

SusanRatiLane (approved):
LGTM

PR #2844, opened 2023-03-08
added PasswordMode property to textbox with versioning done

### LINKED ISSUE #2102: Add PasswordMode property to TextBox component.

It would be useful if we merge the Password TextBox component and the TextBox component by adding a PasswordMode property in the TextBox.

When the Password Mode property is checked, the TextBox will behave as a Password TextBox. Otherwise, It will be a normal TextBox.

Currently, there are two different components i.e. TextBox and PasswordTextBox for the respective purposes.

It would be convenient for the user if he/she can switch the input type to password by just switching a property instead of changing the component.

dhruvpatidar359:
closes #2102 
This PR is adding PasswordMode property to the TextBox class with conditions involved  and version files has been changed

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/youngandroid/YoungAndroidFormUpgrader.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/youngandroid/YoungAndroidFormUpgrader.java
@@ -1783,6 +1783,10 @@ public final class YoungAndroidFormUpgrader {
       // ReadOnly property was added
       srcCompVersion = 6;
     }
+    if (srcCompVersion < 7) {
+      // PasswordMode property was added
+      srcCompVersion = 7;
+    }
     return srcCompVersion;
   }
 
--- a/appinventor/blocklyeditor/src/versioning.js
+++ b/appinventor/blocklyeditor/src/versioning.js
@@ -3054,7 +3054,10 @@ Blockly.Versioning.AllUpgradeMaps =
     5: "noUpgrade",
 
     // AI3: Added ReadOnly property
-    6: "noUpgrade"
+    6: "noUpgrade",
+
+    // AI3: Added PasswordMode property
+    7: "noUpgrade"
 
   }, // End TextBox upgraders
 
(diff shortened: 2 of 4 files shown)


AppInventorWorkerBee:

jisqyv:

dhruvpatidar359:
image

It says the NumberOnly property test is failed , but the thing is i have added the PasswordMode property . why it is giving the error for the NumberOnly then .

By the way , should i also have to add the PasswordMode property test like this in the **TextBoxTest file**? 
image

SusanRatiLane:
> image
> 
> It says the NumberOnly property test is failed , but the thing is i have added the PasswordMode property . why it is giving the error for the NumberOnly then .
> 
> By the way , should i also have to add the PasswordMode property test like this in the **TextBoxTest file**? image

Does that test run correctly on a branch without your changes?

dhruvpatidar359:
> > image
> > It says the NumberOnly property test is failed , but the thing is i have added the PasswordMode property . why it is giving the error for the NumberOnly then .
> > By the way , should i also have to add the PasswordMode property test like this in the **TextBoxTest file**? image
> 
> Does that test run correctly on a branch without your changes?

I have to check that . How can i specifically run that test only ? or i have to run the whole  "ant  tests"

dhruvpatidar359:
> > image
> > It says the NumberOnly property test is failed , but the thing is i have added the PasswordMode property . why it is giving the error for the NumberOnly then .
> > By the way , should i also have to add the PasswordMode property test like this in the **TextBoxTest file**? image
> 
> Does that test run correctly on a branch without your changes?

yes , without changes the tests are running good

SusanRatiLane:
> > > image
> > > It says the NumberOnly property test is failed , but the thing is i have added the PasswordMode property . why it is giving the error for the NumberOnly then .
> > > By the way , should i also have to add the PasswordMode property test like this in the **TextBoxTest file**? image
> > 
> > 
> > Does that test run correctly on a branch without your changes?
> 
> yes , without changes the tests are running good

The test said it was expecting an answer of 1 and received 145. I think the best thing to do is determine what exactly that test is and examine how your change might have affected the answer.

SusanRatiLane:

PR #2873, opened 2023-04-09
Firebase component: Migrate changes from Kodular

pavi2410:
1. Add GoOnline and GoOffline methods.
2. Make FirebaseURL property user visible.

### DIFF
--- a/appinventor/blocklyeditor/src/versioning.js
+++ b/appinventor/blocklyeditor/src/versioning.js
@@ -1937,7 +1937,10 @@ Blockly.Versioning.AllUpgradeMaps =
     // AI2 Added AppendValue, RemoveFirst and FirstRemoved
     2: "noUpgrade",
     // AI2 Added ClearTag function, GetTagList and Persist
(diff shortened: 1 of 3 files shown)


appinventor/blocklyeditor/src/versioning.js
```diff
@@ -1937,7 +1937,10 @@ Blockly.Versioning.AllUpgradeMaps =
     // AI2 Added AppendValue, RemoveFirst and FirstRemoved
     2: "noUpgrade",
     // AI2 Added ClearTag function, GetTagList and Persist
-    3: "noUpgrade"
+    3: "noUpgrade",
+    // Added GoOnline and GoOffline functions
+    // Made the FirebaseURL property user visible.
```
pavi2410, line 1942: @ewpatton Does this require an upgrader?

ewpatton (commented):
I think this looks fine, but you should run `ant docs` and then commit the updated Markdown and HTML files to include the new functions.

PR #2892, opened 2023-05-19
Bugfix. Make sure extension temp file is in the APK bucket

jisqyv:
We were looking for it in the APK bucket, but we put it in the Source bucket. This change fixes that!

Change-Id: I89d68b5a5f6b369c9a57e433806d7fc563d6e9d0


jisqyv:
Tested.

ewpatton (approved):
LGTM. Since this function is only called for temporary file uploads putting it in the APK bucket should be fine.

jisqyv:
@ewpatton Thanks. The temp file literally lasts only seconds...

PR #2969, opened 2023-08-30
Make AssetBox always visible

SusanRatiLane:
AssetBox appears on both Blocks and Designer views, but code in unloadDesigner.

This is now fixed.


jisqyv (approved):
LGTM

PR #2971, opened 2023-08-31
Remove buggy build caching code

jisqyv:
As part of the UI Binder update, code was added to cache builds. In particular this code would avoid calling the buildserver if the project had not been modified since the last build.

However, it turns out that we *always* update the project modification date when we do a build. So we do not really know whether or not the project was really modified. However the caching code would get triggered if you did two builds very close in time (within a few seconds) only because the updateProjectModDate() call in ObjectifyStorageIo.java would only do an update if the last modification was a while ago. This change was made years ago to cut down on the number of API calls we made every time a file was saved.

And... When the cache code was triggered, it didn't actually result in the dialog box with the download link and barcode.

This commit removes the caching code and the updateProjectBuildDate interface in StorageIo, as it is no longer needed.

Change-Id: I6acc567c8bc52076a5d43a47f3a481c2b402112b


SusanRatiLane (approved):
LGTM

PR #2972, opened 2023-09-01
Only send Strict Transport Security Header using https

jisqyv:
If AppInventorFeatures.enableHttpRedirect() is false, don't send the Strict-Transport-Security header. Chrome appears to ignore it when it is sent over http instead of https, but I wouldn't want to depend on that behavior!

Change-Id: I585fc6164a57d7502b737c4596cbd42b84489dd2

### DIFF
--- a/appinventor/appengine/war/index.jsp
+++ b/appinventor/appengine/war/index.jsp
@@ -13,7 +13,9 @@
         }
      return;
    }
-   response.setHeader("Strict-Transport-Security", "max-age=3600");
+   if (AppInventorFeatures.enableHttpRedirect()) {
+       response.setHeader("Strict-Transport-Security", "max-age=3600");
+   }
 %>
 <!-- Copyright 2007-2009 Google Inc. All Rights Reserved. -->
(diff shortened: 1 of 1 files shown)


jisqyv:
This came up when I was merging the latest code for the standalone branch and the branches for code.appinventor.mit.edu and Coolthink

ewpatton (commented):
I reviewed the MDN docs re. HSTS at https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security#description. In particular, it notes:

> The Strict-Transport-Security header is ignored by the browser when your site has only been accessed using HTTP. Once your site is accessed over HTTPS with no certificate errors, the browser knows your site is HTTPS capable and will honor the Strict-Transport-Security header.

So while this change still includes the HSTS header on HTTPS connections, which satisfies the rules, it might result in a situation where folks have previously connected via HTTPS (sans HSTS header) but in the future connect via HTTP (with HSTS header). Since we already redirect to HTTPS only in prod, we might want to move this conditional before the initial conditional along with a check for being in prod (i.e., in dev we never send the HSTS header but we always send it in prod).

ewpatton (approved):

PR #3012, opened 2023-10-08
Download phantomJS during build on github action

josmas:
Build fails when running tests because phantomJS is not available.

Initial failures and current passing build can be seen in my [Actions dashboard](https://github.com/josmas/app-inventor/actions).


AppInventorWorkerBee:

ewpatton (approved):
LGTM

PR #3015, opened 2023-10-10
Ios layout fix

daki7711:
This branch will fix ios apps screen crashes

### DIFF
--- a/appinventor/components-ios/src/LinearView.swift
+++ b/appinventor/components-ios/src/LinearView.swift
@@ -263,7 +263,7 @@ public class LinearView: UIView {
     }
   }
 
-  open func resetView(){
+  open func resetView() {
     /* Resets the all the width*/
     widthFillParent = 0
     widthFillParentConstraint?.isActive = false
@@ -278,10 +278,6 @@ public class LinearView: UIView {
     _innerEqualConstraint.isActive = false
     _innerHeadZero.isActive = true
     _innerTailZero.isActive = false
-
-    /* Resets the outer constraints*/
-    /* This line was causing the problem for the button centering*/
-    //_outerEqualConstraint.isActive = false
   }
 
   open func removeAllItems() {
--- a/appinventor/components-ios/src/LinearView.swift
+++ b/appinventor/components-ios/src/LinearView.swift
@@ -274,10 +274,8 @@ public class LinearView: UIView {
     heightFillParentConstraint?.isActive = false
     heightFillParentConstraint = nil
 
-    /* Resets the inner constraints*/
-    _innerEqualConstraint.isActive = false
-    _innerHeadZero.isActive = true
-    _innerTailZero.isActive = false
+    updatePositioningConstraints()
+    updatePriorities()
   }
 
   open func removeAllItems() {
--- a/appinventor/components-ios/src/LinearView.swift
+++ b/appinventor/components-ios/src/LinearView.swift
@@ -280,7 +280,8 @@ public class LinearView: UIView {
     _innerTailZero.isActive = false
 
     /* Resets the outer constraints*/
-    _outerEqualConstraint.isActive = false
+    /* This line was causing the problem for the button centering*/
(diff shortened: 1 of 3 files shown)


appinventor/components-ios/src/LinearView.swift
```diff
@@ -263,6 +263,24 @@ public class LinearView: UIView {
     }
   }
 
+  open func resetView(){
+    /* Resets the all the width*/
+    widthFillParent = 0
+    widthFillParentConstraint?.isActive = false
+    widthFillParentConstraint = nil
+    _innerEqualConstraint.isActive = true
+    _innerHeadZero.isActive = false
```
ewpatton, line 272: It seems like the default should be for the head zero to be true and equal constraints to be false so that this aligns-head.

appinventor/components-ios/src/LinearView.swift
```diff
@@ -263,6 +263,24 @@ public class LinearView: UIView {
...
+    _innerTailZero.isActive = false
+
+    /* Resets the all the height*/
+    heightFillParent = 0
+    heightFillParentConstraint?.isActive = false
+    heightFillParentConstraint = nil
+    _innerEqualConstraint.isActive = true
+    _innerHeadZero.isActive = false
+    _innerTailZero.isActive = false
```
ewpatton, line 281: This seems redundant with the above.

I feel like we should also be resetting the outer constraints as well in this function.

appinventor/components-ios/tests/Unit Tests/util/LinearViewTests.swift
```diff
@@ -73,6 +73,35 @@ class LinearViewTests: XCTestCase {
     Button1.Width = kLengthFillParent
   }
 
+  func testAutomaticParent() {
```
ewpatton, line 76: Add a second test function that mimics switching between projects as we did in the office to verify the fix. This way if we make further changes and break it we will know before merging.

ewpatton (changes requested):

appinventor/components-ios/src/LinearView.swift
daki7711: What should the outer constraints be setted?

ewpatton:
In both cases the head rule should be active and the tail rule inactive since the default alignments are top and left. Sent from my iPhoneOn Oct 10, 2023, at 13:29, David Young-Jae Kim ***@***.***> wrote:﻿
@daki7711 commented on this pull request.



In appinventor/components-ios/src/LinearView.swift:
> +    _innerEqualConstraint.isActive = true
+    _innerHeadZero.isActive = false
+    _innerTailZero.isActive = false

What should the outer constraints be setted?

ewpatton (changes requested):
Two minor things otherwise LGTM.

appinventor/components-ios/src/LinearView.swift
```diff
@@ -263,6 +263,27 @@ public class LinearView: UIView {
...
+    heightFillParentConstraint = nil
+
+    /* Resets the inner constraints*/
+    _innerEqualConstraint.isActive = false
+    _innerHeadZero.isActive = true
+    _innerTailZero.isActive = false
+
+    /* Resets the outer constraints*/
+    /* This line was causing the problem for the button centering*/
+    //_outerEqualConstraint.isActive = false
```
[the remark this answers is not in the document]
daki7711: so we will not be concerned if that is the right fix as of now?

ewpatton:
Unfortunately, something still isn't quite right in reseting of the view. If you test the project linked from #3054, the label should be centered both horizontally and vertically, but it is only centered horizontally. Rather than activating/deactivating the constraints at the end of the function, we may want to call `updatePositioningConstraints` and `updatePriorities` to reset to the desired state.

ewpatton:
@daki7711 Let me know if you agree with these changes and then we can merge.

daki7711:
I'm good with it

ewpatton (approved):

PR #3038, opened 2023-10-24
Fix Tree Structure

barreeeiroo:
- Fixes missing Rename/Delete buttons
- Brings back expand/collapse buttons

Known issues:

- Now All/Visible/Non-visible state is on a per-form level (avoids resetting when the tree gets updated), but this does not get reflected on the button.

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockForm.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockForm.java
@@ -53,6 +53,8 @@ import java.util.Map;
 public final class MockForm extends MockContainer {
   private static final Logger LOG = Logger.getLogger(MockForm.class.getName());
 
+  private Integer view = 1;
+
   /*
    * Widget for the mock form title bar.
    */
@@ -1341,7 +1343,7 @@ public final class MockForm extends MockContainer {
    * @return  tree showing the component hierarchy of the form
    */
   public TreeItem buildComponentsTree() {
-    return buildComponentsTree(1);
+    return buildComponentsTree(view);
   }
 
   /**
@@ -1351,7 +1353,12 @@ public final class MockForm extends MockContainer {
    * @return  tree showing the component hierarchy of the form
    */
   public TreeItem buildComponentsTree(Integer view) {
+    this.view = view;
     return buildTree(view);
+  }
+
+  public Integer getView() {
+    return this.view;
   }
 
   // PropertyChangeListener implementation
--- a/appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockForm.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockForm.java
@@ -1357,10 +1357,6 @@ public final class MockForm extends MockContainer {
     return buildTree(view);
   }
 
-  public Integer getView() {
-    return this.view;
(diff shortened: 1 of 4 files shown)


appinventor/appengine/src/com/google/appinventor/client/explorer/SourceStructureExplorer.java
```diff
@@ -68,10 +68,14 @@ public void onBrowserEvent(Event event) {
...
 
+  public SourceStructureExplorer() {
+    this(true);
+  }
+
   /**
    * Creates a new source structure explorer.
    */
-  public SourceStructureExplorer() {
+  public SourceStructureExplorer(boolean includeButtonPanel) {
```
SusanRatiLane, line 78: You restored the includeButtonPanel parameter, which is commented out in the code. Is there some reason to keep it?

barreeeiroo: So, ideally, those buttons should be located somewhere else, not at the bottom of that panel. That's why the conditional rendering, even though now we are always adding them as we don't have any other location.

SusanRatiLane: > So, ideally, those buttons should be located somewhere else, not at the bottom of that panel. That's why the conditional rendering, even though now we are always adding them as we don't have any other location.

Hmm. If we allow the buttons to move in the future, it will probably be done a different way using UIBinder. I don't think leaving in the unused boolean will save us any time if we add this functionality later.

barreeeiroo: Okay, I'll get rid of it

appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockForm.java
```diff
@@ -1351,9 +1353,14 @@ public TreeItem buildComponentsTree() {
    * @return  tree showing the component hierarchy of the form
    */
   public TreeItem buildComponentsTree(Integer view) {
+    this.view = view;
     return buildTree(view);
   }
 
+  public Integer getView() {
```
SusanRatiLane, line 1360: This does not appear to be used.

Could you talk through the reasoning on introducing a new view attribute?

barreeeiroo: So, there was another issue: when you drag and drop an element to the Mock, the state of the tree is not preserved. For example, if you were viewing only Visible components, the tree will reset to the All components on any tree change.

This `view` variable preserves the selected view state on a per-Form level, avoiding this unintended behavior.

SusanRatiLane: OK, that makes sense. the getter here doesn't appear to be called anywhere. Is that correct? I think we can skip adding a getter until we need access the attribute outside the class.

barreeeiroo: Resolved. It was added as I was running some tests, but didn't achieve anything (fixing the Known Issue mentioned above).

SusanRatiLane (approved):
LGTM

PR #3063, opened 2023-11-21
Webview changes for reload, stop loading, javascript and clear cookies.

Codnelle:


AppInventorWorkerBee:

PR #3069, opened 2023-12-07
Ukrainian Translation

SusanRatiLane:


ewpatton (approved):
The technical updates LGTM. I put some of the strings through Google Translate to verify. I think this iteration is good to go.

PR #3079, opened 2023-12-20
User-selectable modern ui style

SusanRatiLane:
This PR contains the new modern ui that began as a project explorer prototype created by Vishwas Adiga during Google Summer of Code 2021. 

It includes ui switching and dark mode support by Jennet Zamanova. Dark mode needs more review of its color selection before it is ready for release. 

The user interface can be switched using the new "User Interface Settings" option under Settings (which is the gear icon in the modern ui).

Closes #2514 
Note: Include authorship attribution for Vishwas Adiga.
Also include Jennet Zamanova as collaborator.


mark-friedman:
It's not the most important thing, but I notice that there are a bunch of files/folders with "GSoC" in their name.  I understand why they may have started that way but perhaps they should be renamed something more meaningful before this all gets merged.

SusanRatiLane:
> It's not the most important thing, but I notice that there are a bunch of files/folders with "GSoC" in their name. I understand why they may have started that way but perhaps they should be renamed something more meaningful before this all gets merged.

Yes, Evan and I were just talking about that. They're going to be renamed. The question is to what. We've also used "modern," but that seems excessively generic to me. In theory, any new UI is going to be more modern than the last.

So, really, what do we want to call it? I've kicked around something color-related, like Citrus or Key Lime, assuming we're going to keep the green accent color. (I like it a lot, but it seems to be controversial.) I thought it would be fun to come up with a word or name that evokes this decade, being that this is a big modernization. But I haven't come up with one. Artemis might be a possibility, named for the Artemis Project. 

Suggestions welcome!  I'd love to get people to throw some stuff out.

halatmit:
Android uses dessert names.

mark-friedman:
> I thought it would be fun to come up with a word or name that evokes this decade, being that this is a big modernization.

Some possibilities:
- GenZ
- TaylorSwift
- Beyonce

SusanRatiLane:
> > I thought it would be fun to come up with a word or name that evokes this decade, being that this is a big modernization.
> 
> Some possibilities:
> 
>     * GenZ

Another term for GenZ would be Zoomer.

appinventor/appengine/src/com/google/appinventor/client/style/GSoC/UIFactoryGSoC.java
```diff
@@ -0,0 +1,42 @@
...
+  @Override
+  public ProjectFolder createProjectFolder(JSONObject json, ProjectFolder parent) {
+    return new ProjectFolderGSoC(json, parent);
+  }
+
+  @Override
+  public SimpleVisibleComponentsPanel createSimpleVisibleComponentsPanel
+      (YaFormEditor editor, SimpleNonVisibleComponentsPanel nonVisPanel) {
+    return new SimpleVisibleComponentsPanelGSoC(editor, nonVisPanel);
+  }}
```
ewpatton, line 40: Code style: The closing curly brace for the class should be on its own line. Also, there's more whitespace at the end of the file here than strictly necessary.

appinventor/appengine/src/com/google/appinventor/client/style/GSoC/darkModern.css
```diff
@@ -0,0 +1,37 @@
+@def background-image                   url('static/images/squairy_dark.jpg');
```
ewpatton, line 1: We should think about adding a comment with copyright year and license here.

appinventor/appengine/src/com/google/appinventor/client/style/GSoC/darkModern.css
```diff
@@ -0,0 +1,37 @@
...
+    /* orange */
+@def blocks-designer-color              #f89406;
+@def project-background-color           #7a914a;
+    /* blue */
+@def explorer-border                    #D6E9F8;
+@def simple-panel-border                #67A7E3;
+@def project-color                      #112abb;
+    /* red */
+@def destructive-action-hover           #FF6666;
+@def item-color-red                     #ff0000;
```
ewpatton, line 37: Missing newline at end of file.

appinventor/appengine/src/com/google/appinventor/client/explorer/folder/FolderManager.java
```diff
@@ -59,7 +58,7 @@ public void loadFolders() {
     }
 
     LOG.info("folderJSON - " + folderJSON);
-    globalFolder = new ProjectFolder(folderJSON, null);
+    globalFolder = Ode.getUiFactory().createProjectFolder(folderJSON, null);
```
ewpatton, line 61: I'd like to consider using dependency injection in classes like this rather than coupling them even more tightly to Ode. I.e., we should have the constructor take the UIFactory object and store it for references like here and elsewhere in the file. Ode, when it instantiates the folder manager, can pass the factory along. The field can be `final`.

appinventor/appengine/src/com/google/appinventor/client/explorer/folder/FolderManager.java
```diff
@@ -29,7 +28,7 @@ public final class FolderManager {
 
   private boolean foldersLoaded;
 
-  private final List<FolderManagerEventListener> folderManagerEventListeners;
+  private ArrayList<FolderManagerEventListener> folderManagerEventListeners;
```
ewpatton, line 31: I don't see why this needed to be changed.

appinventor/appengine/src/com/google/appinventor/client/explorer/folder/FolderManager.java
```diff
@@ -5,14 +5,13 @@
 
 package com.google.appinventor.client.explorer.folder;
 
+import com.google.gwt.json.client.JSONObject;
```
ewpatton, line 8: The imports are not in lexicographical order as required by the Google Java Style.

appinventor/appengine/src/com/google/appinventor/client/explorer/project/Project.java
ewpatton, line 1: You could revert the whitespace only change to this file.

ewpatton (changes requested):
Here are my first 50 comments.

appinventor/appengine/src/com/google/appinventor/client/explorer/commands/ShowBarcodeCommand.java
```diff
@@ -60,14 +60,16 @@ public boolean willCallExecuteNextCommand() {
   }
 
   @Override
-  public void execute(final ProjectNode node) {
-    // Display a barcode for an url pointing at our server's download servlet
+  public void execute(final ProjectNode node)
+  {
+    // Display a barcode for an url pointing at our server's download servl}t
     String barcodeUrl = GWT.getHostPageBaseURL()
```
jisqyv, line 66: Typo: servl}t?

appinventor/appengine/src/com/google/appinventor/client/explorer/folder/FolderManager.java
SusanRatiLane: > I'd like to consider using dependency injection in classes like this rather than coupling them even more tightly to Ode. I.e., we should have the constructor take the UIFactory object and store it for references like here and elsewhere in the file. Ode, when it instantiates the folder manager, can pass the factory along. The field can be `final`.

OK. I think I understand what you're suggesting here, but I'd like to review it with you.

appinventor/appengine/src/com/google/appinventor/client/style/neo/images/delete.png
ewpatton, line 1: Ditto

appinventor/appengine/src/com/google/appinventor/client/style/neo/images/dictionaries.png
ewpatton, line 1: Ditto

appinventor/appengine/src/com/google/appinventor/client/style/neo/images/extension.png
ewpatton, line 1: Ditto

appinventor/appengine/src/com/google/appinventor/client/explorer/folder/FolderManager.java
```diff
@@ -29,9 +29,11 @@ public final class FolderManager {
 
   private boolean foldersLoaded;
 
-  private final List<FolderManagerEventListener> folderManagerEventListeners;
+  private final ArrayList<FolderManagerEventListener> folderManagerEventListeners;
```
ewpatton, line 32: I'm not sure why this was changed to ArrayList but this shouldn't be necessary.

appinventor/appengine/src/com/google/appinventor/client/explorer/folder/ProjectFolder.java
```diff
@@ -170,20 +169,27 @@ public void addProject(Project project) {
...
 
   public void refresh() {
+    LOG.info("Refresh folder " + name);
+    if (nameLabel == null) {
+      LOG.info("Name Label is null ");
+    }
     nameLabel.setText(name);
+    LOG.info("Set date created ");
     dateCreatedLabel.setText(DATE_FORMAT.format(new Date(dateCreated)));
+    LOG.info("Set date modified ");
```
ewpatton, line 179: We are doing a lot of logging here. We may want to revisit this or change the logging configuration in such a way that it won't spend much time in production.

appinventor/appengine/src/com/google/appinventor/client/explorer/folder/ProjectFolder.java
```diff
@@ -245,7 +255,6 @@ public List<Project> getSelectedProjects() {
   public List<Project> getProjects() {
     return projects;
   }
-
```
ewpatton, line 248: Not sure why this line was removed but it should be restored for code cleanliness.

appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/ProjectList.java
```diff
@@ -210,28 +220,38 @@ public void onSelectionChange(boolean selected) {
...
         continue;
       }
+      LOG.info("Set selection change handler");
       childFolder.setSelectionChangeHandler(selectionEvent);
+      LOG.info("Refresh child folder");
       childFolder.refresh();
+      LOG.info("Add child to container");
       container.add(childFolder);
+      LOG.info("Done");
     }
```
ewpatton, line 235: This is a lot of logging for an inner loop operation.

appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/ProjectListItem.java
```diff
@@ -34,18 +35,18 @@ interface ProjectListItemUiBinder extends UiBinder<FlowPanel, ProjectListItem> {
   private static final ProjectListItemUiBinder UI_BINDER =
```
ewpatton, line 35: If this is overridden by a subclass, you may want to move it within bindUI otherwise both the classic and neo versions will get bundled together.

appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/ProjectToolbar.java
```diff
@@ -9,19 +9,20 @@
 import com.google.appinventor.client.Ode;
 import com.google.appinventor.client.boxes.ProjectListBox;
 import com.google.appinventor.client.widgets.Toolbar;
-import com.google.appinventor.client.widgets.ToolbarItem;
 import com.google.gwt.core.client.GWT;
 import com.google.gwt.uibinder.client.UiBinder;
 import com.google.gwt.uibinder.client.UiField;
+import java.util.logging.Logger;
 
+import com.google.gwt.user.client.ui.Label;
```
ewpatton, line 17: Fix import order.

appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/ProjectToolbar.java
```diff
@@ -47,30 +50,44 @@ public ProjectToolbar() {
...
       if (!Ode.getInstance().getGalleryReadOnly()) {
         setButtonVisible(WIDGET_NAME_SENDTONG, true);
       }
     }
-
     setTrashTabButtonsVisible(false);
   }
 
+  protected void bindProjectToolbar()
+  {
```
ewpatton, line 65: Opening curly should be on previous line.

appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/ProjectToolbar.java
```diff
@@ -9,19 +9,20 @@
...
 
 /**
  * The project toolbar houses command buttons in the Young Android Project tab.
  *
  */
 public class ProjectToolbar extends Toolbar {
+  private static final Logger LOG = Logger.getLogger(ProjectToolbar.class.getName());
   interface ProjectToolbarUiBinder extends UiBinder<Toolbar, ProjectToolbar> {}
-
   private static final ProjectToolbar.ProjectToolbarUiBinder UI_BINDER =
```
ewpatton, line 26: I don't recall if we have a custom project toolbar at this point but if so you may want to move the initialization of this into the bindUI function so that it won't bundle both the classic and neo UIs into the same JS file.

appinventor/appengine/src/com/google/appinventor/client/style/neo/DesignToolbarNeo.java
```diff
@@ -0,0 +1,35 @@
+package com.google.appinventor.client.style.neo;
```
ewpatton, line 1: Missing copyright preamble.

appinventor/appengine/src/com/google/appinventor/client/style/neo/DesignToolbarNeo.ui.xml
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2009-2011 Google, All Rights reserved -->
+<!-- Copyright 2011-2023 MIT, All rights reserved -->
```
ewpatton, line 3: Since the neo UI is our own, we don't need to include the Google copyright here.

appinventor/appengine/src/com/google/appinventor/client/style/neo/ImagesNeo.java
```diff
@@ -0,0 +1,938 @@
+package com.google.appinventor.client.style.neo;
```
ewpatton, line 1: Add missing copyright preamble.

appinventor/appengine/src/com/google/appinventor/client/style/neo/ImagesNeo.java
```diff
@@ -0,0 +1,938 @@
...
+
+import com.google.appinventor.client.Images;
+import com.google.appinventor.client.Ode;
+import com.google.gwt.resources.client.ImageResource;
+
+public interface ImagesNeo extends Images {
```
ewpatton, line 7: Maybe we should only override the methods here of images that are actually changed?

appinventor/appengine/src/com/google/appinventor/client/style/neo/ImagesNeo.java
```diff
@@ -0,0 +1,938 @@
...
+  @Source("images/codi_vert.png")
+  ImageResource codiVert();
+
+  /**
+   * App Inventor Logo 
+   * {@link Ode#createWelcomeDialog(boolean)}
+   */
+  @Override
+  @Source("images/appinvlogo-32.png")
+  ImageResource appInventorLogo();
```
ewpatton, line 23: I don't think this is used any more.

appinventor/appengine/src/com/google/appinventor/client/style/neo/ImagesNeo.java
```diff
@@ -0,0 +1,938 @@
...
+
+  /**
+   * Close button image for the box widget
+   */
+  @Override
+  @Source("images/boxClose.png")
+  ImageResource boxClose();
```
ewpatton, line 30: See comments elsewhere about duplicated images that aren't any different from the superclass.

appinventor/appengine/src/com/google/appinventor/client/style/neo/NewYoungAndroidProjectWizardNeo.java
```diff
@@ -0,0 +1,53 @@
+package com.google.appinventor.client.style.neo;
```
ewpatton, line 1: Add missing copyright preamble.

appinventor/appengine/src/com/google/appinventor/client/style/neo/NewYoungAndroidProjectWizardNeo.ui.xml
```diff
@@ -0,0 +1,41 @@
+<!-- Copyright 2009-2011 Google, All Rights reserved -->
+<!-- Copyright 2011-2023 MIT, All rights reserved -->
```
ewpatton, line 2: Our own feature. No need to have a Google copyright on it.

appinventor/appengine/src/com/google/appinventor/client/style/neo/NewYoungAndroidProjectWizardNeo.ui.xml
```diff
@@ -0,0 +1,41 @@
...
+<!-- Released under the Apache License, Version 2.0 -->
+<!-- http://www.apache.org/licenses/LICENSE-2.0 -->
+
+<!-- As of Apr 2024, this template is not being used. CSS changes are handling all the changes
+to dialog we need currently. It is left here as an example of a dialog template in case we want
+to do that in the future. -->
```
ewpatton, line 8: Maybe put this into documentation instead rather than leaving vestigial code that was never used in the source tree.

appinventor/appengine/src/com/google/appinventor/client/style/neo/Ode.ui.xml
```diff
@@ -0,0 +1,58 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2009-2011 Google, All Rights reserved -->
+<!-- Copyright 2011-2023 MIT, All rights reserved -->
```
ewpatton, line 3: No need for assigning Google copyright to a new file.

appinventor/appengine/src/com/google/appinventor/client/style/neo/ProjectFolderNeo.java
```diff
@@ -0,0 +1,72 @@
+package com.google.appinventor.client.style.neo;
```
ewpatton, line 1: Add missing copyright preamble.

appinventor/appengine/src/com/google/appinventor/client/style/neo/Ode.ui.xml
```diff
@@ -0,0 +1,58 @@
...
+                </widgets:Toolbar>
+              </g:FlowPanel>
+              <box:SourceStructureBox ui:field="sourceStructureBox" />
+              <box:BlockSelectorBox />
+              <g:FlowPanel>
+                <g:Label text="{messages.assetListBoxCaption}" styleName="ode-AssetsTitle" />
+                <box:AssetListBox  ui:field="assetListBox" />
+              </g:FlowPanel>
+            </g:FlowPanel>
+            <box:PropertiesBox ui:field="propertiesBox" width="222px"  />
```
ewpatton, line 48: Consider moving the width setting to the CSS.

appinventor/appengine/src/com/google/appinventor/client/style/neo/Ode.ui.xml
```diff
@@ -0,0 +1,58 @@
...
+      <ed:TutorialPanel ui:field="tutorialPanel" width="100%" height="100%" visible="false" />
+      <g:DeckPanel ui:field="deckPanel" animationEnabled="true" styleName="ode-DeckPanel">
+        <g:FlowPanel styleName="ode-ProjectListView" width="100%" >
+          <neo:ProjectToolbarNeo ui:field="projectToolbar" />
+          <box:ProjectListBox ui:field="projectListbox" />
+        </g:FlowPanel>
+        <g:FlowPanel styleName="ode-ProjectEditor" ui:field="projectEditor">
+          <neo:DesignToolbarNeo ui:field="designToolbar" />
+          <g:FlowPanel ui:field="workColumns" styleName="ode-WorkColumns" >
+            <box:PaletteBox width="265px" ui:field="paletteBox"/>
```
ewpatton, line 30: Maybe consider moving the width setting to CSS.

appinventor/appengine/src/com/google/appinventor/client/style/neo/Ode.ui.xml
```diff
@@ -0,0 +1,58 @@
...
+             ui:generatedKeys="com.google.gwt.i18n.server.keygen.MethodNameKeyGenerator"
+             ui:generateLocales="default">
+  <ui:with field="style" />
+  <ui:with field="messages" type="com.google.appinventor.client.OdeMessages" />
+  <g:FlowPanel styleName="ode-Root">
+    <neo:TopPanelNeo ui:field="topPanel" />
+    <g:FlowPanel styleName="ode-TutorialWrapper" ui:field="overDeckPanel">
+      <ed:TutorialPanel ui:field="tutorialPanel" width="100%" height="100%" visible="false" />
+      <g:DeckPanel ui:field="deckPanel" animationEnabled="true" styleName="ode-DeckPanel">
+        <g:FlowPanel styleName="ode-ProjectListView" width="100%" >
```
ewpatton, line 23: Consider moving width/height settings to CSS.

appinventor/appengine/src/com/google/appinventor/client/style/neo/ProjectFolderNeo.java
```diff
@@ -0,0 +1,72 @@
...
+import com.google.gwt.uibinder.client.UiBinder;
+import com.google.gwt.uibinder.client.UiField;
+import com.google.gwt.user.client.ui.CheckBox;
+import com.google.gwt.user.client.ui.FlowPanel;
+import com.google.gwt.user.client.ui.Label;
+
+import java.util.logging.Logger;
+
+public class ProjectFolderNeo extends ProjectFolder {
+  private static final Logger LOG = Logger.getLogger(ProjectFolderNeo.class.getName());
```
ewpatton, line 20: This logger doesn't seem like it is used anywhere.

appinventor/appengine/src/com/google/appinventor/client/style/neo/ProjectFolderNeo.ui.xml
```diff
@@ -0,0 +1,25 @@
+<!-- Copyright 2009-2011 Google, All Rights reserved -->
```
ewpatton, line 1: No Google copyright on new files.

appinventor/appengine/src/com/google/appinventor/client/style/neo/ProjectListItemNeo.java
```diff
@@ -0,0 +1,51 @@
+package com.google.appinventor.client.style.neo;
```
ewpatton, line 1: Add missing copyright preamble.

appinventor/appengine/src/com/google/appinventor/client/style/neo/ProjectListItemNeo.java
```diff
@@ -0,0 +1,51 @@
...
+public class ProjectListItemNeo extends ProjectListItem {
+
+  interface ProjectListItemUiBinderNeo extends UiBinder<FlowPanel, ProjectListItemNeo> {}
+
+  private static final ProjectListItemUiBinderNeo UI_BINDER =
+      GWT.create(ProjectListItemUiBinderNeo.class);
+
+  @UiField FlowPanel container;
+  @UiField
+  Label nameLabel;
```
ewpatton, line 23: Make this consistent with the surrounding lines.

ewpatton (changes requested):

appinventor/appengine/src/com/google/appinventor/client/boxes/PaletteBox.java
```diff
@@ -38,5 +39,9 @@ private PaletteBox() {
         false,     // startMinimized
         false,     // usePadding
         false);    // highlightCaption
+
+    // UIBinder calls the private constructor, presumably through reflection, which
```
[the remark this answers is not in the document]
SusanRatiLane: > I don't understand this comment because UiBinder cannot call private constructors. The Java code is generated and so still follows all of the standard Java visibility rules.

I'm not sure what the justification is, but objects declared in UiBinder that do not have public no-parameter constructors do seem to be able to call the private constructor. This was leading to two instances floating around of classes that were supposed to be singletons.

I found a better way of handling this for most cases. I'm not sure if this is one I forgot to update or if there's a reason it didn't work. Looking at it.

appinventor/appengine/src/com/google/appinventor/client/resources/base.css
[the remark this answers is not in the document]
SusanRatiLane: > It's not clear to me the relationship between the CSS files defined in this `resources` package and the files of the same name one level up in the `client` package. Were these duplicated at some point? If so, we should determine which ones are correct and delete the dupes.

I no longer see these css files at the client directory level, so I assume this has been sorted. I'm not sure that this is the way I want them sorted, however, since there's nothing in the resources directory except these three css files.

appinventor/appengine/src/com/google/appinventor/client/widgets/properties/PropertiesPanel.java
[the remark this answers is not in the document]
SusanRatiLane: > A logger was added here but never used. Consider reverting the changes to this file.

I'm in favor of leaving in unused logger declarations because they're heavily used in debugging. It's not worth it if there are no other changes to the file, though. Reverting.

appinventor/appengine/src/com/google/appinventor/client/boxes/PaletteBox.java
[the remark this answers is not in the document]
SusanRatiLane: OK, this should be resolved.

appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
```diff
@@ -802,8 +824,13 @@ public interface OdeMessages extends Messages, AutogeneratedOdeMessages {
...
   String userAdminMenuItem();
 
+  @DefaultMessage("Advanced")
+  @Description("Label of the button leading to admin functionality")
+  String advancedMenuItem();
+
+
   //Tabs
-  @DefaultMessage("My Projects")
+  @DefaultMessage("Projects")
```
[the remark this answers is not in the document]
SusanRatiLane: > We should double check this change. In particular, I think this is used for the button in the top right and we do not want to have this be called Projects in addition to the menu on the left side.

The very next string reads "My projects." Is there a reason not to use that one when we want that string?

appinventor/appengine/src/com/google/appinventor/client/actions/NewAction.java
```diff
@@ -0,0 +1,15 @@
+package com.google.appinventor.client.actions;
```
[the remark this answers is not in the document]
SusanRatiLane: This is not in use. It appears to be a duplicate of NewProjectAction. Deleting.

appinventor/appengine/src/com/google/appinventor/client/style/neo/NewYoungAndroidProjectWizardNeo.ui.xml
SusanRatiLane: > Maybe put this into documentation instead rather than leaving vestigial code that was never used in the source tree.

Yeah, this probably isn't the best way to handle it. This is just such a good example to use for future work when we actually do want to do something with a dialog that requires a template. We should discuss where it should go in documentation so that it's easily discoverable.

appinventor/appengine/src/com/google/appinventor/client/style/neo/ImagesNeo.java
SusanRatiLane: > I don't think this is used any more.

Since it wasn't replaced in Neo, I removed it from there. I don't know if this PR is the place to remove it from Classic.

appinventor/appengine/src/com/google/appinventor/client/style/neo/images/button.png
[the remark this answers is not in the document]
SusanRatiLane: > I wonder if it makes sense to add an arrow cursor to this to differentiate it from a rectangle component.

To my surprise, Material Icons doesn't have what I consider a great button icon. I replaced it with their "button_alt" icon, which is a rectangle with a + on it. I don't know what it is an alternative to, since there doesn't seem to be an icon named "button." I wonder if that exists in an older version someplace.

appinventor/appengine/src/com/google/appinventor/client/style/neo/images/arduino.png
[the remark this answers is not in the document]
SusanRatiLane: > Not sure about this one since Arduino has a logo and this removes a lot of connection to that.

Replaced with actual Arduino logo in a style that matches.

appinventor/appengine/src/com/google/appinventor/client/explorer/commands/RenameProjectsCommand.ui.xml
```diff
@@ -0,0 +1,21 @@
+<!-- RenameProjectsCommand.ui.xml -->
```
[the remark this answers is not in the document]
SusanRatiLane: > This comment seems redundant. Instead, replace it with the copyright information as seen in other ui.xml files.
> 
> Also, it would be more appropriate to call this RenameProjectsDialog since it defines a dialog.

This is a remnant from Vishwas's prototype. I recall that renaming projects is actually difficult to do in any way other than just copying the whole project, so he never implemented it. Removing.

appinventor/appengine/src/com/google/appinventor/client/style/neo/Ode.ui.xml
SusanRatiLane: > Consider moving width/height settings to CSS.

To my surprise, I can't find a definition for ode-ProjectListView anywhere. It's used in both Ode.ui.xml files, and I'm sure it contained something at one time. Both layouts display fine without reference to this class. I'm removing it.

appinventor/appengine/src/com/google/appinventor/client/style/neo/Ode.ui.xml
SusanRatiLane: > Maybe consider moving the width setting to CSS.

There's no css class for PaletteBox or the individual work columns. I'm reluctant to create a class just to put the width in it.

appinventor/appengine/src/com/google/appinventor/client/style/neo/Ode.ui.xml
SusanRatiLane: > Consider moving the width setting to the CSS.

Same issue with no obvious class. You may have a different vision for the css. If so we should probably talk about it.

appinventor/appengine/src/com/google/appinventor/client/actions/ImportComponentAction.java
```diff
@@ -0,0 +1,11 @@
+package com.google.appinventor.client.actions;
```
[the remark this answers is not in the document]
SusanRatiLane: > Add the appropriate preamble here.

Does not appear to be used. Removing.

appinventor/appengine/src/com/google/appinventor/client/UIStyleFactory.java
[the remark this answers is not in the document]
SusanRatiLane: > Google Java Style naming conventions would have this class name as `UiStyleFactory`.

Since User Interface is an acronym, this has always looked weird to me. But no biggie.

appinventor/appengine/src/com/google/appinventor/client/UIStyleFactory.java
```diff
@@ -0,0 +1,47 @@
...
+import com.google.appinventor.client.explorer.folder.ProjectFolder;
+import com.google.appinventor.client.explorer.youngandroid.ProjectList;
+import com.google.appinventor.client.wizards.youngandroid.NewYoungAndroidProjectWizard;
+import com.google.gwt.core.client.GWT;
+import com.google.gwt.json.client.JSONObject;
+import com.google.gwt.uibinder.client.UiBinder;
+import com.google.gwt.uibinder.client.UiTemplate;
+import com.google.gwt.user.client.ui.FlowPanel;
+
+public class UIStyleFactory {
```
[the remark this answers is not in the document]
SusanRatiLane: > It seems to me that it would be better for this to be an `interface` and then have two implementations for classic and modern. The current setup with subclassing in the modern version overrides every single thing about this implementation.

The rationale here was that I wanted the base class to be the default for all additional styles. As we add features, the likelihood increases that styles will lag behind. I even had in mind the possibility that styles could be added that we are not responsible for maintaining. This way, if a method is not implemented in a given style, the system will fall back to Classic.

I also did the Images resource for Neo this way.

appinventor/appengine/src/com/google/appinventor/client/UIStyleFactory.java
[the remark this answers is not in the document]
mark-friedman: I agree with you Susan and tried to get this changed at Google, to no avail 😞

appinventor/appengine/src/com/google/appinventor/client/UIStyleFactory.java
```diff
@@ -0,0 +1,47 @@
...
+
+  @UiTemplate("Ode.ui.xml")
+  public interface OdeUiBinder extends UiBinder<FlowPanel, Ode> {}
+  @UiTemplate("style/GSoC/Ode.ui.xml")
+  interface OdeUiBinderGSoC extends UiBinder<FlowPanel, Ode> {}
```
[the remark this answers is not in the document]
SusanRatiLane: > Also, couldn't the GSOC version be defined in the UiFactoryGSOC?

Defining the Neo version of Ode in UiFactoryNeo does not compile. There is an issue with two templates for the same class Ode in two different locations. I don't remember all the details, but I tried several things and spent time on StackOverflow looking for solutions before setting on this one. 

I just now refreshed my memory by trying a couple of things which did not compile.

So this is not ideal, and there may well be a better way to handle it. We can discuss it if we have resources. This isn't ideal, but it does work.

ewpatton:

appinventor/appengine/src/com/google/appinventor/client/TopToolbar.java
```diff
@@ -182,7 +186,7 @@ public void updateMenuState(int numSelectedProjects, int numProjects) {
         : MESSAGES.exportProjectMenuItem();
     fileDropDown.setItemHtmlById(WIDGET_NAME_EXPORTPROJECT, exportProjectLabel);
     fileDropDown.setItemEnabledById(WIDGET_NAME_EXPORTPROJECT, allowExport);
-    fileDropDown.setItemEnabled(MESSAGES.exportAllProjectsMenuItem(), allowExportAll);
+    fileDropDown.setItemEnabledById(MESSAGES.exportAllProjectsMenuItem(), allowExportAll);
```
ewpatton: setItemEnabled works on the text of the item, whereas setItemEnabledById works on the name of the item. It's not clear to me why this was changed, and it may break in non-English locales.

appinventor/appengine/src/com/google/appinventor/client/TopToolbar.java
```diff
@@ -296,13 +300,16 @@ public void updateFileMenuButtons(int view) {
...
+      fileDropDown.setItemVisible(MESSAGES.deleteFromTrashButton(), false);
       fileDropDown.setItemEnabled(MESSAGES.trashProjectMenuItem(), projectCount == 0);
-      fileDropDown.setItemEnabled(MESSAGES.exportAllProjectsMenuItem(), projectCount > 0);
-      fileDropDown.setItemEnabledById(WIDGET_NAME_EXPORTPROJECT, false);
-      fileDropDown.setItemEnabled(MESSAGES.saveMenuItem(), false);
-      fileDropDown.setItemEnabled(MESSAGES.saveAsMenuItem(), false);
+      fileDropDown.setItemEnabledById(MESSAGES.exportAllProjectsMenuItem(), projectCount > 0);
+      fileDropDown.setItemEnabled(WIDGET_NAME_EXPORTPROJECT, false);
+      fileDropDown.setItemEnabledById(MESSAGES.saveMenuItem(), false);
+      fileDropDown.setItemEnabledById(MESSAGES.saveAsMenuItem(), false);
```
ewpatton, line 312: See my previous comment about setItemEnabled vs setItemEnabledById

appinventor/appengine/src/com/google/appinventor/client/TopToolbar.java
```diff
@@ -312,6 +319,9 @@ public void updateFileMenuButtons(int view) {
         buildDropDown.setItemEnabled(MESSAGES.showExportAndroidAab2(), false);
       }
     } else { // We have to be in the Designer/Blocks view
+      if (fileDropDown.getName() == "ProjectDesignOnly") {
```
ewpatton, line 322: Technically we should use `.equals` here for correct Java semantics since `==` compares for identity. We just get lucky that GWT does the wrong thing when converting to JavaScript.

appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/ProjectToolbar.java
```diff
@@ -35,10 +35,12 @@ interface ProjectToolbarUiBinder extends UiBinder<Toolbar, ProjectToolbar> {}
   private static final String WIDGET_NAME_SENDTONG = "Send to Gallery";
   private static final String WIDGET_NAME_LOGINTOGALLERY = "Login to Gallery";
 
-  private final boolean isReadOnly;
+  public final boolean isReadOnly;
+  public final boolean galleryEnabled;
+  @UiField public Label projectLabel;
+  @UiField public Label trashLabel;
```
ewpatton, line 41: It seems like these should be `protected`, and the flags could even be `private` since they are only accessed from within this class.

appinventor/appengine/src/com/google/appinventor/client/style/neo/ProjectListNeo.ui.xml
```diff
@@ -0,0 +1,43 @@
...
+             xmlns:g="urn:import:com.google.gwt.user.client.ui"
+             xmlns:ai="urn:import:com.google.appinventor.client.components"
+             ui:generatedFormat="com.google.gwt.i18n.server.PropertyCatalogFactory"
+             ui:generatedKeys="com.google.gwt.i18n.server.keygen.MethodNameKeyGenerator"
+             ui:generateLocales="default">
+  <ui:with field="messages" type="com.google.appinventor.client.OdeMessages"/>
+  <g:FlowPanel styleName='ode-ProjectTable'>
+    <g:FlowPanel styleName='ode-ProjectHeaderRow'>
+      <g:CheckBox ui:field='selectAllCheckBox' styleName="ode-ProjectCheckBox gwt-CheckBox"/>
+      <g:FlowPanel styleName="ode-ProjectHeaderLabel ode-ProjectListName">
```
ewpatton, line 14: In the future we may want to make this into a proper widget rather than repeating the structure across both styles and multiple times within a style.

appinventor/appengine/war/static/css/Ya.css
```diff
@@ -3145,15 +3256,7 @@ div.dropdiv p {
...
 
-.download-barcode:before {
-    content: "";
-    background-color: #888;
-    position: absolute;
-    width: 1px;
-    height: 160px;
-    top: 40px;
-    left: 50%;
-    display: block;
```
ewpatton, line 3156: What is the justification for removing this?

ewpatton (changes requested):
Mostly there, minus the button fix in Neo.

appinventor/appengine/war/static/css/Ya.css
SusanRatiLane: Perhaps this would be better as a call for help. I've found the before syntax for css to be very difficult to work with in UIBinder. I eventually rage quit, since I don't see that this vertical line brings much to the dialog anyway.

appinventor/appengine/src/com/google/appinventor/client/style/neo/neo.css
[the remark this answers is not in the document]
SusanRatiLane: > We should review the use of !important in this file as it is generally considered bad practice.

There are places where upstream css needs to be refactored. This was particularly true for the download QR code dialog, where !important seemed to be the expected way to override things that were added in code. I didn't look to hard at how to refactor it in this pass.

appinventor/appengine/src/com/google/appinventor/client/variableColors.css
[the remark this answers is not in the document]
SusanRatiLane: > It's not clear to me what the purpose of this file is. It seems like the intention was to use CSS vars, but it doesn't actually do that. The result is that it injects a lot of non-working CSS into the page, which the browser then ignores. Removing the reference and recompiling didn't seem to break anything.

Hmm. It sounds like this might be an incomplete part of Jennet's dark mode work. Removing the reference seems like the right thing to do at this stage.

appinventor/appengine/src/com/google/appinventor/client/TopToolbar.java
```diff
@@ -172,7 +176,7 @@ public void updateMoveToTrash(boolean moveToTrash) {
   }
 
   public void updateMenuState(int numSelectedProjects, int numProjects) {
-    boolean allowDelete = hasWriteAccess && numSelectedProjects > 0;
+    boolean allowDelete = readOnly && numSelectedProjects > 0;
```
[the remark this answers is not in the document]
SusanRatiLane: > This seems to flip the semantics.

Trying to think this through. You can only delete something if you have write access and have selected something, so it certainly looks like the original is correct.

However, I can use booleans in UIBinder, but I can't negate them in the template. I'll have to double-check the behavior.

ewpatton (approved):

ewpatton:
There are still a bunch of "todos" here but at this point I don't think it is worth holding up the PR any further. I've addressed a few low hanging fruits and the rest we can revisit in another update.

PR #3136, opened 2024-03-07
add keydown events

jennet-zamanova:
the key press is recognized only in case you are writing in the textbox. the validator to check the project name was added


AppInventorWorkerBee:

ewpatton:

SusanRatiLane:
@jennet-zamanova The functionality seems to work, but a bunch of space has been introduced after the project name.

This:

Screenshot 2024-03-08 at 4 09 24 PM

Became this:

Screenshot 2024-03-08 at 4 09 15 PM

jennet-zamanova:
yes, that's because the setValidator() for labeledTextBox sets the height of validator to automatically be 85px when there is no error. Would you prefer it to be set to 0?

SusanRatiLane (approved):

PR #3142, opened 2024-03-08
Standardization of all the icon sizes to 16px

Venkata-Sai-Vishwanath-robo:
The added code standardizes all icon sizes to 16px, relieving extension makers from concerns about their extension's icon size.

The problem that this PR solves is that a few extension icons have larger sizes than 16x16px and this PR standardizes all the icons across the app inventor to 16px.

Example of this issue :
LargeIcon


AppInventorWorkerBee:

PR #3159, opened 2024-03-31
Expand dropdown palette

jennet-zamanova:
make a collapsable pallete, so you can close all component drawers


AppInventorWorkerBee:

PR #3170, opened 2024-04-18
Make YoungAndroidPalettePanel a singleton

SusanRatiLane:
Performance improvement by @ewpatton:

Creating a YaFormEditor or a YaBlocksEditor takes roughly between 250-300 ms on a relatievly powerful laptop. Most of this time is spent in constructing the YoungAndroidPalettePanel (which the YaBlocksEditor doesn't even make use of). This change makes it so that there is a single YoungAndroidPalettePanel that uses a model-view-controller approach to render the current editor's component database rather than constructing an instance per-editor. This reduces the creation of editors down to around 10 ms or less.

Change-Id: I8cd301384afd673ff870664dcb1d5698e200e9ea


SusanRatiLane:
Closed in favor of newer updated branch.

PR #3178, opened 2024-05-15
Call FileUploadWizard.show()

jisqyv:
Make sure we call the show() method each time we construct a FileUploadWizard

Change-Id: I324c75e63a4071a710dd01b93868141b5255ecf7


jisqyv:

SusanRatiLane (approved):
This checks out, but it's actually a fix for the switchable ui rather than master, so that needs to be merged first or this needs to be pointed at my switchable ui branch.

PR #3184, opened 2024-05-20
Fix IndeterminateColor of LinearProgress not being changed.

m-anshuman2166:
Fix indeterminate color not being changed, which was caused due to a typo.


AppInventorWorkerBee:

PR #3185, opened 2024-05-24
Deprecate the Edit Image block without a mask

jisqyv:
ChatGPT does not support image editing without a mask. Using this block simply results in an error being generated from the ChatBot proxy.

We have marked it as deprecated, so projects that have it in them will still load properly, but the block will be disabled. The block will also not appear in the blocks drawer.

If in the future ChatGPT, or another provider supports image editing without requiring a mask, we can "un" deprecate it.

Change-Id: I7cd625ecf68b1db43b90488f385489cb8acbcd02


jisqyv:
Note: I intentionally based this change on master, even though it is a change to a component file. However, the impact is really on the Web UI, and is safe for a non-component release.

ewpatton (approved):

PR #3186, opened 2024-05-26
Update README.ios.md

petersmythe:
Add requirement to install iOS SDK (but the minimum version of 17.5 is mostly likely wrong and needs to be adjusted)

### DIFF
--- a/README.ios.md
+++ b/README.ios.md
@@ -2,7 +2,7 @@
 
 ## Prerequisites (Mac OS X)
 
-* Xcode 14 or higher
+* Xcode 14 or higher (and install platform iOS 17.5 SDK or higher)
 * macOS 12 or higher
 
 (optional) To automatically install the companion app on a connected, authorized iOS device, install ideviceinstaller via Homebrew:


AppInventorWorkerBee:

petersmythe:
* [ ] TODO: what capabilities must be specified for the App?

image

README.ios.md
```diff
@@ -2,7 +2,7 @@
 
 ## Prerequisites (Mac OS X)
 
-* Xcode 14 or higher
+* Xcode 14 or higher (and install platform iOS 17.5 SDK or higher)
```
ewpatton, line 5: We have a minimum SDK of 12.0 for iOS. Specifying to install the latest iOS SDK should be sufficient.

README.md
```diff
@@ -76,10 +76,16 @@ building on versions below the ones stated. To install on a device,
 you **must** have a valid Apple Developer account license and have
 added the relevant mobile provisioning profiles from the Developer
 portal to your Xcode organizer (see Apple's website on instructions on
-how to do this).
+how to do this).  The profile should have these **exact** details, and 
+should include a distribution certificate:
```
ewpatton, line 80: Distribution profiles are different from developer profiles. Unless someone is publishing to the App Store, they shouldn't need to create a distribution profile.

README.md
```diff
@@ -76,10 +76,16 @@ building on versions below the ones stated. To install on a device,
...
+
+```conf
+Name:	MIT App Inventor Development
+App ID: edu.mit.appinventor.*
```
ewpatton, line 84: We do not want people using MIT in the identifiers for their testing. They should be using something a namespace they might reasonably control. Worst case, they could use a reverse DNS style entry based on their email address or their GitHub username.

petersmythe:
Noted, I will fix

README.md
petersmythe: This does not work, in my (very limited) experience.  The file [`AICompanionApp.xcodeproj/project.pbxproj`](https://github.com/mit-cml/appinventor-sources/blob/df152767e9a083536656337297909cbc96a4d34d/appinventor/AICompanionApp.xcodeproj/project.pbxproj#L813) has `PRODUCT_BUNDLE_IDENTIFIER = edu.mit.appinventor.aicompanion3` and this is the App ID that must be included in the provisioning profile, otherwise Xcode reports:

image

and the `ant ios` build fails with:

image

Alternatively, we could include instructions to update AICompanionApp.xcodeproj/project.pbxproj in multiple places to a namespace that they control. What do you think @ewpatton ?

README.md
petersmythe: Documentation updated.

PR #3189, opened 2024-06-13
Fix disappearing extensions category in collapsable palette

SusanRatiLane:
This should fix the extensions category on the component palette disappearing when the palette is refreshed. I also did some refactoring of the collapsable palette code.


ewpatton (approved):
LGTM

PR #3195, opened 2024-07-07
Fix Mark Origin

jisqyv:
Provide an appropriate error message when MarkOrigin is called on an ImageSprite prior to having its image set.

Change-Id: Ia212a5cac8d925bed08fa6a79257ebc37c99f9b1


jisqyv:
Fixes #3194

jisqyv:

SusanRatiLane (approved):
LGTM

PR #3196, opened 2024-07-08
Make project list scrollable in neo

SusanRatiLane:
This fixes the bug reported by the ed team that the project list does not scroll in Neo. 

It also makes the status panel css conform to the dynamic color list for dark mode support.


jisqyv (approved):
LGTM

PR #3197, opened 2024-07-11
Allow style switch in read only, fix Neo buttons

SusanRatiLane:
Miscellaneous minor fixes:

1. Settings menu was set to invisible when the project is read only, but the UI Settings dialog now needs to be accessed there to switch between styles. Fixed so that the Settings menu always appears and the menu items are restricted by write access.

2. Fixed tutorial and gallery-related toolbar buttons that were not displaying with their icons properly in Neo.


jisqyv (approved):
LGTM

PR #3205, opened 2024-07-17
Set designer sidebars to fixed width in Neo

SusanRatiLane:
Large numbers of non-visible components on a screen and some browser resizing was causing the designer sidebars to shrink smaller than their specified widths. This css change should fix that.


jisqyv (approved):
LGTM

PR #3206, opened 2024-07-20
Add font properties to Canvas

gordonlu310:
A small PR that adds the `FontBold`, `FontItalic` and `FontTypeface` properties to the Canvas component, along with relevant documentation.

**Note:** Custom TTFs didn't load in the Companion on my computer (didn't work for other components like Labels), but everything else runs like butter. Would be extremely grateful if someone can test this out for me.


AppInventorWorkerBee:

PR #3209, opened 2024-07-24
Fix typos in logging messages

logresearch:
fix a misuse capitalization in logging statement


AppInventorWorkerBee:

PR #3255, opened 2024-11-04
Add recyclerview.aar to support additional APIs for extension development

jewelshkjony:
**Description:**  
This pull request adds `recyclerview.aar` to fix a `java.lang.NoClassDefFoundError` for `androidx/recyclerview/R$dimen`, which occurs when building extensions that use extra `recyclerview` APIs. The in-built ListView component works fine without this, but `recyclerview.aar` is necessary for extension developers to access all features without crashes. This update ensures compatibility and stability for those building with advanced `recyclerview` functionality.

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/ListView.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/ListView.java
@@ -67,7 +67,7 @@ import java.util.List;
     nonVisible = false,
     iconName = "images/listView.png")
 @SimpleObject
-@UsesLibraries(libraries ="recyclerview.jar, cardview.jar, cardview.aar")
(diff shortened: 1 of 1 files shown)


AppInventorWorkerBee:

jewelshkjony:
@AppInventorWorkerBee, it's ready to review.

SusanRatiLane:

ewpatton:
A couple of things here. First, this should be against ucr since it's adding content to the companion app (otherwise, extensions can just include the .aar in their own `@UsesLibraries` annotation). Second, if we decide to increase the size of the companion app by including the .aar, then we probably should just add this change to #3235. Third, what APIs are you using to trigger the actual error?

jewelshkjony:
The ItemTouchHelper class is throwing the `androidx/recyclerview.R$dimen` missing error.

image

ewpatton:
This one line change was added in #3235.

PR #3281, opened 2024-12-08
Fix margins in ListView

patryk84a:
General items:

- [x] My code follows the:

- [x] I branched from `ucr`

For all other changes:

What does this PR accomplish?

*Description*

There was no implementation for setting margins in the horizontal layout of ListView. Therefore, the horizontal layout displayed margins incorrectly, adjusting them as for the vertical layout.


AppInventorWorkerBee:

patryk84a:
Why do I have 3 additional commits in this PR?

- I synced my fork's ucr branch with mit ucr
- I created a fix_margins branch from my fork ucr
- I made changes to the file and committed them to my fork's fix_margins
- then I created a PR

Did something go wrong or is this how it should be?

ewpatton:

ewpatton:
@patryk84a It looks like because you used the Github-based merge (which creates a merge commit by default), your ucr branch got a bit out of sync with ours. Don't worry about this as we will squash merge this PR into a single commit once approved.

patryk84a:
What do I need to do to revert my fork's ucr branch to its original state for future reference? Because now it shows me it's in sync.

ewpatton:
@patryk84a Assuming you have `origin` set up to be your fork and `upstream` set up to be mit-cml's repo (this one):

```shell
git checkout ucr  # switch to your local ucr branch
git fetch upstream  # gets the latest MIT version
git reset --hard upstream/ucr  # makes your local ucr point to the same commit as our ucr
git push -f origin ucr  # force-push to reset the fork to match MIT and local
```

appinventor/components/src/com/google/appinventor/components/runtime/ListView.java
```diff
@@ -1361,14 +1360,22 @@ public void getItemOffsets(Rect outRect, View view, RecyclerView parent, Recycle
...
-        outRect.left = margins - column * margins / spanCount;
-        outRect.right = (column + 1) * margins / spanCount;
-        if (position < spanCount || first) {
-          first = false;
-          outRect.top = margins;
+        if (orientation == ComponentConstants.LAYOUT_ORIENTATION_HORIZONTAL) {
+          if (position == 0) {
+            outRect.set(margins, margins, margins, margins);
+          } else {
+            outRect.set(0, margins, margins, margins);
```
ewpatton, line 1367: So one issue with doing this here is that it seems to make the elements unbalanced. Here are two screen shots with the first and last elements of the list in horizontal layout mode. Note how the last element is flush with the edge whereas the first element isn't.

Screenshot_20241213-121310

Screenshot_20241213-121314

patryk84a: Yes, I focused on margins, not paying attention to element widths. To fix this we need to set the element width. I also added a screen orientation listener to update widths after screen rotation. I also simplified the code for setting parenthesis for vertical layout, removing lines of code that I had tested with Grid layout before.

patryk84a:
Now is good:

Screenshot_2024-12-14-18-18-08-543_appinventor ai_test listview_margins
Screenshot_2024-12-14-18-18-35-109_appinventor ai_test listview_margins
Screenshot_2024-12-14-18-18-54-638_appinventor ai_test listview_margins

SusanRatiLane:
The behavior in production is really the intended behavior. The intent is to have a swipe left behavior like an album view.

What we really need is a new layout image, main text, and detail text stacked vertically to support a very large image. I put this on a GSoC project idea, but now I'm not sure if it's listed as an issue. If you are interested in continuing work on the ListView, that layout would be very appreciated.

patryk84a:
> @patryk84a So I don't understand what's changed, but in the production app the elements still take up the full width of the screen (less margins) but with this change the elements are just as wide as their text. I'm concerned that people relying on the current sizing won't be pleased with this change. Can you figure out what is causing the elements to no longer be the the same size?
> 
> Before: Screenshot_20250313-145948
> 
> After: Screenshot_20250313-150042

I need to look at this again because I had a completely different result.

patryk84a:
> The behavior in production is really the intended behavior. The intent is to have a swipe left behavior like an album view.
> 
> What we really need is a new layout of image, main text, and detail text stacked vertically to support a very large image. I put this on a GSoC project idea, but now I'm not sure if it's listed as an issue. If you are interested in continuing work on the ListView, that layout would be very appreciated.

Yes, I think this element layout is a good idea. Another good idea would be to add a grid layout. Often users use grid layout to display images.

patryk84a:
> new layout of image, main text, and detail text stacked vertically 

@SusanRatiLane Such a layout name already exists, it would be useful to organize the layout names. I understand that the image and texts are to be aligned to the center of the element? Both texts under the image, or maybe Main above the image and Detail below the image?

patryk84a:
> > > new layout of image, main text, and detail text stacked vertically
> > 
> > 
> > @SusanRatiLane Such a layout name already exists, it would be useful to organize the layout names. I understand that the image and texts are to be aligned to the center of the element? Both texts under the image, or maybe Main above the image and Detail below the image?
> 
> You're right that the name exists, but the "Image, Main Text, Detail Text (Vertical) layout" is a small thumbnail image to the right and the Main and Detail texts stacked vertically. The layout we need is all three elements stacked vertically so that the image can be very large and the text elements can be captions. Considering the name of that existing layout, it's not entirely clear what the name of the new layout should be. Perhaps it should just be called "Card View."

I almost have a ready solution, I still have to work on ListViewMock. I changed the name of the old layout where the image and both texts are to "Image, MainText, DetailText", and I named the new layout "Image, MainText, DetailText (Vertical)". I also plan to rename the ListViewLayout property to ElementLayout, while the name ListViewLayout will be registered for changing the layout of the entire ListView, i.e. Linear or Grid. This can be combined with Orientation, i.e. ListViewLayout would have the options: LinearVertical, LinearHorizontal, GridVertical, GridHorizontal.

SusanRatiLane:
> > > > new layout of image, main text, and detail text stacked vertically
> > > 
> > > 
> > > @SusanRatiLane Such a layout name already exists, it would be useful to organize the layout names. I understand that the image and texts are to be aligned to the center of the element? Both texts under the image, or maybe Main above the image and Detail below the image?
> > 
> > 
> > You're right that the name exists, but the "Image, Main Text, Detail Text (Vertical) layout" is a small thumbnail image to the right and the Main and Detail texts stacked vertically. The layout we need is all three elements stacked vertically so that the image can be very large and the text elements can be captions. Considering the name of that existing layout, it's not entirely clear what the name of the new layout should be. Perhaps it should just be called "Card View."
> 
> I almost have a ready solution, I still have to work on ListViewMock. I changed the name of the old layout where the image and both texts are to "Image, MainText, DetailText", and I named the new layout "Image, MainText, DetailText (Vertical)". I also plan to rename the ListViewLayout property to ElementLayout, while the name ListViewLayout will be registered for changing the layout of the entire ListView, i.e. Linear or Grid. This can be combined with Orientation, i.e. ListViewLayout would have the options: LinearVertical, LinearHorizontal, GridVertical, GridHorizontal.

Could you talk more about your vision for grid layouts? I've never seen a selection list be laid out in a grid.

Also, please don't give the new layout the same name as an old one. That is going to cause confusion for users.

patryk84a:
> For this PR I'd just like the width issue fixed so we can merge it. Other ListView features should be kept to another PR.

Yes, I have a new branch for the new layout. I'm just waiting for this patch to merge so I can update my branch and release a PR.

> Could you talk more about your vision for grid layouts? I've never seen a selection list be laid out in a grid.

I would use the version where all the elements are the same size:



> Also, please don't give the new layout the same name as an old one. That is going to cause confusion for users.

It wouldn't affect old projects because I would map everything. In new projects the user can see what layout they are setting for the list view, both in companion and in the phone view in the designer. I think it's similar to changing from TextSize to FontSize. The name "Card View" doesn't really reflect the layout itself, because we get the card view after setting other parameters like margins, corners, etc. The card view can also be another layout.

SusanRatiLane:
> I would use the version where all the elements are the same size

This is a good description. Could you put this in a new issue so that we can discuss it separately from this PR?The large-image/album-view/etc layout should be a separate PR from grid layouts. 

For this PR, I think the horizontal orientation change needs to be rolled back. Then we can merge this PR.

patryk84a:
> For this PR, I think the horizontal orientation change needs to be rolled back. Then we can merge this PR.

I don't really understand what rolled back has to be.

SusanRatiLane:
> > For this PR, I think the horizontal orientation change needs to be rolled back. Then we can merge this PR.
> 
> I don't really understand what rolled back has to be.

We don't want the "After" image in Evan's comment here: https://github.com/mit-cml/appinventor-sources/pull/3281#issuecomment-2722416070

The correct behavior is for the element to take up the entire screen width on horizontal orientation. Anything else should be a new orientation as you outlined in the comment thread here.

patryk84a:
Oh, yeah, the last patch fixed that.

SusanRatiLane (approved):
LGTM

PR #3348, opened 2025-01-22
Conditionally save project.properties

jisqyv:
We were always saving project.properties every time we loaded a project, even when there are no changes. This is because the computed settings don't store the qualifiedName (package name, called "main") as well as the project name itself. We add them here before the comparison.

Change-Id: I6b0e206dbef6bfc100c5c3ce89db1ca7395e17e0

General items:

- [ ] My code follows the:
- [x] `ant tests` passes on my machine (except the Blockly Tests, which don't run on my machine, yet.)

- [ ] I branched from `ucr`

For all other changes:

What does this PR accomplish?

Reduces the I/O between the client and server by not saving project.properties when it really hasn't changed.

*Description*

Fixes # .

Resolves # .


mark-friedman (approved):
LGTM

ewpatton (approved):

PR #3353, opened 2025-01-23
Issue 3329 and minor commits

cindyloo:
For all other changes:
- [x] I forked from `master`
- [x] I branched from my `master`bramcj
- [x] My pull request has my fork as the base

What does this PR accomplish?

[issue 3329 - don't load from URL repo parameter if the UI needs to be chosen

[marginRight instead of margin-right]

[handle assertion error in devmode]


ewpatton (changes requested):
I still need to test it locally but here are a few minor things.

ewpatton (approved):
LGTM

PR #3358, opened 2025-01-26
 Fixed Folder Name Validation Bug in New Folder Wizard

ONKARBH:
Resolved an issue in the "New Folder Wizard" where folder name validation was incorrectly referencing "project" instead of "folder." The bug also prevented automatic replacement of invalid characters (such as spaces being replaced by underscores), which was supposed to occur during validation, mirroring the behavior for project names. Now, folder name validation properly handles invalid characters, and error messages correctly reference "folder" as intended.


AppInventorWorkerBee:

Issue #3345, opened 2025-01-22
New folder dialog does not handle name validation correctly


SusanRatiLane:
**Describe the bug**

The new folder wizard uses the project name validation methods and displays the related error messages. In addition to the error messages referencing "project" instead of "folder," they also don't correctly describe the wizard behavior. The folder wizard does not automatically remove or replace invalid characters, e.g. replacing " " with "_", and the error message says it will.

**Affects**

- [ ] Designer
- [ ] Blocks editor
- [X] Projects Explorer
- [ ] Companion
- [ ] Compiled apps
- [ ] Buildserver
- [ ] Debugging
- [ ] Other... (please describe)

**Expected behavior**

Folder name validation should handle automatically handle invalid characters the same way project name validation does, and the strings referencing projects should be fixed.

**Steps to reproduce**

Click the new folder button on the project list and try some names.

d1vyanshu-kumar:
@SusanRatiLane If no one is currently working on this issue, could you please assign it to me?

SusanRatiLane:
> [@SusanRatiLane](https://github.com/SusanRatiLane) If no one is currently working on this issue, could you please assign it to me?

Done. Thanks for taking it on.

d1vyanshu-kumar:
Hi @SusanRatiLane,

I need your help with an issue I encountered.

Currently, when entering an invalid project name with spaces between letters, no alert message is triggered. However, for folder names, an alert does appear. The folder wizard does not replace spaces with underscores ("_"), unlike expected behaviour.

can you please verify this approach that I am on the right track or not?
Remove the alert for invalid folder names.
Instead of prompting the user to enter a valid folder name, automatically replace spaces with underscores if the user clicks "OK" with an invalid folder name, similar to the project configuration.
Let me know your thoughts. Thanks!

PR #3361, opened 2025-01-29
"Fix folder name validation to replace invalid characters"

ONKARBH:
This PR fixes the issue in the "New Folder Wizard" where invalid characters in folder names were not being automatically replaced (e.g., spaces not being converted to underscores). Additionally, the error messages incorrectly referenced "project" instead of "folder," which has now been corrected.

Changes made:
- Updated folder name validation logic to replace invalid characters like spaces with underscores.
- Corrected error messages to refer to "folder" instead of "project."

This ensures that the folder naming process is consistent and user-friendly, improving the overall experience when creating folders.

Fixes issue: #3345


AppInventorWorkerBee:

PR #3376, opened 2025-02-13
ios companion app fix for troubleshooting, minor tweaks to buttons

cindyloo:
For all other changes:

- [ x] I branched from `master`

What does this PR accomplish?
 please review #3374 here, but basically
on iOS, there are 3 screens before the final "scan/code" screen that are to inform the user of what s/he should do. However, many don't read and just click through to the scan/code screen and then don't know what to do. Also, if they Deny access to the camera or to the wifi, they will not be able to connect at all. We can help them by giving them a Need Help button and informing them what to do to remedy their situation. (rather than deleting the app and starting over) 

basic modifications: 
swap blue for green scan button, add graphic, move and resize current buttons. Remove large dark grey instruction graphic. add Need Help? button that shows ip/version info and help in popup


appinventor/aicompanionapp/src/Base.lproj/Main.storyboard
```diff
@@ -152,10 +186,15 @@
     </scenes>
     <resources>
         <image name="Instructions.png" width="800" height="200"/>
-        <image name="connectwCode.png" width="300" height="49"/>
-        <image name="connectwQR.png" width="300" height="49"/>
+        <image name="Mooning" width="400" height="400"/>
```
cindyloo, line 189: this is actually not in the codebase any longer. Beyond perplexed why it is showing up

cindyloo:
should it be code.appinventor.edu or ai2.appinventor.edu? (in onboarding images vs Need help troubleshooting msgs)

Issue #3365, opened 2025-02-04
Hidden Components Checkbox not working


SusanRatiLane:
**Describe the bug**

The checkbox at the top of the Viewer panel on the project designer does not seem to be operating in either Neo or Classic.

**Affects**

- [X] Designer
- [ ] Blocks editor
- [ ] Projects Explorer
- [ ] Companion
- [ ] Compiled apps
- [ ] Buildserver
- [ ] Debugging
- [ ] Other... (please describe)

**Expected behavior**

When the checkbox is checked, any components for which VISIBLE=false should become visible on the Viewer panel.

**Steps to reproduce**

Create a project.
Add a component
In Properties, set the VISIBLE property to false
See that the component is no longer visible.
Check the Hidden Components Checkbox. This is the checkbox by the eye icon in Neo and labeled "Display hidden components in Viewer" in Classic.
The component should become visible in the Viewer, but it does not.

d1vyanshu-kumar:
@SusanRatiLane, I followed the steps you mentioned in the issue, but I wasn't able to reproduce the problem. When I set a component's **VISIBLE** property to `false` and then check the **"Display hidden components"** checkbox, the hidden component appears as expected in the Viewer panel. Let me know if there's any specific scenario I might be missing!

https://github.com/user-attachments/assets/8feb6425-3dd9-4e7e-96bf-9c2f80fe1832

SusanRatiLane:
> [@SusanRatiLane](https://github.com/SusanRatiLane), I followed the steps you mentioned in the issue, but I wasn't able to reproduce the problem. When I set a component's **VISIBLE** property to `false` and then check the **"Display hidden components"** checkbox, the hidden component appears as expected in the Viewer panel. Let me know if there's any specific scenario I might be missing!

OK, I was responding to a forum post, and when I tested against ai2, it did not work. The report was specifically about non-Screen1 screens, but when I tested, Screen1 didn't work either. I wonder if there is something more complicated needed to trigger this bug. Could you try some testing on non-Screen1 screens? I will do the same and see if I can find something that can be reproduced consistently.

d1vyanshu-kumar:
@SusanRatiLane, could you please help me understand why this method is only called for Screen1 while it never gets triggered for other screens? I’m trying to figure out what might be causing this behaviour.

[...]

PR #3378, opened 2025-02-16
 Need Help: Hidden Components Checkbox not working #3365

d1vyanshu-kumar:
From the logs:

- The toggleHiddenComponents() method is triggered successfully when you click the checkbox.
- The checkbox state is updated (Screen1 = false is logged).
- Components are being processed for visibility updates, but they are still not appearing in the UI.

the issues i get:

- Hidden components are not appear in the viewer when showHidden = true.
- Warnings like "An internal error has occurred. Report a bug?" appear.
- The checkbox state is stored correctly, but it does not affect component visibility.

issue number: #3365

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockComponent.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockComponent.java
@@ -15,6 +15,7 @@ import com.google.appinventor.client.Ode;
 import com.google.appinventor.client.boxes.SourceStructureBox;
 import com.google.appinventor.client.editor.simple.SimpleEditor;
 import com.google.appinventor.client.editor.simple.components.utils.PropertiesUtil;
+import com.google.appinventor.client.editor.youngandroid.HiddenComponentsCheckbox;
 import com.google.appinventor.client.editor.youngandroid.YaBlocksEditor;
 import com.google.appinventor.client.editor.youngandroid.YaFormEditor;
 import com.google.appinventor.client.explorer.SourceStructureExplorerItem;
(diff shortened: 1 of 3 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

d1vyanshu-kumar:
### 🔍 Summary of Changes & Thought Process to Fix the Issue

#### Problem Overview
Initially, the checkbox only worked for `Screen1`, but not for other screens. Even when clicking the checkbox, hidden components did not appear.

---

Making the Checkbox Work for All Screens
**Problem:**
- The original code stored only one `MockForm` instance (`MockForm form;`).
- When switching screens, it forgot the new screen, so the checkbox only worked for `Screen1`.

**Fix:**
- Replaced the single `MockForm` instance with a `Map<String, MockForm>` to track multiple screens.
- Added a `String currentFormTitle` variable to track the current active screen.



---

Storing & Retrieving Checkbox State Per Screen
**Problem:**
- The checkbox did not remember if it was checked or unchecked when switching screens.
- There was no mechanism to track the checkbox state for different screens.

**Fix:**
- Added a `Map<String, Boolean> screenCheckboxState` to store the checkbox state per screen.
- Added the method:
  ```java
  public static boolean getCheckboxStateForScreen(String formTitle) {
      return screenCheckboxState.getOrDefault(formTitle, false);
  }
  ```



---

Updating Component Visibility Dynamically
**Problem:**
- Even when clicking the checkbox, hidden components did not appear.
- The code only collected hidden components but did not update their visibility dynamically.

**Fix:**
- Introduced `updateVisibility()` method inside `MockComponent` to set visibility dynamically based on checkbox state.
- Updated `getHiddenVisibleChildren()` to check `showHidden` and update component visibility.




 Modified getHiddenVisibleChildren()**
**Problem:**
- The old method only collected hidden components but didn’t modify their visibility dynamically.
- It didn’t check the checkbox state (`showHidden`), which meant hidden components were not shown when the checkbox was checked.

**Solution:**
- Checked `showHidden` state to determine whether to show or hide the component dynamically.
- Refactored the loop to update visibility on the spot.

 Modified doRefresh()**
**Problem:**
- The "Internal Error Has Occurred" message was happening because `doRefresh()` wasn’t properly refreshing the UI.
- Some components were not updating their layout, causing the error.

**Solution:**
- Improved `doRefresh()` to handle layout updates correctly and prevent null errors.

PR #3420, opened 2025-03-18
Fix MockListView in Neo

patryk84a:
General items:

- [ ] My code follows the:

- [ ] I branched from `ucr`

For all other changes:

What does this PR accomplish?

I tested everything mainly in the classic interface, but while testing the Neo interface I noticed inconsistencies in look and feel. I realized that Neo uses different css and I need to update it according to ya.css.

Fixes # .

Resolves # .

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/style/neo/neo.css
+++ b/appinventor/appengine/src/com/google/appinventor/client/style/neo/neo.css
@@ -1113,15 +1113,52 @@ they get cut off in Windows.
   }
 
   /* class for the ListView Mock component */
-  .listViewComponentStyle {
-    width: 315px;
-    display: block;
-    overflow: unset;
-  }
+.listViewComponentStyle {
+  width: 315px;
+  display: flex;
+  flex-direction: column;
+  overflow: hidden;
+}
 
-  .listViewItemStyle {
-    border-bottom: 1px solid gray;
-  }
+.listViewComponentStyle .listViewFilterStyle {
+  flex: 0 0 30px;
+  width: calc(100%-6px);
+  border-radius: 0px;
+  margin-bottom: 0px;
+  padding: 0px 2px;
+}
+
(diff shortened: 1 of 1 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

SusanRatiLane:

appinventor/appengine/src/com/google/appinventor/client/style/neo/neo.css
```diff
@@ -1113,15 +1113,52 @@ they get cut off in Windows.
...
+  width: 315px;
+  display: flex;
+  flex-direction: column;
+  overflow: hidden;
+}
 
-  .listViewItemStyle {
-    border-bottom: 1px solid gray;
-  }
+.listViewComponentStyle .listViewFilterStyle {
```
SusanRatiLane, line 1123: I'm sure this isn't obvious, but neo.css actually loads on top of ya.css. Any class that is identical in both UI styles should be in ya.css and not in neo.css.

SusanRatiLane (changes requested):

appinventor/appengine/src/com/google/appinventor/client/style/neo/neo.css
patryk84a: I don't know what you mean but without adding these styles to neo.css, the ListView view is not correct. For example, the items have a fixed bottom border which is set in neo.css.

patryk84a: I think it might work if we completely removed the ListView style from neo.css, I'll have to try that. I don't understand why there are duplicate styles for views in neo.css, copied from ya.css. If even in Neo's interface styles from ya.css are taken into account, then I don't think we need duplicate styles.

SusanRatiLane (approved):
This tests out for me.

PR #3429, opened 2025-03-20
Slider property name fix

SusanRatiLane:
Quick fix for #3152. This just changes the property name ColorThumb to ThumbColor to match the equivalent property in the Switch component.


patryk84a:
Could you please add the new Slider features to ai2-test before the weekend, so the community has more time to test it with different projects? We have seen many times that community testing can uncover more non-obvious issues.

PR #3476, opened 2025-06-09
Fix hanging builds due to malformed keystores

barreeeiroo:
General items:

- [x] My code follows the:

- [ ] I branched from `ucr`

For all other changes:

What does this PR accomplish?

*Description*

* Prevents a potential Denial Of Service when using customized keystores; see below.
* Prevents any subprocess from requesting stdin, and hanging around without exiting.
* Ensures all subprocesses will terminate within 5 minutes.

App Inventor uses keystores to sign apps, more specifically `apksigner` for APKs and `jarsigner` for AABs. AI2 relies on the keystores on having a password as `android` in order to decrypt and read the private key inside them. However, apparently there is a two-level password in keystores: one to unlock the container, and another one to unlock the `androidkey` which AI2 uses.  
Some keystores do have the container password as `android`, but the actual key as something else (probably generated using Android Studio or just changing the password of the AI2 one). This causes `apksigner` and `jarsigner` to change to interactive mode, and prompts for a password. As such, the process keeps running indefinitely, and never halts. It doesn't even complain, because the first layer password is correct, but prompts for the second one. As such, AI2 thinks that `apksigner` / `jarsigner` is still running, although they are waiting for some input from stdin, and never kills them. And finally, as a result, this build process consumes "one build slot", and if it gets filled up, no new builds are allowed. And in case of unlimited builds, this might lead to the buildserver running out of space, as the temporary resources for each build request would never get cleaned up.

The actual documentation states that there is one extra parameter to prevent this behaviour:

```text
--key-pass            Password with which the private key is protected.
                      The following formats are supported:
                          pass:<password> password provided inline
                          env:<name>      password provided in the named
                                          environment variable
                          file:<file>     password provided in the named
                                          file, as a single line
                          stdin           password provided on standard input,
                                          as a single line
                      If --key-pass is not specified for a KeyStore key, this
                      tool will attempt to load the key using the KeyStore
                      password and, if that fails, will prompt for key password
                      and attempt to load the key using that password.
                      If --key-pass is not specified for a private key file key,
                      this tool will prompt for key password only if a password
                      is required.
                      When the same file (including standard input) is used for
                      providing multiple passwords, the passwords are read from
                      the file one line at a time. Passwords are read in the
                      order in which signers are specified and, within each
                      signer, KeyStore password is read before the key password
                      is read.
```

Note it says "it will prompt for key password if using the same password as the keystore fails". When preparing a fix for the DoS, it was considered to introduce those flags into `apksigner` and `jarsigner`. However, as of now, anyone using a keystore with `android` as password and no password for the actual key works as intended; introducing those parameters may break that behaviour and invalidate their own keystores. Considering that, it seemed like a safer non-breaking change approach was to just completely disable stdin.

The preconditions for the DoS to apply are:
* Buildservers must be limiting number of builds based on maximum number of concurrent ones.
* Keystores must use a password other than `android` for the `androidkey.`
* _Unless this is some version-specific behaviour, `apksigner` and `jarsigner` must yield to IO in case of incorrect password._

### DIFF
--- a/appinventor/buildserver/src/com/google/appinventor/buildserver/Main.java
+++ b/appinventor/buildserver/src/com/google/appinventor/buildserver/Main.java
@@ -14,6 +14,7 @@ import java.io.IOException;
 import java.util.logging.Logger;
 import java.util.zip.ZipFile;
 
+import com.google.appinventor.buildserver.util.Execution;
 import org.kohsuke.args4j.CmdLineException;
 import org.kohsuke.args4j.CmdLineParser;
 import org.kohsuke.args4j.Option;
@@ -115,6 +116,12 @@ public final class Main {
       LOG.severe("Problem opening inout zip file: " + commandLineOptions.inputZipFile.getName());
       System.exit(1);
     }
+
+    if (commandLineOptions.isForCompanion) {
+      // If we are building for the Companion, disable timeouts
+      Execution.disableTimeouts();
+    }
+
     Result result = projectBuilder.build(commandLineOptions.userName,
                                          zip,
                                          commandLineOptions.outputDir,
--- a/appinventor/buildserver/src/com/google/appinventor/buildserver/Main.java
+++ b/appinventor/buildserver/src/com/google/appinventor/buildserver/Main.java
@@ -6,6 +6,7 @@
 
 package com.google.appinventor.buildserver;
 
+import com.google.appinventor.buildserver.util.Execution;
 import com.google.appinventor.buildserver.stats.NullStatReporter;
 import com.google.appinventor.buildserver.tasks.android.AndroidBuildFactory;
 
@@ -14,7 +15,6 @@ import java.io.IOException;
 import java.util.logging.Logger;
 import java.util.zip.ZipFile;
 
-import com.google.appinventor.buildserver.util.Execution;
 import org.kohsuke.args4j.CmdLineException;
 import org.kohsuke.args4j.CmdLineParser;
 import org.kohsuke.args4j.Option;
--- a/appinventor/buildserver/src/com/google/appinventor/buildserver/tasks/android/RunApkSigner.java
+++ b/appinventor/buildserver/src/com/google/appinventor/buildserver/tasks/android/RunApkSigner.java
@@ -39,7 +39,9 @@ public class RunApkSigner implements AndroidTask {
 
     if (!Execution.execute(null, apksignerCommandLine,
         System.out, System.err)) {
-      TaskResult.generateError("Error while running ZipAligned tool");
+      context.getReporter().warn("We could not sign your app. In case you are using a custom keystore, please make " +
+        "sure its password is set to 'android', and the key is set to 'androidkey'.");
+      return TaskResult.generateError("Error while running ApkSigner tool");
     }
 
     return TaskResult.generateSuccess();
--- a/appinventor/buildserver/src/com/google/appinventor/buildserver/tasks/android/RunApkSigner.java
+++ b/appinventor/buildserver/src/com/google/appinventor/buildserver/tasks/android/RunApkSigner.java
@@ -33,6 +33,7 @@ public class RunApkSigner implements AndroidTask {
         "-ks", context.getKeystoreFilePath(),
         "-ks-key-alias", "AndroidKey",
         "-ks-pass", "pass:android",
+        // "-key-pass", "pass:android",
         context.getPaths().getDeployFile().getAbsolutePath()
     };
 
--- a/appinventor/buildserver/src/com/google/appinventor/buildserver/tasks/android/RunApkSigner.java
+++ b/appinventor/buildserver/src/com/google/appinventor/buildserver/tasks/android/RunApkSigner.java
@@ -33,7 +33,6 @@ public class RunApkSigner implements AndroidTask {
         "-ks", context.getKeystoreFilePath(),
         "-ks-key-alias", "AndroidKey",
         "-ks-pass", "pass:android",
-        // "-key-pass", "pass:android",
         context.getPaths().getDeployFile().getAbsolutePath()
     };
 
--- a/appinventor/buildserver/src/com/google/appinventor/buildserver/tasks/android/RunApkSigner.java
+++ b/appinventor/buildserver/src/com/google/appinventor/buildserver/tasks/android/RunApkSigner.java
@@ -38,7 +38,7 @@ public class RunApkSigner implements AndroidTask {
     };
 
     if (!Execution.execute(null, apksignerCommandLine,
-        System.out, System.err)) {
+        System.out, System.err, Execution.Timeout.SHORT)) {
       context.getReporter().warn("We could not sign your app. In case you are using a custom keystore, please make " +
         "sure its password is set to 'android', and the key is set to 'androidkey'.");
       return TaskResult.generateError("Error while running ApkSigner tool");
--- a/appinventor/buildserver/src/com/google/appinventor/buildserver/tasks/android/RunBundletool.java
+++ b/appinventor/buildserver/src/com/google/appinventor/buildserver/tasks/android/RunBundletool.java
@@ -237,7 +237,7 @@ public class RunBundletool implements AndroidTask {
     String[] bundletoolBuildCommandLine = bundletoolCommandLine.toArray(new String[0]);
 
     return Execution.execute(null, bundletoolBuildCommandLine,
-        System.out, System.err);
+        System.out, System.err, Execution.Timeout.LONG);
   }
 
   private boolean jarsigner(CompilerContext<AndroidPaths> context) {
@@ -265,6 +265,6 @@ public class RunBundletool implements AndroidTask {
     String[] jarsignerSignCommandLine = jarsignerCommandLine.toArray(new String[0]);
 
     return Execution.execute(null, jarsignerSignCommandLine,
-        System.out, System.err);
+        System.out, System.err, Execution.Timeout.SHORT);
   }
 }
--- a/appinventor/buildserver/src/com/google/appinventor/buildserver/tasks/android/RunBundletool.java
+++ b/appinventor/buildserver/src/com/google/appinventor/buildserver/tasks/android/RunBundletool.java
@@ -100,6 +100,8 @@ public class RunBundletool implements AndroidTask {
 
     context.getReporter().info("Signing bundle");
     if (!jarsigner(context)) {
+      context.getReporter().warn("We could not sign your app. In case you are using a custom keystore, please make " +
+        "sure its password is set to 'android', and the key is set to 'androidkey'.");
       return TaskResult.generateError("Could not sign bundle");
     }
     return TaskResult.generateSuccess();
--- a/appinventor/buildserver/src/com/google/appinventor/buildserver/tasks/android/RunBundletool.java
+++ b/appinventor/buildserver/src/com/google/appinventor/buildserver/tasks/android/RunBundletool.java
@@ -258,8 +258,6 @@ public class RunBundletool implements AndroidTask {
     jarsignerCommandLine.add(context.getKeystoreFilePath());
     jarsignerCommandLine.add("-storepass");
     jarsignerCommandLine.add("android");
-    // jarsignerCommandLine.add("-keypass");
-    // jarsignerCommandLine.add("android");
     jarsignerCommandLine.add(context.getPaths().getDeployFile().getAbsolutePath());
     jarsignerCommandLine.add("AndroidKey");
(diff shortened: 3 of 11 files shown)


appinventor/buildserver/src/com/google/appinventor/buildserver/tasks/android/RunApkSigner.java
```diff
@@ -33,12 +33,15 @@ public TaskResult execute(AndroidCompilerContext context) {
         "-ks", context.getKeystoreFilePath(),
         "-ks-key-alias", "AndroidKey",
         "-ks-pass", "pass:android",
+        // "-key-pass", "pass:android",
```
ewpatton: Is this needed?

appinventor/buildserver/src/com/google/appinventor/buildserver/tasks/android/RunApkSigner.java
```diff
@@ -33,12 +33,15 @@ public TaskResult execute(AndroidCompilerContext context) {
...
         context.getPaths().getDeployFile().getAbsolutePath()
     };
 
     if (!Execution.execute(null, apksignerCommandLine,
         System.out, System.err)) {
-      TaskResult.generateError("Error while running ZipAligned tool");
+      context.getReporter().warn("We could not sign your app. In case you are using a custom keystore, please make " +
+        "sure its password is set to 'android', and the key is set to 'androidkey'.");
+      return TaskResult.generateError("Error while running ApkSigner tool");
```
ewpatton, line 43: I think this is okay for now but for critical errors we need to think about how to internationalize them.

appinventor/buildserver/src/com/google/appinventor/buildserver/tasks/android/RunBundletool.java
```diff
@@ -256,6 +258,8 @@ private boolean jarsigner(CompilerContext<AndroidPaths> context) {
     jarsignerCommandLine.add(context.getKeystoreFilePath());
     jarsignerCommandLine.add("-storepass");
     jarsignerCommandLine.add("android");
+    // jarsignerCommandLine.add("-keypass");
+    // jarsignerCommandLine.add("android");
```
ewpatton: Consider removing unless we tie this into the changes in RunApkSigner.

appinventor/buildserver/src/com/google/appinventor/buildserver/util/Execution.java
```diff
@@ -29,6 +30,8 @@ public final class Execution {
   private static final Logger LOG = Logger.getLogger(Execution.class.getName());
   private static final Joiner joiner = Joiner.on(" ");
 
+  private static final int SUBPROCESS_TIMEOUT_SECONDS = 5 * 60; // 5 minutes
```
ewpatton: I think a build is composed of something like 10 tasks, so if we wait 5 minutes for each a failing build could take almost an hour.

appinventor/buildserver/src/com/google/appinventor/buildserver/util/Execution.java
```diff
@@ -132,10 +145,17 @@ public static int execute(File workingDir, String[] command, StringBuffer out,
...
+    // Prevent any interactive shell from waiting for input
+    process.getOutputStream().close();
     Thread outThread = new RedirectStreamToStringBuffer(out, process.getInputStream());
     Thread errThread = new RedirectStreamToStringBuffer(err, process.getErrorStream());
+
     try {
-      process.waitFor();
+      if (!process.waitFor(SUBPROCESS_TIMEOUT_SECONDS, TimeUnit.SECONDS)) {
+        process.destroyForcibly();
+        err.append("Process had to be forcibly terminated due to timeout\n");
```
ewpatton: Shouldn't we return here?

ewpatton (changes requested):

appinventor/buildserver/src/com/google/appinventor/buildserver/tasks/android/RunApkSigner.java
barreeeiroo: So, I left this commented out based on the PR description:

> When preparing a fix for the DoS, it was considered to introduce those flags into apksigner and jarsigner. However, as of now, anyone using a keystore with android as password and no password for the actual key works as intended; introducing those parameters may break that behaviour and invalidate their own keystores. Considering that, it seemed like a safer non-breaking change approach was to just completely disable stdin.

The fix was either to pass this flag, or to disable stdin. I'm fine to remove the comment completely, but I think it's useful to leave the comment, so in case anyone has the same idea, they can do `git blame` and figure out from the PR the reason why it's not added.

appinventor/buildserver/src/com/google/appinventor/buildserver/tasks/android/RunBundletool.java
barreeeiroo: See my comment above from RunApkSigner

appinventor/buildserver/src/com/google/appinventor/buildserver/util/Execution.java
barreeeiroo: Yeah, but things like D8 will definitely take more time than apksigner. Even D8 alone should take around 1-2 minutes maximum. The only reason why I added that much buffer is in case something goes really wrong, and we don't want to terminate too early the processes.

Unless we have data on what's the longest D8 can take on a legitimate, I wouldn't be confident setting a value lower than this. Also, this value varies depending on the CPU specs, hence if using a mix of different CPUs, each one may behave differently.  
Again, the reason why setting this to 5 minutes is as a mere guardrail preventing any hanging build in case we missed out anything else, not as a "global build timeout".

appinventor/buildserver/src/com/google/appinventor/buildserver/util/Execution.java
barreeeiroo: We have to first join the threads below, no? Otherwise they will keep running forever. Basically, it waits, forces stop, joins threads, and finally it returns.

appinventor/buildserver/src/com/google/appinventor/buildserver/util/Execution.java
barreeeiroo: Is it worth adding a global build timeout though? I can actually submit that change as a different PR. With this change, I'm just focusing on terminating all processes within 5 minutes at maximum, so if the buildserver is using a "not that good CPU", and D8 has a lot to do, it leaves buffer to not force exit by mistake on the first build.

ewpatton: Most tasks are quick but some we know are slow. We may want to have a general timeout of 1 or 2 seconds and certain tasks like d8 can override it to a 1 minute or so. We also would want to make sure that any constraint would not prevent the companion from compiling when in CI or as part of a release cycle.

appinventor/buildserver/src/com/google/appinventor/buildserver/util/Execution.java
ewpatton: Yes I think you are right but this feels really unintuitive to me. If the process has failed and then is killed, the IO streams should also close and the threads should terminate. If that's not the case it would be good to understand why.

barreeeiroo: Tbh I was thinking the same, I actually first had a return in case of timeout, but then realized about the hanging threads. In the end, the final result is the same, just some weird code structure...

I just didn't have time to investigate and see what happens with those threads. I'll check later along with your comments on the timeouts above.

barreeeiroo: On a different thought, this function is not used anymore. I guess I'll just remove it 🙂

ewpatton:
Always a good option!Sent from my iPhoneOn Jun 9, 2025, at 16:59, Diego Barreiro Perez ***@***.***> wrote:﻿
@barreeeiroo commented on this pull request.



In appinventor/buildserver/src/com/google/appinventor/buildserver/util/Execution.java:
>      try {
-      process.waitFor();
+      if (!process.waitFor(SUBPROCESS_TIMEOUT_SECONDS, TimeUnit.SECONDS)) {
+        process.destroyForcibly();
+        err.append("Process had to be forcibly terminated due to timeout\n");

On a different thought, this function is not used anymore. I guess I'll just remove it 🙂

appinventor/buildserver/src/com/google/appinventor/buildserver/util/Execution.java
barreeeiroo: I have just made this change. I have defined 3 different tiers of task: short, medium and long. For long ones, now there is a timeout of 2 minutes; for medium, it's 30 seconds; and for short, 5 seconds. To help understand this, there is an enum defining each "task type", and then multiple overloads; default is short, but all the code now explicitly defines the type of task.

* Long Tasks: Dex, D8, BundleTool (yes, it actually takes quite some time)
* Medium Tasks: Kawa, Aapt, Aapt2
* Short Tasks: everything else

I would like to get a review on this configuration though. I'm afraid of setting a too-low value for some task, and users start getting timeouts. The risky part are extensions, that a project with "too many extensions" may now suffer timeouts. Of course, we can dynamically calculate the timeout for each task, but is it really needed?

appinventor/buildserver/src/com/google/appinventor/buildserver/util/Execution.java
barreeeiroo: And it's gone now

appinventor/buildserver/src/com/google/appinventor/buildserver/util/Execution.java
barreeeiroo: > Is it worth adding a global build timeout though?

I have also added a global build timeout of 5 minutes. It should be more than enough, but inside the ProjectBuilder, the compiler thread now exits if the build did not complete within 5 minutes. Maybe it's too high, but I think it serves its purpose.

appinventor/buildserver/src/com/google/appinventor/buildserver/util/Execution.java
```diff
@@ -97,50 +91,68 @@ private Execution() {
...
-    	}
+      for(int i =0; i < command.length; i++){
+        command[i] = command[i].replace("\"", "\\\"");
+      }
     }
+
     try {
       Process process = Runtime.getRuntime().exec(command, null, workingDir);
+      // Prevent any interactive shell from waiting for input
+      process.getOutputStream().close();
```
sanandrea, line 109: why is this needed?

barreeeiroo: If anyone uploaded a "malformed" keystore (having `android` as keystore password but something else as private key password), both `apksigner` and `jarsigner` will not exit but enter interactive mode, prompting for input and waiting for user action. As such, all the builds will get stuck.

Disabling stdin completely from any subprocess will instantly terminate anything asking for a user prompt, hence preventing that issue. From some local tests, apksigner complains by throwing an exception saying that the standard input is closed. So, it solved the issue of stuck builds.

appinventor/buildserver/src/com/google/appinventor/buildserver/tasks/android/RunApkSigner.java
ewpatton: I agree with disabling stdin. I'm not sure it's worth keeping these around as we also have the discussion here in the PR about why this was decided.

barreeeiroo: Aligned, just removed it. My thought was to leave a comment explicitly in the code, so in case anyone goes and tries to fix it sees it commented out and can do the git blame; otherwise it would be kind of an "implicit" check, and we will have to remember in case we see a PR.

appinventor/buildserver/src/com/google/appinventor/buildserver/Main.java
```diff
@@ -14,6 +14,7 @@
 import java.util.logging.Logger;
 import java.util.zip.ZipFile;
 
+import com.google.appinventor.buildserver.util.Execution;
```
ewpatton: This doesn't seem sorted in the right place.

appinventor/buildserver/src/com/google/appinventor/buildserver/util/Execution.java
ewpatton: In App Store uploads "[5 minutes] should be more than  enough" may not be true, but I can adjust that in the iOS build PR.

ewpatton (approved):
LGTM

PR #3534, opened 2025-07-31
Implement Remote Storage outside of GCP

barreeeiroo:
### Background

App Inventor currently uses Google Cloud Storage as the main storage solution for "large" objects, which is mainly build outputs. These large objects are built outside of Google App Engine, and sent back to get them stored in GCS. Then, GAE will proxy them to the user to download it.

This however implies quite a lot of network hops, and increases egress traffic costs: from the buildserver to GAE (if there is egress network charge), from GAE to GCS to get it stored, from GCS back to GAE when downloading it, and from GAE to the user. This last hop is the most expensive, as it charges $0.139 per GB transferred to the Internet.

With that in mind, it makes sense to allow the users to download the files directly, instead of going through GAE (which is also taking resources to load in memory that data).

### Description

This PR implements support for "Remote Storage" in App Inventor. It allows GAE to generate Presigned URLs in a 3P storage provider, and interact with files out there. It is mainly targeted to store build outputs outside of GAE, as these are the largest artifacts eating up the majority of the egress traffic.

This PR also removes the "build size" limitation. By directly letting the user download the output file from storage, we are no longer bound to GAE's request/response payload limits, so it can effectively "download" larger projects.

### Current Capabilities

* GAE has now support to configure "remote storage", which can be interacted using presigned URLs.
* Buildserver can now, optionally, receive an extra argument: `uploadUrl`. If provided, the output file (APK or AAB) gets uploaded there, instead of sent back to GAE.
    * It will still send back to GAE the ZIP file with the full structure, but the output file will be truncated to 0. This is needed to "acknowledge" the build, and make sure the target file in the remote storage can be identified.
* Implementation can be cloud-provider agnostic without any libraries: as of now, only S3 is implemented. However, this is S3-compatible and not vendor locked. Any S3 compatible storage provider can be used, as long as they support presigned URLs. Some examples are:
    * Cloud: MinIO, Blackblaze, Cloudflare R2, Hetzner
    * Self Hosted: MinIO, SeaweedFS

### Pending Items

* [ ] Allow sending exported projects (AIA) to remote storage.
* [ ] Implement GCS support, to avoid leaving Google Cloud.

### Testing

* Presigned URL generation works as intended.
* Locally built using remote storage some apps.
* Tested with both AWS S3 buckets ("native" S3), and Hetzer Object Storage buckets (S3 compatible)

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/server/BuildOutputServlet.java
+++ b/appinventor/appengine/src/com/google/appinventor/server/BuildOutputServlet.java
@@ -1,29 +1,30 @@
 // -*- mode: java; c-basic-offset: 2; -*-
 // Copyright 2009-2011 Google, All Rights reserved
-// Copyright 2011-2019 MIT, All rights reserved
+// Copyright 2011-2025 MIT, All rights reserved
 // Released under the Apache License, Version 2.0
 // http://www.apache.org/licenses/LICENSE-2.0
 
 package com.google.appinventor.server;
 
-import com.google.appinventor.common.utils.StringUtils;
 import com.google.appinventor.server.storage.StorageIo;
 import com.google.appinventor.server.storage.StorageIoInstanceHolder;
+import com.google.appinventor.server.storage.remote.RemoteStorage;
+import com.google.appinventor.server.storage.remote.RemoteStorageInstanceHolder;
+import com.google.appinventor.shared.util.BuildOutputFiles;
 import com.google.appinventor.server.util.CacheHeaders;
 import com.google.appinventor.server.util.CacheHeadersImpl;
 import com.google.appinventor.shared.rpc.Nonce;
-import com.google.appinventor.shared.rpc.ServerLayout;
-import com.google.appinventor.shared.rpc.project.ProjectSourceZip;
 import com.google.appinventor.shared.rpc.project.RawFile;
 import com.google.appinventor.shared.storage.StorageUtil;
 
 import javax.servlet.ServletOutputStream;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.Date;
-import java.util.NoSuchElementException;
+import java.util.List;
 import java.util.logging.Logger;
 
 /**
@@ -68,7 +69,6 @@ public class BuildOutputServlet extends OdeServlet {
 
     RawFile downloadableFile;
 
-    String userId = null;
     String nonceValue = null;
 
     try {
@@ -95,6 +95,37 @@ public class BuildOutputServlet extends OdeServlet {
         resp.sendError(resp.SC_NOT_FOUND, "Link has timed out");
         return;
       }
+
+      final String userId = nonce.getUserId();
+      final long projectId = nonce.getProjectId();
+
+      if (RemoteStorageInstanceHolder.isRemoteConfigured()) {
+        final StorageIo storageIo = StorageIoInstanceHolder.getInstance();
+        // Given all the output files, try to find the one that matches an output extension
+        final List<String> files = storageIo.getProjectOutputFiles(userId, projectId);
+        String outputFileName = null;
+        for (String fileName : files) {
+          if (BuildOutputFiles.isOutputFile(fileName)) {
+            outputFileName = new File(fileName).getName();
+            break;
+          }
+        }
+
+        if (outputFileName == null) {
+          throw new FileNotFoundException("No target file found!");
+        }
+
+        final RemoteStorage remoteStorage = RemoteStorageInstanceHolder.getInstance();
+        // For now, we use Android always. If we eventually support iOS, we should decide here
+        //   whether we are going for Android or Apple, as well as in the BarcodeAction.
+        final String target = BuildOutputFiles.getTargetName();
+        final String objectKey = remoteStorage.getBuildOutputObjectKey(target, userId, projectId, outputFileName);
+        final String remoteUrl = remoteStorage.generateRetrieveUrl(objectKey);
+
+        resp.sendRedirect(remoteUrl);
+        return;
+      }
+
       downloadableFile = fileExporter.exportProjectOutputFile(nonce.getUserId(), nonce.getProjectId(), null);
 
     } catch (FileNotFoundException e) {
--- a/appinventor/appengine/src/com/google/appinventor/server/storage/remote/RemoteStorage.java
+++ b/appinventor/appengine/src/com/google/appinventor/server/storage/remote/RemoteStorage.java
@@ -1,0 +1,85 @@
+// -*- mode: java; c-basic-offset: 2; -*-
+// Copyright 2025-2025 MIT, All rights reserved
+// Released under the Apache License, Version 2.0
+// http://www.apache.org/licenses/LICENSE-2.0
+
+package com.google.appinventor.server.storage.remote;
+
+/**
+ * Interface which abstracts the remote storage access patterns.
+ * This allows AI2 to store different files, like build outputs, outside
+ *   GCP, and provide a presigned URL to download and/or access them.
+ */
+public abstract class RemoteStorage {
+  // We partition the bucket into different usage types, so specific lifecycle rules could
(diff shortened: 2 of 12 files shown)


barreeeiroo:
This PR is ready for review. I know it's a long one, and probably open to a lot of discussion. I will leave some comments regarding minor possible changes.

The main major callout I do have is regarding adding support for uploading AIAs through Remote Storage, as it would be more complicated than I initially thought, as that is client side.  
I would have to create a new servlet to generate the presigned URL, and modify the client to first retrieve the URL, then use that URL for the multipart upload, and then call back the upload servlet with an extra flag so it retrieves the file from remote storage rather than expecting it as a multipart upload.  
An alternative that users have right now, considering they can now export AIAs larger than 32MB, is to temporarily replace the assets with 0-size ones, and then upload them individually again.

The code is partially ready, https://github.com/barreeeiroo/appinventor-sources/compare/external-build-artifacts...support-remote-uploads, but how multipart upload work with presigned URLs is a bit different than "standalone" PUT. New fields have to be added to the form, as well as computing an extra policy, rather than a presigned URL.  
Due to this, the upload form would become more complex, and I think it's not going to work the same way as for GCS...

So, I left it undone to also discuss on this.

appinventor/appengine/src/com/google/appinventor/server/storage/remote/RemoteStorage.java
```diff
@@ -0,0 +1,116 @@
...
+   * Generates a constant object key for a given specific project export.
+   *
+   * @param downloadKind the type of download
+   * @param userId the user ID owning the project
+   * @param fileName the name of the file to store
+   * @return export/userId/downloadKind/timestampHash/file.name
+   */
+  public final String getProjectExportObjectKey(final String downloadKind, final String userId, final String fileName) {
+    final String timestamp = String.valueOf(Instant.now().getEpochSecond());
+    final String timestampHash = generateFieldsHash(userId, downloadKind, fileName, timestamp);
```
barreeeiroo, line 84: I'm introducing this as part of the object key to ensure we don't overwrite previous exports (in case the user still has the previous URL).

We can either keep it as it is, or sync it with how it's done for the build output, and just overwrite it. The previous presigned URL will still be valid, but will "point" to the new object already.

appinventor/appengine/src/com/google/appinventor/server/storage/remote/RemoteStorageInstanceHolder.java
```diff
@@ -0,0 +1,83 @@
...
+        return RemoteStorageProviderGCS.getInstance();
+      } catch (UnsupportedOperationException e) {
+        LOG.severe("Could not initialize Remote GCP Storage in non-Production environment!");
+        return null;
+      }
+    }
+
+    if (providerName.equals("s3")) {
+      return RemoteStorageProviderS3.getInstance();
+    }
```
barreeeiroo, line 79: I could technically have another map and prevent initializing the same provider twice for different usages, but that would be an unlikely situation. The point of allowing to customize the provider is to save costs, and GCP is "available by default"; the point is to set S3 for build outputs.

appinventor/appengine/src/com/google/appinventor/server/BuildOutputServlet.java
```diff
@@ -95,6 +95,37 @@ public void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOExc
...
+        if (outputFileName == null) {
+          throw new FileNotFoundException("No target file found!");
+        }
+
+        final RemoteStorage remoteStorage = RemoteStorageInstanceHolder.getInstance(RemoteStorageInstanceHolder.Usage.BUILD);
+        // For now, we use Android always. If we eventually support iOS, we should decide here
+        //   whether we are going for Android or Apple, as well as in the BarcodeAction.
+        final String target = BuildOutputFiles.getTargetName();
+        final String objectKey = remoteStorage.getBuildOutputObjectKey(target, userId, projectId, outputFileName);
+        final String remoteUrl = remoteStorage.generateRetrieveUrl(objectKey);
```
barreeeiroo, line 123: Another option here would be to modify the Nonce, and along with the user and project, store a pre-calculated presigned URL, valid for as long as the Nonce is valid. This would remove generating in runtime the URL, and "make the URL consistent" if the same Nonce is accessed twice.

The only reason why I went with this approach to generate the URL on the fly is to not modify the current datastore schema, and as this is a local operation, it should be fine to spend a few ms more here. Accessing the same nonce twice is an unlikely situation too imo.

appinventor/appengine/src/com/google/appinventor/server/DownloadServlet.java
```diff
@@ -94,6 +99,21 @@ public class DownloadServlet extends OdeServlet {
...
 
+  // If any file we try to export exceeds 20MB, then use the remote storage solution if
+  //   configured.
+  // If unconfigured, still use GAE, but it may fail for large files due to response
+  //   payload limit.
+  private static final int DIRECT_DOWNLOAD_MAX_FILE_SIZE = 20_000_000;
+
+  // Only use remote downloads for specific kind of downloads (and avoid using for other
+  //   files like assets, even if "larger", although impossible).
+  private static final Set<String> REMOTE_DOWNLOAD_KINDS = Set.of(
```
barreeeiroo, line 110: I didn't have this initially, and I was just proxying all the files through Remote Storage if the size was larger. However, the purpose here is to just allow exporting projects that exceed 32MB, so I decided to limit to project-related download kinds.

appinventor/appengine/src/com/google/appinventor/server/DownloadServlet.java
```diff
@@ -268,6 +289,15 @@ public void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOExc
       } else {
         throw new IllegalArgumentException("Unknown download kind: " + downloadKind);
       }
+
+      if (!isInline) {
+        final String downloadObjectUrl = shouldUseRemoteStorageDownload(downloadKind, userId, downloadableFile);
```
barreeeiroo, line 296: I couldn't really find where the inline mode is used, but I think it's safe to assume that if something should render "inline", it's better to not use foreign origins or anything similar...

appinventor/appengine/src/com/google/appinventor/server/DownloadServlet.java
```diff
@@ -94,6 +99,21 @@ public class DownloadServlet extends OdeServlet {
   private static final int USERFILE_PATH_INDEX = 4;
   private static final int SPLIT_LIMIT_USERFILE = 5;
 
+  // If any file we try to export exceeds 20MB, then use the remote storage solution if
+  //   configured.
+  // If unconfigured, still use GAE, but it may fail for large files due to response
+  //   payload limit.
+  private static final int DIRECT_DOWNLOAD_MAX_FILE_SIZE = 20_000_000;
```
barreeeiroo, line 106: Technically GAE payload limit is 32MB, but wanted to have some buffer. Maybe it's too much, should this be set to 30MB?

If using GCP actually, we also save costs. GAE sends the AIA to GCS, and user downloads from GCS. Egress traffic from GCS to Internet is much cheaper than using GAE, so there's also a point in storing larger files in GCS.

barreeeiroo:
I actually forgot that some providers, like MinIO, do not support virtual host endpoints, and use path like ones. I have now added support for both.

Tested by running MinIO locally with these flags:

```xml

    <!-- Remote Storage Provider -->
    <property name="remotestorage.build" value="s3" />
    <property name="remotestorage.export" value="s3" />

    <!-- GCP Remote Storage Configuration -->
    <property name="remotestorage.gcp.bucketname" value="" />

    <!-- S3 Compatible Remote Storage Configuration -->
    <property name="remotestorage.s3.protocol" value="http" />
    <property name="remotestorage.s3.endpoint" value="localhost:9000" />
    <property name="remotestorage.s3.pathlike" value="true" />
    <property name="remotestorage.s3.bucketname" value="remotestorage" />
    <property name="remotestorage.s3.bucketregion" value="us-east-1" />
    <property name="remotestorage.s3.accesskeyid" value="minioadmin" />
    <property name="remotestorage.s3.secretaccesskey" value="minioadmin123" />
```

image

PR #3552, opened 2025-08-09
Fix Temporary Files & Cleanup Storage

barreeeiroo:
Change File Role Type to Temporary, and create file prefix constant. Clean up dead code in Storage, and apply some linting suggestions.

General items:

- [x] My code follows the:

- [ ] I branched from `ucr`

For all other changes:

What does this PR accomplish?

*Description*

Standardizes the `__TEMP__` prefix for storage, and ensures those files get the proper role.

It also cleans up the Storage package, removing unused code, and applying linting rules. I was basically working on a POC to replace all GCP services with AWS ones, and got annoyed by the yellow hints in IntelliJ.

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/server/storage/ObjectifyStorageIo.java
+++ b/appinventor/appengine/src/com/google/appinventor/server/storage/ObjectifyStorageIo.java
@@ -18,7 +18,6 @@ import com.google.appengine.api.memcache.ErrorHandlers;
 import com.google.appengine.api.memcache.MemcacheService;
 import com.google.appengine.api.memcache.MemcacheServiceFactory;
 import com.google.appengine.api.memcache.Expiration;
-import com.google.apphosting.api.ApiProxy;
 import com.google.appinventor.server.CrashReport;
 import com.google.appinventor.server.FileExporter;
 import com.google.appinventor.server.GalleryExtensionException;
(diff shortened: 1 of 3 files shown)


appinventor/appengine/src/com/google/appinventor/server/storage/ObjectifyStorageIo.java
```diff
@@ -2397,18 +2397,18 @@ private static String collectUserProjectErrorInfo(final String userId, final lon
...
 
   @Override
   public InputStream openTempFile(String fileName) throws IOException {
-    if (!fileName.startsWith("__TEMP__")) {
+    if (!fileName.startsWith(TEMP_PREFIX)) {
       throw new RuntimeException("deleteTempFile (" + fileName + ") Invalid File Name");
     }
     // Use FileData.RoleEnum.TARGET because these temp files never live very long
-    GcsFilename gcsFileName = new GcsFilename(getGcsBucketToUse(FileData.RoleEnum.TARGET), fileName);
+    GcsFilename gcsFileName = new GcsFilename(getGcsBucketToUse(FileData.RoleEnum.TEMPORARY), fileName);
```
barreeeiroo, line 2278: Note this did not have any impact, but worth to keep it with the expectation to prevent surprises in the future...

barreeeiroo:
Also, I'm not very familiar with how Objectify / Datastore transactions work. I've seen some calls to `runJobWithRetries` that only perform a single change, and they are marked to be executed as a transaction. Is this expected, or how does this work?

And from the kind of transactions I looked at, I also don't find them as critical as to require transactions. I'm comparing here against DynamoDB, and despite DDB supports transactions too, it seems like just running a single operation with conditional writes would suffice and prevent the need for a transaction.

barreeeiroo:
nit: this PR would be superseded by https://github.com/mit-cml/appinventor-sources/pull/3572

jisqyv (changes requested):
I want to hold off on this change, as it complicates the merge with the stand alone and PostgreSQL versions. In particular removing setUserEmail is a problem for now.

PR #3558, opened 2025-08-13
Reduce calls to Datastore to assert ownership

barreeeiroo:
General items:

- [x] My code follows the:

- [ ] I branched from `ucr`

For all other changes:

What does this PR accomplish?

*Description*

Leverage memcache to store the project ownership, rather than going to Datastore for each single assertion. This reduces the overall cost of Datastore in Google App Engine, as Memcache is included in its pricing.

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/server/storage/ObjectifyStorageIo.java
+++ b/appinventor/appengine/src/com/google/appinventor/server/storage/ObjectifyStorageIo.java
@@ -135,6 +135,9 @@ public class ObjectifyStorageIo implements StorageIo {
   private static final String GCS_BUCKET_NAME;
   private static final String APK_BUCKET_NAME;
 
+  private static final String BUILD_STATUS_CACHE_KEY_PREFIX = "40bae275-070f-478b-9a5f-d50361809b99";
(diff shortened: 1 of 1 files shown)


appinventor/appengine/src/com/google/appinventor/server/storage/ObjectifyStorageIo.java
```diff
@@ -2706,6 +2721,9 @@ public void run(Objectify datastore) throws ObjectifyException, IOException {
             throw new SecurityException("Unauthorized access");
           }
           // User has data for project, so everything checks out.
+          // We just store it now in the cache for future access, as we know the user requesting
+          //   this project owns it.
+          memcache.put(cacheKey, userId);
```
barreeeiroo, line 2726: Just one clarification: we only cache the project on happy authorized paths. If an user requests a project not owned by them, and it's not in the cache, we have no way to identify the correct owner. So, we will only cache for legitimate calls, which actually makes sense given the access pattern (the user creates a project, tries to read it within a few milliseconds, now it's cached).

I did also not set any TTL as this field does not mutate, hence safe to never expire.

Issue #3545, opened 2025-08-05
ValueFormat in Chart throws error in iOS


bobparks:
**Describe the bug**

ValueFormat in Chart throws an error in iOS. This property is often used in data science materials with time series data involving a year. 

Using Companion 2.74 (Build 20) and Companion 2.73.1.

**Affects**

ValueFormat property in Chart 

- [ x] Designer
- [ ] Blocks editor
- [ ] Projects Explorer
- [ ] Companion
- [ ] Compiled apps
- [ ] Buildserver
- [ ] Debugging
- [ ] Other... (please describe)

**Steps to reproduce**

[Here's an image of the error](https://drive.google.com/file/d/12W_R2BwEDITWnlFzDY5Rct01geVx8jtF/view?usp=drive_link
) in iOS.

ewpatton:
I believe this is just a situation where the Android version got ahead of the iOS version. We'll need to implement the iOS version of this property.

cindyloo:
add   `var _valueType: Int = 0`, ```  public func setValueType(valueType: Int){}``` and in `ChartDataModel` convert `Int` to `String`

cindyloo:
@bobparks can you check this one? I see it working on my phone?

cindyloo:
![Image](https://github.com/user-attachments/assets/181b3da7-47a2-4e78-9a38-3921781c2c44)

bobparks:
This looks like it's using Date in the ValueFormat. Let’s make sure we can do “Integer” ValueFormat as the data type instead of Date. The issue is that when there’s a year in a spreadsheet, we just want to see it without a decimal point.  There are probably other things that need to appear as integers as well. However, currently, everything in charts on iOS has a decimal.

cindyloo:
if you choose ValueFormat as Date or Time, it will do date or time without decimals... So the focus is on Integer?

bobparks:
Yes, we need to show things as integers for some use cases. That's one of the options in Designer, and in general, it would be weird if it did not work on iOS. 

In this special case, the example I have (a year on a spreadsheet) cannot be used with the Date or Time format because the spreadsheet does not have a column with UTC values. Lots of student spreadsheets will simply have a Year as an integer and not convert to UTC.

Issue #3544, opened 2025-08-05
Labels running over 8 characters in HorizontalArrangments causing problems in iOS


bobparks:
**Describe the bug**

Labels running over 8 characters in HorizontalArrangments with other labels will skip to the next line, only in iOS. 

Using Companion 2.74 (Build 20) and Companion 2.73.1.

<img width="331" height="127" alt="Image" src="https://github.com/user-attachments/assets/3ce1da21-0b3f-4207-9e31-703b42c6dab3" />

They are not running over to the next line for lack of space. In practice, there is no way to remedy the issue in iOS without it looking very strange (for example, swithing the label's Width property from Automatic to Percent).

**Affects**

- [ x] Designer
- [ ] Blocks editor
- [ ] Projects Explorer
- [ ] Companion
- [ ] Compiled apps
- [ ] Buildserver
- [ ] Debugging
- [ ] Other... (please describe)

**Expected behavior**

The example AIA will show how the label along the bottom of the screen should look. In practice, there is no way to remedy the issue in iOS without it looking very strange (for example, switching the label's Width property from Automatic to Percent).

**Steps to reproduce**

[Here is an example AIA](https://drive.google.com/file/d/1aZMLJ-VF5qdefOKftRORljS-Y3Iu7_IW/view?usp=drive_link). Please pay attention to the HorizontalArrangement along the bottom of the screen in Android and iOS. 

Here is [an image of a screen with the issue](https://drive.google.com/file/d/14mTfDEBgZYGFxiqhQCTZHuASSzrNQOex/view?usp=drive_link
(break is along the bottom).

cindyloo:
not exactly sure where the problem may be, but upon initial review

`HVArrangement` sets the childWidth like this:
```
  open func setChildWidth(of component: ViewComponent, to width: Int32) {
    guard let form = form else {
      return
    }
    if width <= kLengthPercentTag {
      _view.setWidth(of: component.view, to: Length(percent: width, of: form.scaleFrameLayout))
    } else if width == kLengthPreferred {
      _view.setWidth(of: component.view, to: .Automatic)
    } else if width == kLengthFillParent {
      _view.setWidth(of: component.view, to: .FillParent)
    } else {
      _view.setWidth(of: component.view, to: Length(pixels: width))
    }
    _view.setNeedsLayout()
  }
```

and the `.Automatic` static var is defined in `Length` class inside `LinearView`
```
public class Length: Equatable {
  static let Automatic = Length(-1)
[...]

PR #3580, opened 2025-08-28
ios updates for Robert's issues 

cindyloo:
Fixes # .
https://github.com/issues/assigned?issue=mit-cml%7Cappinventor-sources%7C3544
https://github.com/issues/assigned?issue=mit-cml%7Cappinventor-sources%7C3545


ewpatton:
I'm staging this for TestFlight.

ewpatton:
@bobparks This fix is on TestFlight. Can you test the features that weren't working for you and confirm it addresses the issues you encountered?

appinventor/components-ios/src/Chart.swift
```diff
@@ -78,6 +79,23 @@ import DGCharts
...
+  @objc open var ValueFormat: Int {
+     get {
+       return _valueType;
+     }
+    set{
+      _valueType = newValue;
+      _chartView?.setValueType(valueType: newValue);
+      
+      for c in _dataComponents {
+        if let dc: ChartDataModel = c.dispatchDelegate as? ChartDataModel{
```
ewpatton: Note that the type assertion `: ChartDataModel` here is redundant since it is inferrable from the `as?` cast.

appinventor/components-ios/src/Chart.swift
```diff
@@ -78,6 +79,23 @@ import DGCharts
...
+          dc.setValueType(valueType:newValue)
+        }
+        
+      }
+    }
+   }
```
ewpatton: This looks fine just clean up the whitespace/indentation.

appinventor/components-ios/src/ChartDataModel.swift
```diff
@@ -226,4 +227,13 @@ open class ChartDataModel: DataModel {
...
   }
+  
+  public func setValueType(valueType: Int){
+    _valueType = valueType;
+    
+    dataset?.valueFormatter = DefaultValueFormatter(formatter: NumberFormatter())
+    
+    view.setValueType(valueType: valueType)
+  }
+
```
ewpatton: It would be more Swifty to make this a var with getter/setter.

appinventor/components-ios/src/PointChartDataModel.swift
```diff
@@ -47,7 +47,19 @@ open class PointChartDataModel: Chart2DDataModel {
...
-      return Double(value)
+      let retVal = Double(value)
+      if retVal == nil {
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = "yyyy-MM-dd"
+        let date = dateFormatter.date(from: value)
+        if date == nil {
+          dateFormatter.dateFormat = "HH:mm:ss"
+          return dateFormatter.date(from: value)?.timeIntervalSince1970
+        }
```
ewpatton: It's more efficient to have statically allocated variables for these DateFormatters rather than allocating them on every call. This might be called hundreds of times to draw a graph.

appinventor/components-ios/src/LinearView.swift
ewpatton, line 1: Is this a critical change to address one of the mentioned PRs, or could it be its own change in case we need to revert it?

appinventor/components-ios/src/LinearView.swift
```diff
@@ -42,7 +42,7 @@ public class Length: Equatable {
     view = other.view
   }
 
-  private init(_ fixed: Int32) {
+  public init(_ fixed: Int32) {
```
ewpatton: Why was this made public? It seems like whatever code would call this can just call `LinearView(pixels:)`

appinventor/components-ios/src/ChartView.swift
```diff
@@ -9,13 +9,22 @@ import DGCharts
...
 
   var data: DGCharts.ChartData?
+  
+  /*
+  Chart value types
+   */
+  public static var CHART_VALUE_DECIMAL: Int = 0;
+  public static var CHART_VALUE_INTEGER: Int = 1;
+  public static var CHART_VALUE_DATE: Int = 2;
+  public static var CHART_VALUE_TIME: Int = 3;
```
ewpatton: Remove the extraneous semicolons.

In the long term, we may want to introduce this as an enum internally since it's a closed set.

appinventor/components-ios/src/ChartView.swift
```diff
@@ -89,6 +98,11 @@ open class ChartView {
     self.chart?.setNeedsDisplay()
   }
 
+  public func setValueType(valueType: Int){
```
ewpatton: Whitespace before curly brace

appinventor/components-ios/src/ChartView.swift
```diff
@@ -89,6 +98,11 @@ open class ChartView {
...
+    _valueType = valueType;
```
ewpatton: Extraneous semicolon

appinventor/components-ios/src/ChartView.swift
```diff
@@ -9,13 +9,22 @@ import DGCharts
 open class ChartView {
   unowned let _chartComponent: Chart
   private let _workQueue = DispatchQueue(label: "Chart", qos: .userInitiated)
-
+  var _valueType: Int = 0
```
ewpatton: Consider making this private.

appinventor/components-ios/src/BarChartDataModel.swift
```diff
@@ -38,15 +38,30 @@ open class BarChartDataModel: Chart2DDataModel {
...
+      //}
       // Add the new entry at the end
       _entries.append(entry)
     }
   }
 
+  
+  public func formatDateTime(_ value: String) -> Double? {
+    
+    let dateFormatter = DateFormatter()
```
ewpatton, line 48: While it's less critical here most likely versus the other place where I commented about this, but we probably want DateFormatters to be statically cached so that if this is used inside of a loop we don't pay the penalty for every new data point added.

appinventor/components-ios/src/BarChartDataModel.swift
```diff
@@ -62,8 +77,10 @@ open class BarChartDataModel: Chart2DDataModel {
       let flooredX = Int(floor(x))
       return BarChartDataEntry(x: Double(flooredX), y: y)
     } else {
-      // Handle number format exception
-      return nil
+      let dateX = formatDateTime(rawX) ?? 0.0
+      let y = Double(rawY) ?? 0.0
```
ewpatton: Is casting to 0.0 on failure how the Android code works? I though we raised an error via the form.

bobparks:
> @bobparks This fix is on TestFlight. Can you test the features that weren't working for you and confirm it addresses the issues you encountered?

Thanks for working on this; however, neither issue appears to be fixed on 2.74(29) on Testflight. There is a new issue with buttons, described below. 

- [#3544 ](https://github.com/mit-cml/appinventor-sources/issues/3544) does not appear to be fixed. The original problem was that iOS puts one label on two lines and then the following label on the first line. A text analogy would be:

A quick                 fox
              brown

[Here's the test AIA I am using](https://drive.google.com/file/d/1VuhYuxGYeWLiWuk-fKQ4BshTmj3Jp5p7/view?usp=sharing). Here's an image of what I see on Testflight on 2.74 (The part in question is the bottom row of labels): 
image

Here's what it should look like on Android: 
image

More alarming is the new way the buttons have become stretched across the screen in 2.74(29). Compare the image above of the 2.74 versions with the one using 2.73.1: 
image

Perhaps we don't worry about the original #3544 issue with the text jumping across two lines. (That can be fixed in App Inventor by scrunching up the labels to make them fewer than 8 characters.) More important would be to make the buttons go back to their original state. 

- [#3545](https://github.com/mit-cml/appinventor-sources/issues/3545) With this issue, the property no longer throws an error, but the UI does not obey the property. Look at the X axis in [2.74(29)](https://drive.google.com/file/d/1oC0HU6fxJjSgrB_oZB9_U2IBJkuAFOeA/view?usp=drive_link) and in [Android](https://drive.google.com/file/d/1SwdU1CP5bwA01Qynx4rwfDLzDWgs9_Nn/view?usp=drive_link).

cindyloo:
@bobparks thanks for providing a project and some additional examples. Clearly while I simulated what I thought was happening and address what appeared to be IOS "layout priorities", I needed a more in-depth test. I'll look at this again

appinventor/components-ios/src/PointChartDataModel.swift
cindyloo:
```if retVal == nil {
        self.dateFormatter.dateFormat = "yyyy-MM-dd"
        let date = self.dateFormatter.date(from: value)
        if date == nil {
          self.dateFormatter.dateFormat = "HH:mm:ss"
          return self.dateFormatter.date(from: value)?.timeIntervalSince1970
        }
        return date?.timeIntervalSince1970
      }``` or should I make one for date and one for time?

File: appinventor/components-ios/src/LinearView.swift
Comment (cindyloo) [reply appinventor/components-ios/src/LinearView.swift:1]: rolled back

**Comment by cindyloo** [comment]
IMG_2887

**Comment by bobparks** [comment]
Looks good!

File: appinventor/components-ios/src/PointChartDataModel.swift (hunk 4 above)
Comment (ewpatton) [reply appinventor/components-ios/src/PointChartDataModel.swift]: Hmm. For this change I was thinking that we would have two static DateFormatters, one each for date and time. While generally we never would access these functions on multiple threads, there is an obvious issue with reusing that object if multiple threads are ever involved (mostly here I'm thinking about a test branch I did where the blocks and UI ran on separate threads). We should aim to try and make the code thread safe.

Comment (cindyloo) [reply appinventor/components-ios/src/PointChartDataModel.swift]: ok, will edit

File: appinventor/components-ios/src/BarChartDataModel.swift [hunk 12]
```diff
@@ -38,15 +38,30 @@ open class BarChartDataModel: Chart2DDataModel {
...
       _entries.insert(entry, at: x)
     } else {
+      // CSB removing b/c we don't need intermediary entries
       // If x is beyond the current range of entries, fill the gap with placeholder entries
-      while entries.count < x {
-        _entries.append(BarChartDataEntry(x: Double(entries.count), y: 0.0))
-      }
+      //while entries.count < x {
+      //  _entries.append(BarChartDataEntry(x: Double(entries.count), y: 0.0))
+      //}
```
SusanRatiLane: Do we need to keep this commented code?

appinventor/components-ios/src/BarChartDataModel.swift
```diff
@@ -62,8 +77,18 @@ open class BarChartDataModel: Chart2DDataModel {
...
+      do {
+        guard let dateX = formatDateTime(rawX) else {
+            throw ChartDataError.invalidDateString(rawX)
+        }
+        guard let y = Double(rawY) else {
+            throw ChartDataError.invalidValue(rawY)
+        }
+        return BarChartDataEntry(x: dateX, y: y)
+      } catch {
+        print("Error parsing x or y coordinates: \(rawX), \(rawY), error:\(error)")
```
SusanRatiLane: It looks like this is just writing an error to the log. Do we want to throw an error here?

appinventor/components-ios/src/AxisChartView.swift
```diff
@@ -103,12 +106,38 @@ open class AxisChartView : ChartView {
...
         return _axisLabels[integerValue]
       } else {
-        return String(value)
+        return parseForFormat(value)
       }
     }
 
     public func stringForValue(_ value: Double, axis: DGCharts.AxisBase?) -> String {
+      if _vType == CHART_VALUE_INTEGER {
+        return String(Int(value));
```
SusanRatiLane: Another of those pesky ;

appinventor/components-ios/src/ChartView.swift
```diff
@@ -89,6 +98,16 @@ open class ChartView {
     self.chart?.setNeedsDisplay()
   }
 
+  
+  public var ValueType: Int {
+    get {
+      return _valueType
+    }
+    set(valueType) {
+      _valueType = valueType;
```
SusanRatiLane: Semicolon

SusanRatiLane (changes requested):
Just a couple more items.

appinventor/components-ios/src/BarChartDataModel.swift
cindyloo: I wasn't sure. I guess since the java correlative does it, I should

SusanRatiLane: > I wasn't sure. I guess since the java correlative does it, I should

Yes, I think so.

SusanRatiLane (approved):

PR #3648, opened 2025-10-30
Add "App Library": Ability to cache apps in the Companion

jisqyv:
This change permits the companion app to save projects for use when not connected. Two new menu choices are added to the system menu

 * Save Project: Saves the connected project in its current state
 * Load Project: Brings up a page where you can load or delete a saved project

Change-Id: Ifc23e45a2f9a49c45db36fba16f4f01d7b7f957f

General items:

- [ ] My code follows the:

- [x] I branched from `ucr`

For all other changes:

What does this PR accomplish?

*Description*

Fixes # .

Resolves # .

### DIFF
--- a/appinventor/aiplayapp/assets/projects.html
+++ b/appinventor/aiplayapp/assets/projects.html
@@ -29,7 +29,7 @@
             } else {
               html += "<table border=1>\n";
               for (let i=0; i < projects.length; i++) {
-                html += "<tr><td>" + projects[i] + "</td><td><button onClick=\"loadproject('" + projects[i] + "');\">Run</button></td><td>&nbsp;</td><td><button onClick=\"deleteproject('" + projects[i] + "');\">Delete</button></td></tr>\n"
+                html += "<tr><td>" + projects[i] + "</td><td><button onClick=\"loadproject('" + projects[i] + "');\">Run</button></td><td>&nbsp;-&nbsp;</td><td><button onClick=\"deleteproject('" + projects[i] + "');\">Delete</button></td></tr>\n"
               }
               html += "</table>\n";
             }


SusanRatiLane:
First observations from testing: 
1. When I save a project, there's no progress bar or notification that it has succeeded.
2. The iOS Companion has a button in that menu to close the project and go back to the connect screen. Android doesn't have that and really should
3. If the app using Classic theme or I hide the title bar, the hamburger menu does not appear
4. iOS has a My Projects button on the connect screen. Android really ought to match.

jisqyv:
@SusanRatiLane 
> First observations from testing:
>
> 1. When I save a project, there's no progress bar or notification that it has succeeded.
> 2. The iOS Companion has a button in that menu to close the project and go back to the connect screen. Android doesn't have that and really should
> 3. If the app using Classic theme or I hide the title bar, the hamburger menu does not appear
> 4. iOS has a My Projects button on the connect screen. Android really ought to match.

1. (point 1 above) I didn't add a progress bar but did add a “Toast” when the project
   is saved.
4. (point 4 above) I added the “my saved apps” button on the home screen. If you do
   not have any saved apps, it will take you to the saved apps screen
   which because you do not have any apps, will instead display a
   message (not yet written) about what saved apps are.
2. (point 2 above) Not sure that we can easily do this. Once a project is loaded, the
   initial Companion UI is gone (this is a difference, I believe, from
   the iOS Companion). We could have an implied saved project, which is
   the Companion itself, which we can switch to. But it may require
   cleaning up WebRTC state, etc. It may not be worth the effort at
   this point. From the menu, you **can** go back to the project
   choice screen and select another project.
3. (point 3 above) This is an Android limitation. Older devices had a different way to
   get to the app “menu.” This will still work with the Classic theme
   on an older device. However, I’m not sure what we can do when
   someone uses a Classic theme and loads the App on a newer device.
   Keep in mind that the Companion itself uses “Device Default,” so
   people should always be able to get to the projects menu. Just once
   a project using the Classic theme is loaded, there isn’t a way back
   to the projects screen.

Foo! Markdown renumbered the points above...(I just added some annotations to point out the correct bullet point I am responding too).

PR #3659, opened 2025-11-14
Include backtick in disallowed component name charset

ewpatton:
Change-Id: I8cdcbe0810b5afbda01782897048f61dce6a3281

General items:

- [x] My code follows the:

For all other changes:

What does this PR accomplish?

This PR fixes a bug [reported in the community](https://community.appinventor.mit.edu/t/backtick-component-name-bug/159445?u=ewpatton) where a student changed a component name to <code>Clock\`</code>, breaking the generated YAIL code. Backtick was included in the regex, but would only match on the first char. I refactored how the regex is constructed to make the relationship between the groups more clear (relatively speaking).


ewpatton:
@mit-cml/core-reviewers I'd like to squeeze this one into the nb203 release. The low level change is just the addition of the backtick character to the disallow list, but I did refactor a bit so we don't have so much repetition, which caused it to be overlooked in the first place.

PR #3675, opened 2025-12-03
fix: clarify PNG-only support for App Icon instructions (i18n + docs)

apexpdl:
This PR clarifies that only PNG images are supported for the App Icon property.

Changes:
- Added `ComponentPropertyTranslations.properties` with base English text specifying PNG-only support.
- Updated the `userinterface.md` documentation to reflect the same PNG-only instruction.

This prevents confusion for users uploading JPEG icons, which can cause compilation or installation failures.

Fixes: none
Resolves: none


AppInventorWorkerBee:

AppInventorWorkerBee:

SusanRatiLane:

ewpatton:
@SusanRatiLane @jisqyv I'd rather we figure out why JPEG icons are no longer working on the buildservers rather than restricting users to just PNGs. Once upon a time we definitely allowed JPEGs.

apexpdl:
Thanks @ewpatton — agreed, the better long-term fix is to restore JPEG support instead of restricting users. I’ll try to solve this by reproducing the issue with the sample AIA from #3641 and checking where the buildserver might be rejecting JPEGs — does that sound good?

ewpatton:
We ended up fixing JPG support in the build server image, so we no longer need to enforce PNG only.

PR #3682, opened 2025-12-08
Fix CSV parsing error by stripping UTF-8 BOM before parsing

harshrajsinh-dev:
## Fix: Remove UTF-8 BOM Before CSV Parsing

This PR fixes an issue where CSV files containing a UTF-8 BOM (`EF BB BF`) fail when used with the **list from CSV table** and **list from CSV row** blocks.

Certain editors automatically insert a BOM at the start of text files. When passed into `fromCsvTable` or `fromCsvRow`, the BOM becomes the first character of the first cell, causing the CSV parser to throw an error.

### ✅ What This PR Does
- Detects a leading UTF-8 BOM (`\uFEFF`) in the input CSV string
- Safely removes it before creating the `CsvParser`
- Prevents “Cannot parse list as CSV table” errors
- Has no effect on files that do not contain a BOM

### 🔧 Modified Methods
- `fromCsvTable`
- `fromCsvRow`

### 📌 Why This Matters
This improves App Inventor’s robustness, especially when users load CSV files created in Excel, Google Sheets, or text editors that include a BOM by default.

### 🔗 Related Issue
Fixes Issue #948: “Cannot parse list as CSV table error with BOM marker”

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/util/CsvUtil.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/util/CsvUtil.java
@@ -26,7 +26,23 @@ public final class CsvUtil {
   private CsvUtil() {
   }
 
+  // ! old code 
+  // public static YailList fromCsvTable(String csvString) throws Exception {
+  //   CsvParser csvParser = new CsvParser(new StringReader(csvString));
+  //   ArrayList<YailList> csvList = new ArrayList<YailList>();
+  //   while (csvParser.hasNext()) {
+  //     csvList.add(YailList.makeList(csvParser.next()));
+  //   }
+  //   csvParser.throwAnyProblem();
+  //   return YailList.makeList(csvList);
+  // }
+  
   public static YailList fromCsvTable(String csvString) throws Exception {
+    // FIX: strip UTF-8 BOM if present
+    if (csvString != null && csvString.startsWith("\uFEFF")) {
+        csvString = csvString.substring(1);
+    }
+
     CsvParser csvParser = new CsvParser(new StringReader(csvString));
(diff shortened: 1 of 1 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

appinventor/components/src/com/google/appinventor/components/runtime/util/CsvUtil.java
```diff
@@ -26,7 +26,23 @@ public final class CsvUtil {
...
+  // ! old code 
+  // public static YailList fromCsvTable(String csvString) throws Exception {
+  //   CsvParser csvParser = new CsvParser(new StringReader(csvString));
+  //   ArrayList<YailList> csvList = new ArrayList<YailList>();
+  //   while (csvParser.hasNext()) {
+  //     csvList.add(YailList.makeList(csvParser.next()));
+  //   }
+  //   csvParser.throwAnyProblem();
+  //   return YailList.makeList(csvList);
+  // }
```
ewpatton, line 38: There's no need to keep the old code in a comment. That's what version control is for.

appinventor/components/src/com/google/appinventor/components/runtime/util/CsvUtil.java
```diff
@@ -36,6 +52,7 @@ public static YailList fromCsvTable(String csvString) throws Exception {
     return YailList.makeList(csvList);
   }
 
+
```
ewpatton, line 55: Spurious newline added.

ewpatton (changes requested):

ewpatton:
We may want to think about a similar check in the iOS code.

PR #3685, opened 2025-12-13
build: add AAPT2 command/env logging (debug JPEG icon issue)

apexpdl:
Temporary logging to print AAPT2 compile/link commands and key paths (android.jar, assets, manifest) so we can diagnose why JPEG app icons fail on CI.

This PR does NOT change behavior—only adds logs.
icon

Use this icon.jpg as the App Icon (JPEG) to reproduce.


AppInventorWorkerBee:

AppInventorWorkerBee:

apexpdl:

Repro: square icon.jpg (512×512). Build with this JPEG as App Icon.
Purpose: capture AAPT2 command + env logs to debug JPEG icon failure on CI.

apexpdl:
@ewpatton @SusanRatiLane could you please trigger CI on this PR?
It only adds logging so we can capture the AAPT2 commands/paths for the JPEG icon repro.

ewpatton:
The core issue was that in the production environment we run on Alpine Linux and we didn't have the necessary shared libraries needed to process JPEG files. This has been resolved.

PR #3704, opened 2026-01-05
Make Spinner trigger AfterSelection even if invisible

ewpatton:
The Android Spinner widget does not call the onItemSelected callback if its visibility is View.GONE. This commit updates our Spinner implementation so that it will be View.INVISIBLE with a fixed width and height of 0. This gives the effect of being View.GONE while still allowing for the callbacks to occur.

Fixes mit-cml/appinventor-sources#922

Change-Id: I2e944c97f6520cd0f3154d3886f632fb645ccb03

General items:

- [ ] My code follows the:

- [ ] I branched from `ucr`

For all other changes:

What does this PR accomplish?

*Description*

Fixes # .

Resolves # .

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/Spinner.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/Spinner.java
@@ -56,6 +56,8 @@ public final class Spinner extends AndroidViewComponent implements OnItemSelecte
   private YailList items = new YailList();
   private int oldAdapterCount;
   private int oldSelectionIndex;
+  private int savedWidth;
+  private int savedHeight;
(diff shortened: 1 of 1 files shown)


ewpatton:
Note that I've submitted this against master rather than ucr. My logic is that it changes the internal workings of existing properties without really changing the semantics of those properties. Existing apps will continue to work with both old and new companions.

SusanRatiLane:
This did not test out with this basic test app. 
[Test.aia.zip](https://github.com/user-attachments/files/25243785/Test.aia.zip)

Do I not understand how this is supposed to work?

PR #3711, opened 2026-01-07
URL-encode brackets in Material Symbols font path

mjgallag:
This fixes the local dev server failing to load the icon font, causing Material Symbols to render as text.

Jetty 12.0.31 (shipped in gcloud SDK 550.0.0) enforces strict RFC 3986 URL parsing and rejects unencoded square brackets in paths.

https://github.com/jetty/jetty.project/pull/14013


mjgallag:
This is working on ai2-test.

PR #3746, opened 2026-01-31
Add defensive guards to Safari focus implementation

nitin-kumar-code:
This PR adds defensive checks to the Safari-specific focus implementation to prevent JavaScript runtime errors in edge cases.

Specifically, it:
1. Guards against null or unsupported elements before calling focus
2. Wraps the focus call in a try/catch to avoid crashes when elements are detached or focus is disallowed
3. Preserves existing behaviour for valid focus calls (preventScroll and timing unchanged)

This is a Safari/iOS-specific safety improvement with no API or functional behaviour changes in normal cases.


AppInventorWorkerBee:

AppInventorWorkerBee:

josmas:
The formatting looks incorrect, there is no explanation of why this is needed and it introduces an ignored catch, which is never a good sign. Closing for now. Please make a stronger case for changes like this in the future.

nitin-kumar-code:
Thanks for the feedback- understood

PR #3756, opened 2026-02-03
Clarify non-null contract for identifier validators

nitin-kumar-code:
### Summary

This PR makes the non-null contract of identifier validation methods explicit by enforcing it with ⁠ Objects.requireNonNull ⁠ and documenting the behavior via Javadoc.

### Background / Motivation

The identifier validators in ⁠ TextValidators ⁠ assume non-null inputs and are used to validate syntax, not to sanitize or recover from programming errors. Previously, adding silent null handling (e.g., returning ⁠ false ⁠ for ⁠ null ⁠) unintentionally changed this contract and could mask upstream bugs.

As discussed in earlier review feedback, failing silently in these cases is undesirable.

### What this PR does

•⁠  ⁠Enforces the existing non-null contract explicitly using ⁠ Objects.requireNonNull ⁠
•⁠  ⁠Documents the contract in Javadoc (⁠ @param ⁠ and ⁠ @throws ⁠)
•⁠  ⁠Preserves all existing validation semantics for non-null inputs
•⁠  ⁠Avoids introducing silent failure or behavioral changes

### Scenario

If a ⁠ null ⁠ value reaches an identifier validator, this indicates a programming error upstream rather than an invalid user input. By failing fast with a clear exception, the issue becomes immediately visible and debuggable instead of being silently converted into a validation failure.

### What this PR does NOT do

•⁠  ⁠Does not treat ⁠ null ⁠ as a valid or expected input
•⁠  ⁠Does not change validation rules or accepted identifiers
•⁠  ⁠Does not introduce new error handling paths or logging

### Notes

This PR follows up on the discussion in #3721 and addresses the feedback about avoiding silent failure by making the non-null contract explicit.

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/youngandroid/TextValidators.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/youngandroid/TextValidators.java
@@ -8,6 +8,7 @@ package com.google.appinventor.client.youngandroid;
 
 import com.google.appinventor.client.Ode;
 import static com.google.appinventor.client.Ode.MESSAGES;
+import java.util.Objects;
 
 import com.google.appinventor.client.editor.simple.SimpleComponentDatabase;
 import com.google.appinventor.client.editor.youngandroid.YaProjectEditor;
@@ -173,10 +174,12 @@ public final class TextValidators {
    * (unaccented English) letters, digits, or underscores.
    *
    * @param text the proposed identifier
+   * @throws NullPointerException if text is null
    * @return {@code true} if the argument is a legal identifier, {@code false}
    *         otherwise
    */
   public static boolean isValidIdentifier(String text) {
+    Objects.requireNonNull(text, "Identifier must not be null");
     return text.matches("^[a-zA-Z]\\w*$");
   }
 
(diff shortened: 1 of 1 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

ewpatton:
Reviewing the code paths into these functions, I do not see any path that could lead to an input being null to begin with. Do you have an actual proof of concept to demonstrate that this is needed?

nitin-kumar-code:
Thanks for the review and for prompting a closer look at this.

I traced all call sites of the affected validators (⁠ isValidIdentifier ⁠, ⁠ isValidComponentIdentifier ⁠, ⁠ isValidCharFilename ⁠, and ⁠ isValidLengthFilename ⁠) and followed their inputs back to their sources. This included project, folder, component, and file upload flows (e.g., ⁠ AddFormCommand ⁠, ⁠ MockComponent ⁠, ⁠ SubsetJSONPropertyEditor ⁠, ⁠ FileUploadWizard ⁠, and related paths).

In all cases I reviewed, the values originate from GWT UI widgets or internal normalized strings, which are guaranteed to be non-null. Empty input is represented as an empty string and is explicitly guarded before validation. I also tested dialog cancellation and rapid UI interactions locally and couldn’t reproduce a scenario where validation runs with a null value.

One path I wasn’t able to fully verify locally is the project rename flow, but based on the surrounding code structure and the behavior of other rename dialogs, I didn’t find evidence that it behaves differently or allows null values to reach validation.

Given this, I agree that adding explicit null handling here isn’t justified and that the validators should continue to rely on the existing non-null contract. I’m closing this PR accordingly.

Thanks again for the guidance — the review was helpful in validating the assumptions around these code paths.

PR #3758, opened 2026-02-04
Fix Road map tiles not rendering due to OSM redirects

### LINKED ISSUE #3723: Map component - no map displayed
An app which was working for years using Openstreetmap (OSM) road map type does not show any map anymore. As described in a post in the MIT app inventor forum [(link)](https://community.appinventor.mit.edu/t/map-not-displayed/156514/4), no map is shown in the compiled app. Especially when one switches back to road map type (from a different screen or a map component using a different map type) one gets an empty map.
This behaviour is reproducible and happened seemingly in several published apps since about September 2025.
I have tested this in online and offline app inventor, no difference in the build apps.

**Affects**

- [ ] Designer
- [ ] Blocks editor
- [ ] Projects Explorer
- [ ] Android Companion
- [ ] iOS Companion
- [x ] Android Compiled APK/AAB
- [ ] iOS Compiled IPA
- [ ] Buildserver
- [ ] Debugging

**Expected behavior**
Show a map as defined by the map component (using map type road and the default tile server from Openstreetmap).

**Steps to reproduce**
As described above: Build test app only with a layout to show a map component (map type road).

Akashgite1:
General items:

- [X] My code follows the:

- [ ] I branched from `ucr`

For all other changes:

What does this PR accomplish?

*Description*
This PR fixes an issue where the Road map type renders as a blank or grey map.

OpenStreetMap tile servers now return HTTP redirects, which are not reliably
handled by the osmdroid version used by App Inventor. As a result, Road map
tiles fail to load, especially after switching map types (Road → Aerial → Road).

The fix replaces the redirecting endpoint with explicit OpenStreetMap tile
subdomains (a/b/c.tile.openstreetmap.org) and clears the tile cache when
switching back to Road, restoring correct behavior across Companion and
compiled apps.

Fixes #3723.

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/util/NativeOpenStreetMapController.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/util/NativeOpenStreetMapController.java
@@ -106,6 +106,20 @@ class NativeOpenStreetMapController implements MapController, MapListener {
   /* end copied from SVG */
 
   private static final String TAG = NativeOpenStreetMapController.class.getSimpleName();
+  private static final ITileSource OSM_ROAD_TILE_SOURCE =
+    new XYTileSource(
+        "OSM-Road",
+        1,
+        20,
+        256,
+        ".png",
+        new String[] {
+            "https://a.tile.openstreetmap.org/",
+            "https://b.tile.openstreetmap.org/",
+            "https://c.tile.openstreetmap.org/"
+        },
+        "© OpenStreetMap contributors"
+  );
   private boolean caches;
   private final Form form;
   private RelativeLayout containerView;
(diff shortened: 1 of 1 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

Akashgite1:
Verified locally on Android Companion and compiled APK.

Switching map types now works correctly:
- Road map loads as expected
- Switching Road → Aerial → Road no longer results in a blank/grey map

image

Screenshots below show:
1) Initial Road map state


2) Aerial map view


3) Road map correctly reloaded after switching back

appinventor/components/src/com/google/appinventor/components/runtime/util/NativeOpenStreetMapController.java
```diff
@@ -452,12 +452,28 @@ public String getTileURLString(MapTile aTile) {
     return tileSource;
   }
 
+  private ITileSource getRoadTileSource() {
```
ewpatton: Rather than constructing this as a new object every time, create a static final field. Also, make sure to include the attribution string "© OpenStreetMap contributors" in the extra argument to XYTileSource to provide proper attribution.

appinventor/components/src/com/google/appinventor/components/runtime/util/NativeOpenStreetMapController.java
```diff
@@ -452,12 +452,28 @@ public String getTileURLString(MapTile aTile) {
...
+    );
+  }
+
   @Override
   public void setMapTypeAbstract(MapType type) {
     tileType = type;
     switch (type) {
       case Road:
-        view.setTileSource(TileSourceFactory.MAPNIK);
+        view.getTileProvider().clearTileCache();
```
ewpatton: It looks like clearTileCache is indirectly called by setTileSource, so this is probably unnecessary.

Akashgite1:
Thanks for the suggestions!

I've updated the implementation to:
- Use a static final XYTileSource for the Road map
- Include proper "© OpenStreetMap contributors" attribution
- Remove the explicit clearTileCache call since setTileSource already handles it

Please let me know if this looks good now.

ewpatton:

ewpatton (approved):
LGTM

PR #3765, opened 2026-02-07
Add support for android.minSdk project property

### LINKED ISSUE #3761: Implement a AndroidMinSdk project property
Provide a new project property that allows for the user to specify a minSDK in the project.properties file. The value must default to ComponentConstants.APP_INVENTOR_MIN_SDK.

Akashgite1:
General items:

- [ ] My code follows the:

- [ ] I branched from `ucr`

For all other changes:

What does this PR accomplish?

*Description*
This PR adds support for an optional android.minSdk project property that allows advanced users to override the minimum Android SDK version used during builds.

The default App Inventor behavior remains unchanged, and the legacy androidminsdk property is still supported for backward compatibility. Invalid values fall back to the default minimum SDK.

Resolves #3761 .

### DIFF
--- a/appinventor/buildserver/src/com/google/appinventor/buildserver/Project.java
+++ b/appinventor/buildserver/src/com/google/appinventor/buildserver/Project.java
@@ -105,6 +105,7 @@ public final class Project {
   private static final String USESLOCATIONTAG = "useslocation";
   private static final String ANAMETAG = "aname";
   private static final String ANDROID_MIN_SDK_TAG = "androidminsdk";
+  private static final String ANDROID_MIN_SDK_DOT_TAG = "android.minSdk";
   private static final String ACTIONBAR_TAG = "actionbar";
   private static final String COLOR_THEMETAG = "theme";
   private static final String COLOR_PRIMARYTAG = "color.primary";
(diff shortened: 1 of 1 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

josmas:
Hey @Akashgite1 is this PR finished? I can see how you read the value, but how do you set it?

Akashgite1:
Hi @josmas, thanks for checking.

This PR is intentionally limited to buildserver support only. It adds the ability for the build to read an optional android.minSdk value from project.properties and use it during compilation.

At the moment, there is no UI or designer-side mechanism to set this value. The expectation is that advanced users or downstream tooling (e.g., custom exporters, scripts, forks, or manual edits to project.properties) would set it explicitly.

If the property is not present, the build behavior is unchanged and continues to use ComponentConstants.APP_INVENTOR_MIN_SDK. The legacy androidminsdk key is also still supported.

If the preferred direction is to include a setter path (e.g., designer support or project export/import changes), I’m happy to follow up with a separate PR—this one was scoped to unblock the buildserver side first.

Please let me know how you’d like to proceed.

josmas:
I see, I missed the pointer to project.properties. I'll defer to Evan as he was the one opening the issue in the first place.

Akashgite1:
Thanks for the clarification, @josmas — this really helped.

I realize now that my initial understanding was incomplete. I was mostly focusing on buildserver-side support via project.properties, but it’s clear from your explanation that this is intended to be a user-visible project property in App Inventor, which also requires UI changes in the Project Properties dialog, along with the related settings and validation flow.

With this clearer picture, I agree the implementation needs a broader scope than what I originally proposed. I’ll rework the approach and follow up with a new implementation and a fresh PR that aligns with this direction.

Thanks again for taking the time to review and clarify — much appreciated.

PR #3770, opened 2026-02-10
Improve LocationSensor TimeInterval UI with human-readable labels

### LINKED ISSUE #2190: Location Sensor Time Interval
**Describe**

Steve mentioned on my design doc that the TimeInterval designer UI is a bit unintuitive. I didn't think that fell under my project (at least not the first pass) but I wanted to create an issue so that it doesn't get lost.

TimeInterval

The current options provided for the TimeInterval aren't great. The [documentation](http://localhost:8888/reference/components/sensors.html#LocationSensor.TimeInterval) recommends that you don't set the interval below 30,000 ms, but half the options are less than that.

The "0" option is also confusing because it has custom behavior (I believe it tells the phone to send the location whenever it changes). It doesn't actually have to do with time.

And lastly the options don't have any unit information attached to them. I think adding "ms" to the end of each option might make it more intuitive.

This is beneficial to AI's educational mission because it makes the interface more intuitive for users, which allows them to get learning faster.

**Additional info**

I'm not sure how achievable any changes to the UI actually are, because I haven't really dug into the PropertyEditor system. This is just a little bit of wishful thinking. I.e. if it's possible these changes might be nice. :slightly_smiling_face:

Sahitya0805:
General items:

- [x] My code follows the:

For all other changes:

This PR improves the LocationSensor TimeInterval property editor UI to make it more intuitive and user-friendly.

### Problem
The current TimeInterval property editor has several usability issues:
1. Half the options (1000ms, 10000ms) are below the recommended 30,000ms minimum
2. The "0" option is confusing because it has special behavior (updates on location change) but this isn't clear
3. Options don't have unit labels, making it unclear that values are in milliseconds

### Solution
Updated the time interval choices with human-readable labels and appropriate values:

**Before:**
- 0
- 1000
- 10000
- 60000
- 300000

**After:**
- On change (0 ms)
- 30 seconds
- 1 minute
- 5 minutes
- 10 minutes

**Changes:**
- Added descriptive labels (e.g., "1 minute" instead of "60000")
- Clarified the "0" option as "On change (0 ms)" to explain its special behavior
- Removed options below 30 seconds which are below the recommended minimum and can drain battery
- Added "10 minutes" option for longer intervals

### Testing
- ✅ Build successful: `ant webapp` passes
- ✅ Code follows Java style guide
- ✅ Minimal change with proper formatting

Fixes #2190.

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
@@ -5823,4 +5823,19 @@ public interface OdeMessages extends Messages, ComponentTranslations {
   @DefaultMessage("Welcome to App Inventor Neo! If you are looking for the classic App Inventor look, you can switch in the User Interface Settings, or <a href=\"\">click here</a>.")
   @Description("Message shown in the info popup when the user first opens the Neo UI.")
   String neoWelcomeMessage();
+
+  @DefaultMessage("On change (0 ms)")
+  String timeIntervalOnChange();
+
+  @DefaultMessage("30 seconds")
+  String timeInterval30Seconds();
+
+  @DefaultMessage("1 minute")
+  String timeInterval1Minute();
+
+  @DefaultMessage("5 minutes")
+  String timeInterval5Minutes();
+
+  @DefaultMessage("10 minutes")
+  String timeInterval10Minutes();
 }
--- a/appinventor/appengine/src/com/google/appinventor/client/editor/youngandroid/properties/YoungAndroidSensorTimeIntervalChoicePropertyEditor.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/editor/youngandroid/properties/YoungAndroidSensorTimeIntervalChoicePropertyEditor.java
@@ -17,11 +17,11 @@ import com.google.appinventor.client.widgets.properties.ChoicePropertyEditor.Cho
 public class YoungAndroidSensorTimeIntervalChoicePropertyEditor extends ChoicePropertyEditor {
   // sensor time interval choices
   private static final Choice[] timeIntervalChoices = new Choice[] {
-      new Choice("On change (0 ms)", "0"),
-      new Choice("30 seconds", "30000"),
-      new Choice("1 minute", "60000"),
-      new Choice("5 minutes", "300000"),
-      new Choice("10 minutes", "600000"),
+      new Choice(MESSAGES.timeIntervalOnChange(), "0"),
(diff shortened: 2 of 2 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

Sahitya0805:
Just following up on the PR I raised.
Looking forward to your feedback. Thanks!

jisqyv:

appinventor/appengine/src/com/google/appinventor/client/editor/youngandroid/properties/YoungAndroidSensorTimeIntervalChoicePropertyEditor.java
```diff
@@ -14,16 +14,14 @@
...
-    new Choice("0", "0"),
-    new Choice("1000", "1000"),
-    new Choice("10000", "10000"),
-    new Choice("60000", "60000"),
-    new Choice("300000", "300000"),
+      new Choice("On change (0 ms)", "0"),
+      new Choice("30 seconds", "30000"),
+      new Choice("1 minute", "60000"),
+      new Choice("5 minutes", "300000"),
+      new Choice("10 minutes", "600000"),
```
SusanRatiLane: These strings need to be internationalized. This is done by adding the strings to OdeMessages.java and referencing them through the Ode.MESSAGES object. There should be plenty of examples in the code. MESSAGES is already accessible through a parent class.

Sahitya0805:
I've updated the PR to internationalize the LocationSensor labels by adding them to OdeMessages.java and referencing them through the MESSAGES object. I have also verified that the client builds correctly with these changes.

Thanks for the feedback!

appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
```diff
@@ -16,5811 +16,5844 @@
  */
 @LocalizableResource.Generate(format = "com.google.gwt.i18n.rebind.format.PropertiesFormat")
 public interface OdeMessages extends Messages, ComponentTranslations {
-  // Used in multiple files
+    // Used in multiple files
```
SusanRatiLane, line 20: Looks like your editor added whitespace at the beginning of every line in this file. It needs to be reverted.

original_file.java.tmp
```diff
@@ -0,0 +1,249 @@
+// -*- mode: java; c-basic-offset: 2; -*-
```
SusanRatiLane: Are these files you renamed to save for yourself? Do you intend them to be part of your PR?

PR #3781, opened 2026-02-24
Improve PIC image processing on iOS

ewpatton:
Change-Id: I5eceabcf4323b26fbdf0ce360d8f88ade689e067

General items:

- [ ] My code follows the:

For all other changes:

What does this PR accomplish?

This PR fixes a bug in the ClassifyImageData block of PIC. Base64 encoding the image results in lines ending with "\r\n" (Windows) rather than the assumed "\n" (macOS/iOS/Unix), so I adjusted the logic to take that into account. I also made it so that images are centered vertically in the viewport rather than truncated at the bottom. I think this is more closely what people might expect given the target of any given photo is likely to be close to the center of the camera.

### DIFF
--- a/appinventor/components-ios/src/PersonalImageClassifier.swift
+++ b/appinventor/components-ios/src/PersonalImageClassifier.swift
@@ -12,7 +12,7 @@ fileprivate let PERSONAL_MODEL_PREFIX = "appinventor://personal-image-classifier
 @objc open class PersonalImageClassifier: BaseClassifier, LifecycleDelegate {
   public static let MODE_VIDEO = "Video"
   public static let MODE_IMAGE = "Image"
-  public static let IMAGE_WIDTH = 500
+  public static let IMAGE_WIDTH = 224
   public static let ERROR_INVALID_INPUT_MODE = -6;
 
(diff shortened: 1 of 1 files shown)


appinventor/components-ios/src/PersonalImageClassifier.swift
```diff
@@ -75,8 +75,8 @@ fileprivate let PERSONAL_MODEL_PREFIX = "appinventor://personal-image-classifier
       let width = PersonalImageClassifier.IMAGE_WIDTH
       let height = Int(image.size.height * CGFloat(PersonalImageClassifier.IMAGE_WIDTH) / image.size.width)
 
-      UIGraphicsBeginImageContext(CGSize(width: width, height: height))
-      image.draw(in: CGRect(x: 0, y: 0, width: width, height: height))
+      UIGraphicsBeginImageContext(CGSize(width: width, height: width))
```
jisqyv, line 78: height: width ??

ewpatton: Yes. In reality the variable should probably be renamed to size since the inputs are square.

ewpatton:
This is already in production for iOS so I am going ahead and merging it.

PR #3790, opened 2026-03-05
iOS: Gracefully handle unsupported components instead of crashing

Divyansh5070:
## Problem
When a project built in App Inventor contains a component that is not implemented in the iOS Companion (e.g. BluetoothServer, NxtDirectCommands), the Companion crashes with an unhandled picrin/YAIL error in `yail_make_instance()`.

## Root Cause
`yail_make_instance()` in `schemekit/src/yail.m` calls `pic_error()` (a fatal, unrecoverable error) when it cannot find an initializer for a requested class. If a class does not exist in the iOS binary at all, the symbol resolution step also fails silently, leading to a crash when instantiation is attempted.

## Fix
This PR implements a three-layer solution:

**1. `schemekit/src/yail.m`**
- Guard `yail_make_instance()` against `nil` class objects — returns `pic_undef_value` instead of calling `pic_error()`.
- Add a new `yail:class-available?` built-in function so the Scheme runtime can pre-check whether a component exists on iOS before trying to instantiate it.
- Post an `AIUnsupportedComponentNotification` via `NSNotificationCenter` so the UI layer can show a user-visible warning.

**2. `components-ios/src/runtime.scm`**
- Wrap the `make` helper with a `yail:class-available?` guard so unsupported components are skipped at the Scheme level.
- Guard `add-component-within-repl` so that `nil` (skipped) components are not registered or initialized.
- Add `show-unsupported-component-warning` to log and surface the warning to the Form layer.

**3. `components-ios/src/Form.swift`**
- Add `showUnsupportedComponentWarning(_:)` which presents a clear `UIAlertController` explaining which component is unsupported and that the rest of the project continues running.
- Register for `AIUnsupportedComponentNotification` in `viewDidLoad`.

## Result
**Before:** Loading a project with an Android-only component causes an immediate crash.

**After:** The Companion skips the unsupported component, shows the user:
> ⚠️ "BluetoothServer" is not available on iOS. The project will continue running without it.

…and continues executing the rest of the project normally.

---
*This is part of work toward the GSoC project idea: "Better behavior for unimplemented components in the iOS Companion".*

### DIFF
--- a/appinventor/components-ios/src/Form.swift
+++ b/appinventor/components-ios/src/Form.swift
@@ -188,6 +188,46 @@ let kMinimumToastWait = 10.0
     _tapGesture = UITapGestureRecognizer(target: self, action: #selector(HideKeyboard))
     _tapGesture?.cancelsTouchesInView = false
     view.addGestureRecognizer(_tapGesture!)
+    // Listen for unsupported-component notifications posted from the YAIL runtime.
+    NotificationCenter.default.addObserver(
+      self,
+      selector: #selector(handleUnsupportedComponentNotification(_:)),
+      name: NSNotification.Name("AIUnsupportedComponentNotification"),
+      object: nil)
+  }
+
+  /**
+   * Called by the YAIL runtime (via NSNotificationCenter) when a project
+   * contains a component that is not implemented on iOS.
+   *
+   * Instead of crashing, the Companion shows this informational alert so the
+   * user understands what happened and can continue using the rest of the project.
+   */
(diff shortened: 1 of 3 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

SusanRatiLane:

SusanRatiLane:
PR template was ignored: https://community.appinventor.mit.edu/t/google-summer-of-code-gsoc-2026/163196

PR #3795, opened 2026-03-07
Fix #3793: Treat disabled blocks as empty setter sockets

### LINKED ISSUE #3793: Setter blocks with disabled inputs should get an error icon
**Describe the bug**

[From the forum](https://community.appinventor.mit.edu/t/set-and-coerce-property/170047?u=ewpatton): Having a disabled block as an input to a property setter block does not trigger the empty input error check despite the fact that the code generation will result in an effectively empty input.

**Affects**

- [ ] Designer
- [x] Blocks editor
- [ ] Projects Explorer
- [ ] Android Companion
- [ ] iOS Companion
- [ ] iOS Compiled IPA
- [ ] Buildserver
- [ ] Debugging

**Expected behavior**

A setter block should gain an error message if its input is disabled, effectively treating it the same as if it is empty.

**Steps to reproduce**

Create a setter block, such as changing the background color of a button. Disable the color block. You will not see an error associated with the setter. However, if the color block is not connected, the setter will gain the "empty socket" error. Semantically, these are equivalent and the disabled block scenario ought to trigger the error as well.

Akashgite1:
General items:

- [ ] My code follows the:
    - [X] [Google JavaScript style guide](https://google.github.io/styleguide/jsguide.html) (for .js files)

- [ ] I branched from `ucr`

For all other changes:

What does this PR accomplish?

*Description*
Fixes an issue where disabled blocks connected to component property setter
sockets were incorrectly treated as valid inputs.

Previously, the `checkEmptySetterSocket` validation only checked whether a
block existed in the socket. However, disabled blocks still occupy the socket
even though they generate no code.

This change also checks `value.isEnabled()`, ensuring disabled blocks are
treated the same as empty sockets and the correct error is shown.

Fixes #3793 .

### DIFF
--- a/appinventor/blocklyeditor/src/warningHandler.js
+++ b/appinventor/blocklyeditor/src/warningHandler.js
@@ -747,7 +747,7 @@ Blockly.WarningHandler.prototype.checkDisposedBlock = function(block) {
 Blockly.WarningHandler.prototype['checkEmptySetterSocket'] = function(block) {
   if (block.setOrGet === 'set') {
     var value = block.getInputTargetBlock('VALUE');
-    if (!value || value.isEnabled()) {
+    if (!value || !value.isEnabled()) {
       block.setErrorIconText(Blockly.Msg.ERROR_PROPERTY_SETTER_NEEDS_VALUE);
       return true;
     }


AppInventorWorkerBee:

AppInventorWorkerBee:

SusanRatiLane:

appinventor/blocklyeditor/src/warningHandler.js
```diff
@@ -747,7 +747,7 @@ Blockly.WarningHandler.prototype.checkDisposedBlock = function(block) {
 Blockly.WarningHandler.prototype['checkEmptySetterSocket'] = function(block) {
   if (block.setOrGet === 'set') {
     var value = block.getInputTargetBlock('VALUE');
-    if (!value) {
+    if (!value || value.isEnabled()) {
```
jisqyv: Is this the correct logic? Shouldn't it be:

```java
if (!value || !value.isEnabled()) { ...
```

Or am I confused?

Akashgite1: Good catch, thanks!

You’re right — the condition should check for `!value.isEnabled()` so that disabled blocks are treated as empty inputs. I’ve pushed an update to fix the logic.

SusanRatiLane: > Good catch, thanks!
> 
> You’re right — the condition should check for `!value.isEnabled()` so that disabled blocks are treated as empty inputs. I’ve pushed an update to fix the logic.

This kind of error is a strong indicator that you did not build and test this before submitting.

We need proof you can build and run this repository. For more information, read: https://community.appinventor.mit.edu/t/google-summer-of-code-gsoc-2026/163196/9

Akashgite1: Thanks for the feedback.

I did build and run the App Inventor development server locally and tested the change in the Blocks editor.

Specifically, I verified that when a value block connected to a setter socket is disabled, the editor now correctly treats it as an empty input and shows the error indicator.

Here is a screenshot of the behavior after the fix:
Screenshot 2026-03-07 at 18-55-49 MIT App Inventor

Please let me know if there are additional scenarios you’d like me to test.

SusanRatiLane: OK, you could not possibly have done this testing before you submitted the PR because it would not have worked. 

We do not want any PRs that have not been built and tested before submission. Please keep this in mind for the future. If we find a strong indicator that you have submitted a PR without building and testing in the future, we will not be able to consider you for GSoC.

This includes any PRs you have already submitted that we have not reviewed. Be sure that all submitted PRs have been built and tested.

PR #3800, opened 2026-03-11
Fix Blocks Editor not updating when BlocksToolkit property is changed

### LINKED ISSUE #2596: Blocks Toolkit: Blocks Editor does not update when toolkit is changed
**Describe the bug**

In some cases, changing the Blocks Toolkit property on Screen1 does not cause the visible Built-In Blocks in the Blocks Editor to update. When the page is reloaded, the Blocks Editor will update.

**Affects**

- [ ] Designer
- [X] Blocks editor
- [ ] Companion
- [ ] Compiled apps
- [ ] Buildserver
- [ ] Debugging

**Expected behavior**

When the toolkit property is changed, the Designer view updates immediately. The Blocks Editor should also behave this way.

prabhasg5:
General items:

- [x] My code follows the:

- N/A — this change only affects the web-based Blocks Editor UI, not the companion or device runtime.

For all other changes:

### Problem

Changing the `BlocksToolkit` property on Screen1 in the Designer does not cause the Blocks Editor to update. The built-in blocks categories in the sidebar and the blocks shown in drawers remain stale until the page is reloaded.

### Root Cause

The root cause is that three separate caches hold onto the old toolkit state and none of them get invalidated when `BlocksToolkit` changes:

1. `BlocksEditor.onComponentPropertyChanged()` was a no-op, it simply ignored all property changes, so nothing downstream ever got notified.
2. `BlockSelectorBox` caches the built-in blocks tree in a `languageTreeItems` map that only gets populated once and is never cleared.
3. `drawer.js` caches `this.options.languageTree` on first use and never rebuilds it.

To fix this, I updated `onComponentPropertyChanged()` in `BlocksEditor.java` to detect `BlocksToolkit` changes and kick off the invalidation. It calls a new `clearBuiltInBlocksCache()` method on `BlockSelectorBox` to drop the Java-side cached tree, and a new `resetDrawerLanguageTree()` native method on `BlocklyPanel` that reaches into the JavaScript workspace (`workspace_svg.js`) and calls `resetLanguageTree()` on the drawer (`drawer.js`) to null out the cached block tree. After clearing both caches, it rebuilds the source structure explorer and closes any open drawer.

### Testing

1. Create a new project, switch to the Blocks Editor, note the available built-in blocks
2. Switch to Designer, select Screen1, change `BlocksToolkit`
3. Switch back to the Blocks Editor
4. Before this fix, blocks stay the same until page reload. 
5. After, they update immediately.

Fixes #2596.


AppInventorWorkerBee:

AppInventorWorkerBee:

SusanRatiLane:

SusanRatiLane:
This has not been tested. Attempts to change the Blocks Toolkit immediately threw errors about resetDrawerLanguageTree

SusanRatiLane:
Please see: https://community.appinventor.mit.edu/t/google-summer-of-code-gsoc-2026/163196

prabhasg5:
Thanks for reviewing and flagging the runtime error. **I didn’t encounter this error in my local testing, so I suspect there may be a difference in environment or execution timing.**
I’m investigating the resetDrawerLanguageTree issue and reproducing it in a clean environment.
I’ll fix the root cause and validate thoroughly before submitting an updated PR.
 I’ll return with a tested fix.

PR #3815, opened 2026-03-17
Fix long text wrapping issue in button layout

### LINKED ISSUE #462: Inconsistency for showing long text button in layout
When I am working on issue#410, i found an inconsistency between app inventor site ui viewer and the final view on phone. Basically, if I add a long text button (the length of text is longer than the width of screen) to a HorizontalAlignment, there will be an inconsistency for showing the view:

<div id="images">
app_inventor_screen
phone_view
</div>

As above shows, the left one is from app inventor ui viewer, the right one is from phone screen. I also uploaded an aia for testing on google group post: https://groups.google.com/forum/#!topic/app-inventor-dev/wcw6iKvAJrQ

jaykumar9022:
- [x] My code follows the:
    - [x] Google Java style guide

For all other changes:

This PR fixes an inconsistency where long text in buttons was not wrapping properly in the layout.

### Changes:
- Enabled multi-line text for buttons
- Prevented overflow of long text

### Result:
- Long text now wraps correctly inside buttons
- Improved UI consistency between designer and device view

Fixes #462

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/ButtonBase.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/ButtonBase.java
@@ -553,6 +553,9 @@ public abstract class ButtonBase extends TouchComponent<android.widget.Button>
   @SimpleProperty
   public void Text(String text) {
     TextViewUtil.setText(view, text);
+
+    view.setSingleLine(false);
+    view.setMaxLines(2);
   }
 
   /**


AppInventorWorkerBee:

AppInventorWorkerBee:

josmas:
I don't think hardcoding the settings are the way to fix this, mostly because it might affect existing apps which would now behave differently. This could maybe be done with a property to wrap text (or not), which is disabled by default?

josmas:
did not mean to close it.

ewpatton:
I second @josmas 's comments. I think this would be better as a property the user can control. We should also make sure the behavior in MockButtonBase also correctly reflects what's going on.

ewpatton:
Actually, looking at the original issue, I think the best course of action would be to keep the behavior the same in the components and instead make it so that the button wraps correctly in the mock UI.

PR #3829, opened 2026-03-19
Use Screen1 if lastOpened not in loaded project

ewpatton:

This fixes an issue where, if the last opened project property is present after a screen is deleted, we will default to Screen1 when loading the project.

*Description*

Previously, projects would still contain the last opened property if the current screen was deleted. This was patched, but there are still projects out there in an inconsistent state. This fixes it so that if we detect a project has been opened but the last opened entity is missing, we will default to Screen1.

*Testing Guidelines*

You can add a `lastopened` property to any project to point to a non-existent screen and it will fail to load in ai2. After this patch it should switch to Screen1.

**Context for the changes**

**General items:**

- [x] My code follows the:


jisqyv (approved):
LGTM

PR #3832, opened 2026-03-20
Remove tables from component palette

josmas:

*Description*

The current implementation of component palette renders as tables which is a problem for screen readers. This PR replaces the table based layout with html elements and css that is more a11y friendly.

*Testing Guidelines*
- Open the designer and confirm the palette categories collapse and expand
- Drag and drop as many components from the palette onto the canvas and confirm they drop correctly
- Type in the search box and confirm results appear (escape should clear the search)
- Check that there are no major changes in how the palette looks overall (there might be some minor padding/margin differences)
- In dev tools, the following two queries which in prod return `242` and `100`, should return `0`

`document.querySelectorAll('.ode-CollapsablePanel table').length`
`document.querySelectorAll('.ode-SimplePaletteItem table').length`

- [ ] I branched from `ucr`

For all other changes:

**General items:**

- [x] My code follows the:

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/editor/simple/palette/CollapsiblePaletteCategoryPanel.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/editor/simple/palette/CollapsiblePaletteCategoryPanel.java
@@ -1,0 +1,128 @@
+// -*- mode: java; c-basic-offset: 2; -*-
+// Copyright 2009-2011 Google, All Rights reserved
+// Copyright 2011-2026 MIT, All rights reserved
+// Released under the Apache License, Version 2.0
+// http://www.apache.org/licenses/LICENSE-2.0
+
+package com.google.appinventor.client.editor.simple.palette;
+
+import com.google.gwt.event.dom.client.ClickEvent;
+import com.google.gwt.event.dom.client.ClickHandler;
+import com.google.gwt.event.dom.client.KeyCodes;
+import com.google.gwt.event.dom.client.KeyDownEvent;
+import com.google.gwt.event.dom.client.KeyDownHandler;
+import com.google.gwt.user.client.ui.FlowPanel;
+import com.google.gwt.user.client.ui.Label;
+import com.google.gwt.user.client.ui.Widget;
+
+/**
+ * A collapsible panel for palette categories to replace GWT DisclosurePanel.
+ * Uses FlowPanel with CSS flexbox instead of table-rendering widgets.
+ */
+public class CollapsiblePaletteCategoryPanel extends FlowPanel {
+  private static int idCounter = 0;
+
+  private final FlowPanel header;
+  private final Label headerText;
+  private final FlowPanel content;
+  private final FlowPanel contentInner;
+  private boolean isOpen;
+
+  /**
+   * Creates a new collapsible palette category panel.
+   *
+   * @param categoryName the name of the category to display in the header
+   */
+  public CollapsiblePaletteCategoryPanel(String categoryName) {
+    String contentId = "ode-palette-cat-" + (idCounter++);
+
+    setStylePrimaryName("ode-PaletteCategory");
+
+    header = new FlowPanel();
+    header.setStylePrimaryName("ode-PaletteCategoryHeader");
+    header.getElement().setAttribute("role", "button");
+    header.getElement().setAttribute("tabindex", "0");
+    header.getElement().setAttribute("aria-expanded", "false");
+    header.getElement().setAttribute("aria-controls", contentId);
+
+    headerText = new Label(categoryName);
+    headerText.setStylePrimaryName("ode-PaletteCategoryText");
+    header.add(headerText);
+
+    content = new FlowPanel();
+    content.setStylePrimaryName("ode-PaletteCategoryContent");
+    content.getElement().setId(contentId);
+
+    contentInner = new FlowPanel();
+    contentInner.setStylePrimaryName("ode-PaletteCategoryContentInner");
+    contentInner.getElement().setAttribute("role", "list");
+    contentInner.getElement().setAttribute("aria-label", categoryName + " components");
+    content.add(contentInner);
+
+    header.addDomHandler(new ClickHandler() {
+      @Override
+      public void onClick(ClickEvent event) {
+        setOpen(!isOpen);
+      }
+    }, ClickEvent.getType());
+
+    header.addDomHandler(new KeyDownHandler() {
+      @Override
+      public void onKeyDown(KeyDownEvent event) {
+        int key = event.getNativeKeyCode();
+        if (key == KeyCodes.KEY_ENTER || key == 32 /* SPACE */) {
+          event.preventDefault();
+          setOpen(!isOpen);
+        }
+      }
+    }, KeyDownEvent.getType());
+
+    add(header);
+    add(content);
+
+    // Default to closed; callers control initial state via setOpen()
+    setOpen(false);
+  }
+
+  /**
+   * Sets whether the category is expanded or collapsed.
+   *
+   * @param open true to expand, false to collapse
+   */
+  public void setOpen(boolean open) {
+    isOpen = open;
+    header.getElement().setAttribute("aria-expanded", String.valueOf(open));
+    if (open) {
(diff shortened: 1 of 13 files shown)


appinventor/appengine/src/com/google/appinventor/client/editor/simple/palette/PaletteHelper.java
```diff
@@ -5,7 +5,7 @@
 
 package com.google.appinventor.client.editor.simple.palette;
 
-import com.google.gwt.user.client.ui.VerticalPanel;
+import com.google.gwt.user.client.ui.FlowPanel;
```
ewpatton: Is there a class higher up in the hierarchy that would make sense? Really we just care that this is a panel and it contains widgets, and not the specific implementation. That would give us flexibility in the future if we need it.

appinventor/appengine/src/com/google/appinventor/client/editor/youngandroid/palette/PaletteHelper.java
```diff
@@ -7,9 +7,9 @@
 
 import com.google.appinventor.client.editor.simple.palette.SimplePaletteItem;
 
-import com.google.gwt.user.client.ui.VerticalPanel;
+import com.google.gwt.user.client.ui.FlowPanel;
 
 interface PaletteHelper {
-  void addPaletteItem(VerticalPanel panel, SimplePaletteItem component);
+  void addPaletteItem(FlowPanel panel, SimplePaletteItem component);
```
ewpatton: See my earlier comment about specificity of the class here.

appinventor/appengine/src/com/google/appinventor/client/editor/simple/palette/CollapsiblePaletteCategoryPanel.java
ewpatton, line 1: Is there a commonality here with the refactor in the properties panel that would make sense to have in a shared superclass and/or reusable class?

appinventor/appengine/src/com/google/appinventor/client/editor/simple/palette/CollapsiblePaletteCategoryPanel.java
```diff
@@ -0,0 +1,128 @@
...
+  /**
+   * Sets whether the category is expanded or collapsed.
+   *
+   * @param open true to expand, false to collapse
+   */
+  public void setOpen(boolean open) {
+    isOpen = open;
+    header.getElement().setAttribute("aria-expanded", String.valueOf(open));
+    if (open) {
+      content.removeStyleName("ode-PaletteCategoryContent-closed");
```
ewpatton: I think rather than doing this you can do addDependentStyleName/removeDependentStyleName with "closed", and GWT attaches it to the primary style name (so effectively what this code does, but sensitive to if the primary CSS class changes).

appinventor/appengine/war/static/css/Ya.css
```diff
@@ -1040,36 +1040,43 @@ div.StatusPanel {
...
-  border: 1px solid #bbbbbb;
-  border-bottom: 0px;
-  margin-left: 0px;
+.ode-PaletteCategoryContent {
+  max-height: 5000px;
+  overflow: hidden;
+  transition: max-height 0.3s ease-out;
+}
+.ode-PaletteCategoryContent-closed {
+  max-height: 0 !important;
```
ewpatton: Any reason for max-height instead of just height here? Also, I'd prefer if we can not use `!important`. To me that indicates some logic in the CSS rules is broken somewhere.

ewpatton (changes requested):

appinventor/appengine/war/static/css/Ya.css
josmas: Yeah, I agree, the !important is not even needed, I'll remove it. The max-height is because we cannot animate height from auto to zero, but we can with max.

appinventor/appengine/src/com/google/appinventor/client/editor/simple/palette/PaletteHelper.java
josmas: I can use the InsertPanel interface instead.

appinventor/appengine/src/com/google/appinventor/client/editor/simple/palette/CollapsiblePaletteCategoryPanel.java
josmas: Yeah, there is. I wasn't sure if I wanted to abstract or treat them independently but the only main diff is that one uses an icon and the other doesn't. I'll figure out a way to abstract commonality out.

appinventor/appengine/src/com/google/appinventor/client/widgets/AbstractCollapsiblePanel.java
```diff
@@ -0,0 +1,123 @@
...
+  protected final FlowPanel header;
+  protected final FlowPanel content;
+  private boolean isOpen;
+
+  /**
+   * Sets up the shared header/content structure, ARIA attributes, and keyboard/click handlers.
+   * Subclasses must call {@link #finishInit()} after completing their own construction.
+   */
+  protected AbstractCollapsiblePanel() {
+    String contentId = "ode-collapsible-" + (idCounter++);
```
ewpatton: GWT also has a mechanism to do this. I'm not sure if we want to make use of that.

appinventor/appengine/src/com/google/appinventor/client/widgets/AbstractCollapsiblePanel.java
```diff
@@ -0,0 +1,123 @@
...
+      public void onClick(ClickEvent event) {
+        setOpen(!isOpen);
+      }
+    }, ClickEvent.getType());
+
+    header.addDomHandler(new KeyDownHandler() {
+      @Override
+      public void onKeyDown(KeyDownEvent event) {
+        int key = event.getNativeKeyCode();
+        if (key == KeyCodes.KEY_ENTER || key == 32 /* SPACE */) {
```
ewpatton: Maybe make a static final field called KEY_SPACE instead of putting it in a comment.

appinventor/appengine/src/com/google/appinventor/client/widgets/AbstractCollapsiblePanel.java
```diff
@@ -0,0 +1,123 @@
...
+      }
+    }, KeyDownEvent.getType());
+  }
+
+  /**
+   * Attaches header and content to the panel. Call at the end of the subclass constructor,
+   * after setting style names and populating the header.
+   */
+  protected final void finishInit() {
+    add(header);
```
ewpatton, line 60: Is there a reason this needs to be deferred?

appinventor/appengine/src/com/google/appinventor/client/widgets/AbstractCollapsiblePanel.java
ewpatton: This is the GWT version if we decide to use it: https://www.gwtproject.org/javadoc/latest/com/google/gwt/user/client/DOM.html#createUniqueId()

I have no preference.

appinventor/appengine/src/com/google/appinventor/client/widgets/AbstractCollapsiblePanel.java
josmas: Probably not. I'll try to add it at the end of the constructor and the subclasses don't have to worry about it.

appinventor/appengine/src/com/google/appinventor/client/widgets/AbstractCollapsiblePanel.java
josmas: I think it is worth changing or at least testing it, cause that method seems to generate unique ids across widgets. Will give it a go now.

josmas: Fun fact, the docs read: `Generates a unique DOM id. The id is of the form "gwt-id-".` but it actually is `"gwt-uid-"`. I couldn't find it after rebuilding...

ewpatton (approved):
LGTM

PR #3839, opened 2026-03-22
Allow reorg of components from tree (like alt+up and alt+down)

josmas:

*Description*
There is a shortcut (alt+up or alt+down) to move components in the tree and reorganise them. This PR adds mouse drag and drop with visual cues. See image:

dndtree

*Testing Guidelines*
See image above but basically just drag components in the tree to see them get reorganised in the designer. It is a UX improvement (in my opinion!).

**Context for the changes**

- [ ] I branched from `ucr`

For all other changes:

**General items:**

- [x] My code follows the:

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockComponent.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockComponent.java
@@ -1417,11 +1417,6 @@ public abstract class MockComponent extends Composite implements PropertyChangeL
         return;
       }
 
-      // Non-visible components must remain at the root (Screen) level
-      if (!source.isVisibleComponent() && !targetContainer.isRoot()) {
-        return;
-      }
-
       // Determine target container and insertion index
       MockContainer targetContainer;
       int insertIndex;
@@ -1440,6 +1435,11 @@ public abstract class MockComponent extends Composite implements PropertyChangeL
         }
         int targetIdx = targetContainer.getChildren().indexOf(target);
         insertIndex = (position < 0) ? targetIdx : targetIdx + 1;
+      }
+
+      // Non-visible components must remain at the root (Screen) level
+      if (!source.isVisibleComponent() && !targetContainer.isRoot()) {
+        return;
       }
 
       // Handle coordinate properties when moving to/from AbsoluteArrangement
--- a/appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockComponent.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockComponent.java
@@ -1417,6 +1417,11 @@ public abstract class MockComponent extends Composite implements PropertyChangeL
         return;
       }
 
+      // Non-visible components must remain at Form (Screen) level
+      if (source instanceof MockNonVisibleComponent && !(targetContainer instanceof MockForm)) {
+        return;
+      }
+
       // Determine target container and insertion index
       MockContainer targetContainer;
       int insertIndex;
@@ -1435,11 +1440,6 @@ public abstract class MockComponent extends Composite implements PropertyChangeL
         }
         int targetIdx = targetContainer.getChildren().indexOf(target);
         insertIndex = (position < 0) ? targetIdx : targetIdx + 1;
-      }
-
-      // Non-visible components must remain at Form (Screen) level
-      if (source instanceof MockNonVisibleComponent && !(targetContainer instanceof MockForm)) {
-        return;
       }
 
       // Handle coordinate properties when moving to/from AbsoluteArrangement
--- a/appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockComponent.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockComponent.java
@@ -357,52 +357,7 @@ public abstract class MockComponent extends Composite implements PropertyChangeL
     COMPONENT_DATABASE = editor.getComponentDatabase();
     componentDefinition = COMPONENT_DATABASE.getComponentDefinition(type);
 
-    sourceStructureExplorerItem = new SourceStructureExplorerItem() {
-      @Override
-      public void onSelected(NativeEvent source) {
-        // are we showing the blocks editor? if so, toggle the component drawer
-        if (Ode.getInstance().getCurrentFileEditor() instanceof BlocksEditor) {
-          BlocksEditor<?, ?> blocksEditor =
-              (BlocksEditor<?, ?>) Ode.getInstance().getCurrentFileEditor();
-          LOG.info("Showing item " + getName());
-          blocksEditor.showComponentBlocks(getName());
-        } else {
-          select(source);
-        }
-      }
-
-      @Override
-      public void onStateChange(boolean open) {
-        // The user has expanded or collapsed the branch in the components tree corresponding to
-        // this component. Remember that by setting the expanded field so that when we re-build
-        // the tree, we will keep the branch in the same state.
-        expanded = open;
-      }
-
-      @Override
-      public boolean canRename() {
-        return !isRoot();
-      }
-
-      @Override
-      public void rename() {
-        if (!isRoot()) {
-          new RenameDialog(getName()).center();
-        }
-      }
-
-      @Override
-      public boolean canDelete() {
-        return !isRoot();
-      }
-
-      @Override
-      public void delete() {
-        if (!isRoot()) {
-          new DeleteDialog().center();
-        }
-      }
-    };
+    sourceStructureExplorerItem = new ComponentExplorerItem();
     expanded = true;
 
     // Create a default property set for the component
@@ -904,6 +859,12 @@ public abstract class MockComponent extends Composite implements PropertyChangeL
     };
     itemNode.setUserObject(sourceStructureExplorerItem);
 
+    // Mark non-root items as draggable for tree DnD reordering
+    if (!isRoot()) {
+      itemNode.getElement().setAttribute("draggable", "true");
+      itemNode.getElement().setAttribute("data-name", getName());
+    }
+
     // Set alt text on the icon image for accessibility
     // Use Scheduler to defer until DOM is attached
     final String altText = getName() + " " + getVisibleTypeName();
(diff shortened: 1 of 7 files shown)


Akashgite1:
This is a really nice UX improvement — drag-and-drop in the component tree feels much more intuitive compared to the current workflow.

One thing I’m curious about:
- How does this behave with nested arrangements (e.g., Vertical/Horizontal layouts)? Does it always preserve correct parent-child relationships?
- Is the change immediately reflected in the companion, or are there any sync delays/issues?
- Does this impact undo/redo behavior?

Overall, this definitely improves usability a lot .

josmas:
Hi @Akashgite1 the core functionality was already there by using shortcuts, so I am reusing that. This PR adds only UI changes to allow for the dnd behaviour, but refreshing the companion and so forth, that I have not touched.

appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockComponent.java
```diff
@@ -1363,4 +1324,178 @@ public native boolean shouldCancel(Event event)/*-{
...
+      return MockComponent.this;
+    }
+
+    @Override
+    public void onSelected(NativeEvent source) {
+      if (Ode.getInstance().getCurrentFileEditor() instanceof BlocksEditor) {
+        BlocksEditor<?, ?> blocksEditor =
+            (BlocksEditor<?, ?>) Ode.getInstance().getCurrentFileEditor();
+        LOG.info("Showing item " + getName());
+        blocksEditor.showComponentBlocks(getName());
```
ewpatton, line 1354: Not for this PR, but I would really like to adjust the logic here at some point because right now it seems ass backward how the component has to have awareness about what editor is available. Instead, it feels like the active editor should be listening for these events and update its view accordingly.

appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockComponent.java
```diff
@@ -1363,4 +1324,178 @@ public native boolean shouldCancel(Event event)/*-{
...
+      } else {
+        select(source);
+      }
+    }
+
+    @Override
+    public void onStateChange(boolean open) {
+      expanded = open;
```
ewpatton: As implemented, `expanded` is part of the component state but is really specific to the tree state. We may want to think about as we refactor here. Technically what is done is most semantically similar to what was there, but both don't really seem to correctly model what is happening.

appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockComponent.java
```diff
@@ -1363,4 +1324,178 @@ public native boolean shouldCancel(Event event)/*-{
...
+    }
+
+    @Override
+    public boolean canDrag() {
+      return !isRoot();
+    }
+
+    @Override
+    public boolean isContainer() {
+      return MockComponent.this instanceof MockContainer;
```
ewpatton: I would recommend that we have this as `false` and then override it in `MockContainer` rather than doing an instanceof test here.

ewpatton: Never mind, my brain didn't parse that we were in a nested inner class. Please disregard (but I'm still unconvinced we should be using an instanceof here.)

appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockComponent.java
josmas: I see that. I can try and move the state to the explorer instead. We are also clearing and recreating the whole tree for every change, which is not great, but that's a much bigger change and might not be worth it given the number of elements for a tree in an average app, so I'll just deal with the state for now.

appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockComponent.java
josmas: I can add some methods that return true/false instead. Let me try that.

josmas:
@ewpatton moving tree state to explorer meant adding a few methods to check for visibility of components so added "isInitiallyExpanded". Not terribly sure how useful this is, but didn't want users to have a different experience on first load, so replicated it.

appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockComponent.java
josmas: I don't think this is a trivial refactor so opened https://github.com/mit-cml/appinventor-sources/issues/3882

appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockContainer.java
```diff
@@ -56,6 +56,11 @@ public abstract class MockContainer extends MockVisibleComponent implements Drop
    *
    * @param editor  editor of source file the component belongs to
    */
+  @Override
+  public final boolean isContainer() {
+    return true;
+  }
+
```
ewpatton: AI slop?

appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockComponent.java
```diff
@@ -904,6 +869,12 @@ protected Focusable getFocusable() {
     };
     itemNode.setUserObject(sourceStructureExplorerItem);
 
+    // Mark non-root items as draggable for tree DnD reordering
+    if (!isRoot()) {
+      itemNode.getElement().setAttribute("draggable", "true");
+      itemNode.getElement().setAttribute("data-name", getName());
```
ewpatton, line 875: I think this is okay for now, but in the future we may want to consider using the "uuid" instead.

appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockComponent.java
```diff
@@ -1363,4 +1334,183 @@ public native boolean shouldCancel(Event event)/*-{
...
+        }
+        int targetIdx = targetContainer.getChildren().indexOf(target);
+        insertIndex = (position < 0) ? targetIdx : targetIdx + 1;
+      }
+
+      // Non-visible components must remain at Form (Screen) level
+      if (source instanceof MockNonVisibleComponent && !(targetContainer instanceof MockForm)) {
+        return;
+      }
+
```
ewpatton, line 1444: It seems like this check could go earlier.

appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockContainer.java
josmas: This was my attempt to remove the instanceof check in the previous version

appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockComponent.java
ewpatton, line 1444: For future proofing, we should probably have the if check `source.isVisibleComponent() && !targetContainer.isRoot()` since the root objects of the IOT and Alexa UIs are by definition not MockForm and the check as written would fail.

appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockContainer.java
ewpatton: Sure, but it's inserted between a Javadoc and the thing it documents, and not in a place that would make sense following the order that the style guide recommends.

appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockComponent.java
josmas: yeah, this can go before the block on top. Will push.

appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockContainer.java
josmas: Oh, I see, yeah. I am also not sure I like it at all. It's the same thing but more code. I would prefer to revert the change altogether. Thoughts?

josmas: I've pushed the modified version in any case.

ewpatton (changes requested):
A small issue I noticed, and I don't know if it is worth addressing, but one can drag a visible component below a non-visible component. I also encountered another issue, possibly related, where using the Alt+Down shortcut caused an internal error due to an index out of bounds exception.

josmas:
I had some initial feedback that dragging non visible components is a positive change, as it allows users to organise the tree as they want to use it.

I'll have a dig for the exception though.

josmas:
@ewpatton the array issue is pre-existing. New project, drag in 1 button, select button in the tree, alt+down not only generates an error, but it calls delete on the component. It seems to be trying to use indexing coming out of a hash and things go haywire. I can reuse code from this PR to fix it, unless you rather have a different PR. I'll push the changes in any case.

josmas:
The force push was due to the rebase with master

ewpatton:
> I had some initial feedback that dragging non visible components is a positive change, as it allows users to organise the tree as they want to use it.

My issue is not with reordering of non-visible components, but rather that visible components can be placed in the tree after non-visible components, which doesn't really make much sense as they are two distinct classes of objects and aren't mixed in the designer view (non-visible components are always at the bottom).

> @ewpatton the array issue is pre-existing. New project, drag in 1 button, select button in the tree, alt+down not only generates an error, but it calls delete on the component. It seems to be trying to use indexing coming out of a hash and things go haywire. I can reuse code from this PR to fix it, unless you rather have a different PR. I'll push the changes in any case.

It's fine to just fix it in this PR.

josmas:
Just seeing this. Sure, I can guard for non-visible comps. The other issue is already fixed in this branch.

josmas:
@ewpatton non-viz components are now locked at the bottom. I also had to change the visual indicator otherwise it looked like it was broken if you tried to move any viz component under a non-viz one.

ewpatton:
@josmas Thanks. Taking a look.

ewpatton (approved):
LGTM

PR #3845, opened 2026-03-24
Fix SpeechRecognizer lifecycle to prevent overlapping sessions and in…

### LINKED ISSUE #3825: Restart in SpeechRecognizer component
**Describe the bug**

In order to implement continuous automatic speech recognition with the `SpeechRecognizer` component, it is necessary to *restart* the component with `GetText`. That is not always reliable. There are several behaviors related to  two `SpeechRecognizer` timeouts: an utterance timeout (UT &#x2248;a few hundred ms) and a recognizer timeout (RT &#x2248;a few s).

- After UT (when `partial` in `AfterGettingText` is `false`) restarting a `SpeechRecognizer` component with `GetText` sometimes works, but depending when speech actually occurs, the indicator *beeps* can garble speech, or short utterances can get lost, or the RT is not restarted.
- After RT (which seems to vary unpredictably), a global error (trapped by `Screen1.ErrorOccured`) occurs: 3806 *No Match*.
- After stopping a `SpeechRecognizer` component with `Stop`, a global error (trapped by `Screen1.ErrorOccured`) occurs: 3802 *Client Side Error*.

Android v16 | AI2 Companion v2.78

**Affects**

- [ ] Designer
- [ ] Blocks editor
- [ ] Projects Explorer
- [ ] Android Companion
- [ ] iOS Companion
- [ ] iOS Compiled IPA
- [ ] Buildserver
- [ ] Debugging
- [X] Other... (please describe)

Events generated by `SpeechRecognizer` components; 

**Expected behavior**

This `AfterGettingText` event indicates a typical use case:

Image

- As long as `partial` is `true`, display the current transcription appended with `result` &mdash; which keeps updating.
- Once `partial` is `false`, update the current transcription appended with `result` and restart the the `SpeechRecognizer` component with `GetText`.

- I expected the `SpeechRecognizer` to continuously transcribe speech until `Stop` is called on that `SpeechRecognizer` &mdash; or something else&hellip; a `Stopped` event or `isRunning` property?
- The most important expectation is that users can reliably understand and implement the protocol to generate multi-utterance transcriptions with a `SpeechRecognizer` component.

**Steps to reproduce**

- Repeatedly call `GetText` on a `SpeechRecognizer` component. 
- Call `Stop` on a `SpeechRecognizer` component.

Akashgite1:

*Description*
This PR fixes lifecycle/state management issues in the Speech Recognizer component

Previously:
- Multiple GetText() calls could start overlapping recognition sessions
- Calling Stop() when the recognizer was not active could lead to errors
- The recognizer state was not reset consistently after completion or error
- Rapid repeated calls could cause unstable behavior

This PR introduces a minimal internal state (IDLE / LISTENING) to ensure valid transitions and prevent overlapping sessions. A small debounce is also added to guard against rapid repeated calls.

*Testing Guidelines*
The changes were tested using the MIT App Inventor Companion with the following scenarios:
- Rapid repeated GetText() calls (no overlapping sessions observed)
- Continuous recognition loop (restart after final result works correctly)
- Normal speech flow across multiple utterances
- Calling Stop() during active and inactive states
- Silence / no speech scenarios (handled gracefully without crashes)
- Repeated cycles (5–10+) without degradation or stuck states
- No crashes or unexpected behavior were observed.

Fixes #3825 .

**Context for the changes**

This change affects the runtime component (SpeechRecognizer), but it is a minimal internal bug fix with no API or user-facing changes, so targeting the master branch is appropriate instead of ucr.

- [ ] I branched from `ucr`

For all other changes:

**General items:**

- [X] My code follows the:
    - [X] Indentation has been doubled checked
    - [X] This PR does not include unnecessary changes such as formatting or white space
- [ ] `ant tests` passes on my machine master branch.

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/SpeechRecognizer.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/SpeechRecognizer.java
@@ -69,6 +69,15 @@ public class SpeechRecognizer extends AndroidNonvisibleComponent
   private boolean useLegacy = true;
 
   private String language = "";
+  
+  // --- Lifecycle state management ---
+  private enum RecognizerState {
+    IDLE,
+    LISTENING
+  }
+
+  private RecognizerState state = RecognizerState.IDLE;
+  private long lastStartTime = 0;
 
   /**
    * Creates a SpeechRecognizer component.
@@ -148,9 +157,25 @@ public class SpeechRecognizer extends AndroidNonvisibleComponent
       });
       return;
     }
+    
+    long now = System.currentTimeMillis();
+
+    // debounce rapid calls
+    if (now - lastStartTime < 300) {
+      return;
+    }
+
+    // prevent overlapping sessions
+    if (state != RecognizerState.IDLE) {
+      return;
+    }
+    lastStartTime = now;
+
     BeforeGettingText();
     speechRecognizerController.addListener(this);
     speechRecognizerController.start();
+
+    state = RecognizerState.LISTENING;
   }
 
   /**
@@ -161,9 +186,15 @@ public class SpeechRecognizer extends AndroidNonvisibleComponent
    */
   @SimpleFunction
   public void Stop() {
+    if (state == RecognizerState.IDLE) {
+      return;
+    }
+
     if (speechRecognizerController != null) {
       speechRecognizerController.stop();
     }
+
+    state = RecognizerState.IDLE;
   }
 
   /**
@@ -203,6 +234,7 @@ public class SpeechRecognizer extends AndroidNonvisibleComponent
    */
   @Override
   public void onResult(String text) {
+    state = RecognizerState.IDLE;
     result = text;
     AfterGettingText(result, false);
   }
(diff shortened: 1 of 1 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

ewpatton:

appinventor/components/src/com/google/appinventor/components/runtime/SpeechRecognizer.java
```diff
@@ -212,6 +244,12 @@ public void onResult(String text) {
    */
   @Override
   public void onError(int errorNumber) {
+    state = RecognizerState.IDLE;
+
+    // Ignore NO_MATCH (not a real failure)
+    if (errorNumber == 3806) {
+      return;
+    }
```
ewpatton: I'm worried about the change in semantics here. If this error now occurs, the app developer neither receives an AfterGettingText event nor a dispatched error. I think we should leave this out and have the focus of the PR be just on the IDLE/LISTENING state management.

Akashgite1: I agree that ignoring 3806 changes the existing behaviour because the app developer would receive neither AfterGettingText nor an error event.

I’ve removed the special handling for 3806 and kept the PR focused on the IDLE/LISTENING lifecycle state management to prevent overlapping recognition sessions.

PR #3847, opened 2026-03-24
Fix: Resolve extension property descriptions in Designer using metada…

### LINKED ISSUE #3569: Description text is not showing in the designer section for the extension's properties.
**Describe the bug**

The description text is showing for built-in components perfectly. See below:

Image

But it's not showing for extension properties, it always just shows a default message!

Image

**Affects**

- [x] Designer
- [ ] Blocks editor
- [ ] Projects Explorer
- [ ] Companion
- [ ] Compiled apps
- [ ] Buildserver
- [ ] Debugging

**Expected behavior**

It should show the description text from the extension properties, if available.

**Steps to reproduce**

Akashgite1:

*Description*

This PR fixes an issue where property descriptions for extension components are not displayed correctly in the Designer. Instead of showing human-readable text, raw translation keys (e.g., `Component.PropertyDescriptions`) are displayed.

The root cause is that the current implementation assumes all property descriptions exist in the translation table, which is not true for extension components. Extension descriptions are provided via metadata but are not passed through the existing pipeline.

This PR introduces a safe fallback mechanism:
- First, attempt to resolve the description using the existing translation lookup (unchanged behavior)
- If lookup fails, fallback to the description provided by the extension metadata

To maintain backward compatibility, an overloaded `addProperty` method is introduced without modifying existing call sites.

This ensures correct behavior for both built-in and extension components.

*Testing Guidelines*

1. Start App Inventor locally (`ant run`)
2. Create a new project
3. Add an extension component (.aix file)
4. In Designer, hover over a property of the extension:
   - Before fix: shows raw key (e.g., `Component.PropertyDescriptions`)
   - After fix: shows readable description
5. Verify built-in components still display correct property descriptions

Fixes #3569 .

**Context for the changes**
This change only affects the Designer (appengine) and does not impact the companion or runtime behavior.

- [X] I have made no changes that affect the companion
- [ ] I branched from `ucr`

For all other changes:

**General items:**

- [X] My code follows the:
    - [X] Indentation has been doubled checked
    - [X] This PR does not include unnecessary changes such as formatting or white space

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockComponent.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockComponent.java
@@ -623,6 +623,45 @@ public abstract class MockComponent extends Composite implements PropertyChangeL
     properties.addProperty(name, defaultValue, ComponentTranslationTable.getPropertyName(caption),
         ComponentTranslationTable.getCategoryName(category),  propertyDesc, editor, propertyType, editorType, editorArgs);
   }
+  
+  /**
+   * Extended version of addProperty to support extension metadata descriptions.
+   * Falls back to metadata when translation lookup fails.
+   */
+  public final void addProperty(String name, String defaultValue, String caption, String category,
+                               String editorType, String[] editorArgs, String description, PropertyEditor editor) {
+
+    String propertyDesc = ComponentTranslationTable.getPropertyDescription(name 
+      + "PropertyDescriptions");
+
+    // Step 1: component-specific lookup
+    if (propertyDesc.equals(name + "PropertyDescriptions")) {
+      propertyDesc = ComponentTranslationTable.getPropertyDescription(
+          (type.equals("Form") ? "Screen" : type) + "." + name + "PropertyDescriptions");
+    }
+
+    // Step 2: fallback for extensions
+    if ((propertyDesc.equals(name + "PropertyDescriptions") ||
+        propertyDesc.equals((type.equals("Form") ? "Screen" : type) + "." + name + "PropertyDescriptions"))
+        && description != null) {
+
+      propertyDesc = description;
+    }
+
+    int propertyType = EditableProperty.TYPE_NORMAL;
+    if (!isPropertyPersisted(name)) {
+      propertyType |= EditableProperty.TYPE_NONPERSISTED;
+    }
+    if (!isPropertyVisible(name)) {
+      propertyType |= EditableProperty.TYPE_INVISIBLE;
+    }
+    if (isPropertyforYail(name)) {
+      propertyType |= EditableProperty.TYPE_DOYAIL;
+    }
+
+    properties.addProperty(name,defaultValue,ComponentTranslationTable.getPropertyName(caption),
+        ComponentTranslationTable.getCategoryName(category), propertyDesc, editor, propertyType, editorType, editorArgs);
+  }
 
   /**
    * Returns the component name.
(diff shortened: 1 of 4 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

Akashgite1:
To preserve backward compatibility and avoid modifying existing call sites, I introduced an overloaded method specifically for passing extension metadata descriptions.

jewelshkjony:
I don't think there's any need to add a new `addProperty` method. Instead, we can adapt the existing method to include these changes. I've already updated the original method to match your suggested changes. (Please see this PR: https://github.com/Akashgite1/appinventor-sources/pull/1)

Akashgite1:
> I don't think there's any need to add a new `addProperty` method. Instead, we can adapt the existing method to include these changes. I've already updated the original method to match your suggested changes. (Please see this PR: [Akashgite1#1](https://github.com/Akashgite1/appinventor-sources/pull/1))

Thanks for the suggestion!

You're absolutely right that this can be solved by updating the existing addProperty method instead of introducing a new one. I’ve updated the implementation accordingly and modified all the relevant call sites.

Initially, I introduced an overloaded method to avoid potentially breaking existing API usages and to keep the change minimal. However, I agree that maintaining duplicate methods could introduce unnecessary complexity and long-term maintenance issues.

So I’ve removed the additional method and refactored everything to use a single updated method with the fallback logic.

Please take a look and let me know if anything else should be adjusted

Akashgite1:
Tested locally — confirmed that both built-in components and extension components are now displaying property descriptions correctly in the Designer.

@ewpatton @SusanRatiLane could you please review this when you have time and let me know if any changes are needed?

Screenshot 2026-03-25 at 01-51-04 MIT App Inventor

Akashgite1:
Hi @SusanRatiLane, can you please review these changes at your convenience? Please let me know if any changes are required.

appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockComponent.java
```diff
@@ -602,16 +599,24 @@ public final void addProperty(String name, String defaultValue, String caption,
...
       + "PropertyDescriptions");
     if (propertyDesc.equals(name + "PropertyDescriptions")) {
       propertyDesc = ComponentTranslationTable.getPropertyDescription((type.equals("Form")
           ? "Screen" : type) + "." + propertyDesc);
     }
+    // Fallback for extension metadata
+    if ((propertyDesc.equals(name + "PropertyDescriptions") ||
+        propertyDesc.equals((type.equals("Form") ? "Screen" : type) + "." + name + "PropertyDescriptions"))
+        && description != null) {
+
```
jewelshkjony, line 617: There is no need for this new line here.

appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockComponent.java
```diff
@@ -626,7 +631,7 @@ public final void addProperty(String name, String defaultValue, String caption,
     properties.addProperty(name, defaultValue, ComponentTranslationTable.getPropertyName(caption),
         ComponentTranslationTable.getCategoryName(category),  propertyDesc, editor, propertyType, editorType, editorArgs);
   }
-
+
```
jewelshkjony, line 634: Unnecessary white space!

jewelshkjony:
I tested it locally and it works as expected. It seems fine to me.

image

image

PR #3856, opened 2026-03-26
DataFile working for iOS

### LINKED ISSUE #3764: DataFile not implemented in iOS

DataFile component, which read csv files, is not implemented in iOS

@bobparks is there any data science aia files that uses DataFile component that I can use for testing?

daki7711:
PR that makes DataFile Component work on iOS devices

Resolves #3764

### DIFF
--- a/appinventor/components-ios/src/ChartData2D.swift
+++ b/appinventor/components-ios/src/ChartData2D.swift
@@ -35,9 +35,9 @@ import DGCharts
   @objc func DoesEntryExist(_ x: String, _ y: String) -> Bool{
     /* Original:
      DispatchQueue.main.sync {
-      var pair: YailList<AnyObject> = [x, y]
-      return self._chartDataModel!.doesEntryExist(pair) // TODO: is the ! okay?
-    }*/
+     var pair: YailList<AnyObject> = [x, y]
+     return self._chartDataModel!.doesEntryExist(pair) // TODO: is the ! okay?
+     }*/
 
     let group = DispatchGroup()
     group.enter()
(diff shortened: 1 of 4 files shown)


appinventor/components-ios/src/ChartData2D.swift
```diff
@@ -35,9 +35,9 @@ import DGCharts
   @objc func DoesEntryExist(_ x: String, _ y: String) -> Bool{
     /* Original:
      DispatchQueue.main.sync {
-      var pair: YailList<AnyObject> = [x, y]
-      return self._chartDataModel!.doesEntryExist(pair) // TODO: is the ! okay?
-    }*/
+     var pair: YailList<AnyObject> = [x, y]
+     return self._chartDataModel!.doesEntryExist(pair) // TODO: is the ! okay?
+     }*/
```
ewpatton: We may just want to remove this commented out code.

ewpatton:
This is up on TestFlight.

appinventor/components-ios/src/ChartData2D.swift
SusanRatiLane: This is commented-out code. Do we want to remove it?

daki7711: it's been a while since I worked on this branch, but based on what I see this was just some space changes, which means these code was already there before. So my guess is we shouldn't remove it. @ewpatton

SusanRatiLane: OK, I guess remove the whitespace change? We should probably decide what to do about that code, but it shouldn't hold up this PR.

SusanRatiLane (approved):
LGTM

PR #3861, opened 2026-03-27
Add drag-and-drop media asset helper block creation in Blocks workspace

### LINKED ISSUE #2412: Drag and drop media file to create asset helper block

Add the assets helper block to the workspace when the user drags a file from the media container (in the blocks tab, not the designer) to the blocks workspace.

Requested in [this forum post](https://community.appinventor.mit.edu/t/what-is-your-opinion-about-helper-blocks-in-app-inventor-gsoc-project-user-feedback/9057/59?u=beksomega).

Dependent on #2301 

Currently the only way to access the asset helper block is to open the flyout for a component that uses assets, and find the asset block there. This feature would make it easier to access the helper block (although the behavior is somewhat hidden).

App Inventor does address this use case, the feature request just makes it more convenient.

This feature is not specific to App Inventor's educational mission, but would benefit all users of App Inventor (including those using it for educational purposes).

**Next steps**
These are just the first steps that come to mind when I think about completing this project. They may be incomplete, or there may be better ways to implement things =)

1) #2301 needs to be merged before this issue can be worked on.
2) Some sort of drag-and-drop listener will need to be added to the workspace. This behavior could possibly be modeled off the logic for detecting when someone drops a .png representing exported blocks ([source](https://github.com/mit-cml/appinventor-sources/blob/cc121a481d7edd75878f24c3ccac70fa8584daa8/appinventor/blocklyeditor/src/workspace_svg.js#L123)).
3) The assets helper block will need to be added to the workspace. The easiest way to do this is probably using the XML system, eg using [blockTypeToXML](https://github.com/mit-cml/appinventor-sources/pull/2301/files#diff-6e4fb241d9ba8de835db1a6304f1d2d5e3beb8bf388ab62374c24af897580c81R42), and then deserializing the resulting XML using [domToWorkspace](https://developers.google.com/blockly/reference/js/Blockly.Xml#.domToWorkspace).

srijabal:

WhatsApp GIF 2026-03-28 at 03 20 46

*Description*

Implements issue #2412 by allowing users to drag a media file from the Media pane (Blocks view) into the workspace to create an asset helper block.

This PR:
- Adds drag-and-drop handling for media assets in the Blocks workspace.
- Creates a `helpers_assets` block on drop with the dragged asset name selected.
- Preserves existing `.png` workspace import drag/drop behavior.
- Improves drag UX for media items (grab/grabbing cursor and cleaner drag feedback).

*Testing Guidelines*

1. Run App Inventor locally and open/create a project.
2. Go to the **Blocks** tab.
3. Upload a file in the **Media** section.
4. Drag the media item from the Media list into the blocks workspace.
5. Verify an asset helper block is created and uses the dropped file name.
6. Verify standard block drag/drop behavior is unchanged.
7. Verify dragging a `.png` blocks export into the workspace still imports blocks.
8. (Optional) Repeat in both Classic and Neo themes.

Resolves #2412.

**Context for the changes**

- [ ] I branched from `ucr`

For all other changes:

**General items:**

- [x] My code follows the [Google Java style guide](https://google.github.io/styleguide/javaguide.html) (for .java files)
- [x] My code follows the [Google JavaScript style guide](https://google.github.io/styleguide/jsguide.html) (for .js files)
- [ ] My code follows the [Swift style guide](https://google.github.io/swift/) (for .swift files)

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/AssetList.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/AssetList.java
@@ -146,10 +146,10 @@ public class AssetList extends Composite implements ProjectChangeListener {
         // Add the name to the tree. We need to enclose it in a span
         // because the CSS style for selection specifies a span.
         String assetName = node.getName();
-        String nodeName = assetName;
-        if (nodeName.length() > 20)
-          nodeName = nodeName.substring(0, 8) + "..." + nodeName.substring(nodeName.length() - 9,
-              nodeName.length());
+        String displayName = assetName;
+        if (displayName.length() > 20)
+          displayName = displayName.substring(0, 8) + "..." +
+              displayName.substring(displayName.length() - 9, displayName.length());
 
         String fileSuffix = node.getProjectId() + "/" + node.getFileId();
         String treeItemText = "<span style='cursor: grab'>";
@@ -160,7 +160,7 @@ public class AssetList extends Composite implements ProjectChangeListener {
         } else if (StorageUtil.isVideoFile(fileSuffix )) {
           treeItemText += new Image(images.mediaIconVideo());
         }
-        treeItemText += nodeName + "</span>";
+        treeItemText += displayName + "</span>";
         final HTML treeItemWidget = new HTML(treeItemText);
         final TreeItem treeItem = new TreeItem(treeItemWidget);
         // keep a pointer from the tree item back to the actual node
--- a/appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/AssetList.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/AssetList.java
@@ -20,6 +20,7 @@ import com.google.appinventor.shared.rpc.project.youngandroid.YoungAndroidAssetN
 import com.google.appinventor.shared.rpc.project.youngandroid.YoungAndroidAssetsFolder;
 import com.google.appinventor.shared.rpc.project.youngandroid.YoungAndroidProjectNode;
 import com.google.appinventor.shared.storage.StorageUtil;
+import com.google.gwt.dom.client.Element;
 import com.google.gwt.event.dom.client.ClickEvent;
 import com.google.gwt.event.dom.client.ClickHandler;
 import com.google.gwt.event.dom.client.MouseMoveEvent;
@@ -137,13 +138,14 @@ public class AssetList extends Composite implements ProjectChangeListener {
       for (ProjectNode node : assetsFolder.getChildren()) {
         // Add the name to the tree. We need to enclose it in a span
         // because the CSS style for selection specifies a span.
-        String nodeName = node.getName();
+        String assetName = node.getName();
+        String nodeName = assetName;
         if (nodeName.length() > 20)
           nodeName = nodeName.substring(0, 8) + "..." + nodeName.substring(nodeName.length() - 9,
               nodeName.length());
 
         String fileSuffix = node.getProjectId() + "/" + node.getFileId();
-        String treeItemText = "<span style='cursor: pointer'>";
+        String treeItemText = "<span style='cursor: grab'>";
         if (StorageUtil.isImageFile(fileSuffix)) {
           treeItemText += new Image(images.mediaIconImg());
         } else if (StorageUtil.isAudioFile(fileSuffix )) {
@@ -156,6 +158,7 @@ public class AssetList extends Composite implements ProjectChangeListener {
         // keep a pointer from the tree item back to the actual node
         treeItem.setUserObject(node);
         assetList.addItem(treeItem);
+        configureDraggable(treeItem.getElement(), assetName);
       }
     }
   }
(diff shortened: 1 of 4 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

SusanRatiLane:

ewpatton (changes requested):
I was not able to replicate the behavior in the GIF on macOS. I tried Chrome, Firefox, and Safari, and in all browsers the context menu opened instead of initiating a drag.

srijabal:
Thanks for checking, @ewpatton 

The issue was that the media menu was getting triggered before the drag started (due to event order).

I’ve separated the behavior now:

Click → opens the media menu (same as before)
Drag → starts dragging without opening the menu

I also made drag detection slightly more robust across browsers.

Tested on macOS again, works as expected. Attaching an updated GIF, could you please check once on your side?


app_inventor_fix

appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/AssetList.java
```diff
@@ -137,13 +145,14 @@ private void refreshAssetList() {
       for (ProjectNode node : assetsFolder.getChildren()) {
         // Add the name to the tree. We need to enclose it in a span
         // because the CSS style for selection specifies a span.
-        String nodeName = node.getName();
+        String assetName = node.getName();
+        String nodeName = assetName;
```
ewpatton: I don't see why having a second variable was necessary here.

srijabal: I think both variables are needed here. assetName keeps the full file name for drag and drop, while displayName is only used for the shortened name shown in the UI. I renamed the variable to make the code clearer and made a new commit. Thanks for pointing it out.

appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/AssetList.java
```diff
@@ -207,6 +245,104 @@ public void onProjectNodeRemoved(Project project, ProjectNode node) {
...
 
+  private static native void configureDraggable(Element el, String assetName)/*-{
+    function setCursor(root, cursor) {
+      root.style.cursor = cursor;
+      if (root.querySelectorAll) {
+        var descendants = root.querySelectorAll('*');
+        for (var i = 0; i < descendants.length; i++) {
+          descendants[i].style.cursor = cursor;
+        }
+      }
```
ewpatton: This seems really inefficient to me. Why are we not implementing this using CSS?

appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/AssetList.java
```diff
@@ -207,6 +245,104 @@ public void onProjectNodeRemoved(Project project, ProjectNode node) {
...
+    function setAssetDragName(name) {
+      $wnd.__aiAssetDragName = name;
+      try {
+        if ($wnd.top) {
+          $wnd.top.__aiAssetDragName = name;
+        }
+      } catch (err) {
+        // Ignore cross-frame access errors. The data transfer payload is primary.
+      }
+    }
```
ewpatton: Given that the Blockly code/panel is always in `top`, there should never be a case where this can fail (if so, something would have gone wrong sooner.)

appinventor/appengine/src/com/google/appinventor/client/explorer/youngandroid/AssetList.java
```diff
@@ -207,6 +245,104 @@ public void onProjectNodeRemoved(Project project, ProjectNode node) {
...
+    el.addEventListener('dragstart', function(e) {
+      var name = this.getAttribute('data-assetname');
+      this.classList.add('ode-AssetItem-dragging');
+      this.setAttribute('data-ai-assetdragging', 'true');
+      setCursor(this, 'grabbing');
+      e.dataTransfer.effectAllowed = 'move';
+      e.dataTransfer.clearData();
+      e.dataTransfer.setData('application/x-appinventor-asset', name);
+      e.dataTransfer.setData('text/plain', name);
+      setAssetDragName(name);
```
ewpatton: We could probably inline the assignment given my comments above.

appinventor/blocklyeditor/src/workspace_svg.js
```diff
@@ -67,65 +68,302 @@ Blockly.WorkspaceSvg.prototype.blocksNeedingRendering = null;
...
+var hasDragType = function(dataTransfer, type) {
+  if (!dataTransfer || !dataTransfer.types) {
+    return false;
+  }
+  if (typeof dataTransfer.types.indexOf == 'function') {
+    return dataTransfer.types.indexOf(type) >= 0;
+  }
+  if (typeof dataTransfer.types.contains == 'function') {
+    return dataTransfer.types.contains(type);
+  }
```
ewpatton: I don't understand why this logic is here. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/types), the `types` field is just an array of strings, so the properties you test for should always be available.

PR #3871, opened 2026-04-02
Change email query to use emaillower filter

barreeeiroo:

*Description*

Fixes issue in `findUserByEmail` which caused to not filter by the lowercase email fields, probably due to a typo. This should mimic both the `getUser` and `getUserFromEmail` functions, which do indeed have fallback to the actual lowercase version.

*Testing Guidelines*

N/A

**Context for the changes**

- [ ] I branched from `ucr`

For all other changes:

**General items:**

- [x] My code follows the:

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/server/storage/ObjectifyStorageIo.java
+++ b/appinventor/appengine/src/com/google/appinventor/server/storage/ObjectifyStorageIo.java
@@ -2107,7 +2107,7 @@ public class ObjectifyStorageIo implements StorageIo {
     // get the first one. we don't expect this to happen
     UserData userData = datastore.query(UserData.class).filter("email", inputemail).get();
     if (userData == null) {     // Mixed case didn't work, try lower case
-      userData = datastore.query(UserData.class).filter("email", email).get();
+      userData = datastore.query(UserData.class).filter("emaillower", email).get();
       if (userData == null) {
         throw new NoSuchElementException("Couldn't find a user with email " + inputemail);
       }


appinventor/appengine/src/com/google/appinventor/server/storage/ObjectifyStorageIo.java
```diff
@@ -2107,7 +2107,7 @@ public String findUserByEmail(String inputemail) throws NoSuchElementException {
     // get the first one. we don't expect this to happen
     UserData userData = datastore.query(UserData.class).filter("email", inputemail).get();
     if (userData == null) {     // Mixed case didn't work, try lower case
-      userData = datastore.query(UserData.class).filter("email", email).get();
+      userData = datastore.query(UserData.class).filter("emaillower", email).get();
```
josmas, line 2110: email is already lowercased in line 2104. It is an odd sequence of lines and unfortunate naming but it seems like it is already doing what it is supposed to do. The first check is with inputemail and this one with inputemail.toLowerCase()

barreeeiroo: Hmm, no... So the `email` field was the original field which may hold uppercase emails. The `emaillower` is the actual field which holds the lowercased emails. Without the change, we will be comparing `email` (which is always lowercase) against `email` (which may be uppercase).

See the access pattern in lines 293 and 295. Either we run the non-lowercase against `email`, or the lowercase variant against `emaillower`. See also `createUser` in lines 368 to to 381: we only sanitize `emaillower`.

I found out about this bug in Kodular while doing the migration out of GCP. I was expecing this method to match users, which didn't because it was comparing against the non-lowercase option.

josmas: yeah, you are correct, it is the naming that confused me again. If the var was emaillower, it would be more clear, but yeah, I also checked the other functions you pointed to and it is the same pattern, just with better naming! :D

josmas (approved):
LGTM, thanks!

appinventor/appengine/src/com/google/appinventor/server/storage/ObjectifyStorageIo.java
barreeeiroo: Yeah, I guess that to be correct, the first `inputemail` should be the original input, and not the lowercase one...

barreeeiroo: Well no, actually this PR is correct...

* `email` matched against `inputemail` (original input, both may be uppercase)
* `emaillower` matched against `email` (lowercase input, both are always lowercase)

Really bad naming, yes...

josmas: Yeah, the PR is correct. The names are not but we can leave them as is.

PR #3880, opened 2026-04-06
Fix screen dropdown caption inconsistency on hover

Deepupandey-ui:
### Problem
The screen dropdown caption becomes inconsistent when hovering/opening the dropdown. 
This happens because the dropdown is rebuilt, but the caption is not updated accordingly, 
leading to stale or incorrect UI state.

### Fix
Added a call to update the dropdown caption after rebuilding the screen dropdown items:
setDropDownButtonCaption(WIDGET_NAME_SCREENS_DROPDOWN, currentProject.currentScreen);

This ensures that the dropdown caption always reflects the current active screen.

### Testing
- Verified logic ensures caption sync after dropdown rebuild
- Prevents stale UI state during hover/open

### Note
Could not fully test in local environment due to setup complexity,
but fix is based on correct UI state synchronization logic.

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/editor/youngandroid/DesignToolbar.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/editor/youngandroid/DesignToolbar.java
@@ -283,6 +283,7 @@ public class DesignToolbar extends Toolbar {
             screen.screenName, new SwitchScreenAction(projectId, screen.screenName),
             new Image(Ode.getImageBundle().form())));
       }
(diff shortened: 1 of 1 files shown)


Deepupandey-ui:
Hi, thanks for reviewing.

Could you please share feedback on why this PR was closed? 
I would like to improve the approach and contribute better.

Thank you!

SusanRatiLane:
> Hi, thanks for reviewing.
> 
> Could you please share feedback on why this PR was closed? I would like to improve the approach and contribute better.
> 
> Thank you!

We do not accept PRs that were not built and tested locally.

PR #3885, opened 2026-04-11
Reconcile Companion saved apps and connect button spacing with iOS

zikunz:

*Description*

This PR resolves issue #3855 by reconciling the Android Companion with iOS in the two areas @dcpetty called out, plus a supporting change to the bridge that feeds the saved apps screen.

**1. Connect screen spacing.**

The iOS layout places visible breathing room between the QR button and the saved apps button. Android currently jams them together with an 8 unit spacer, so the saved apps button looks like part of the connect button cluster and is easy to mis tap. This PR increases `Spacer4` from 8 to 48 in both `Screen1.scm` (the JSON form descriptor used by the IDE and build) and `Screen1.yail` (the Scheme code executed by the Companion at runtime). Both files have to move together because `Screen1.yail` is what the Kawa runtime actually evaluates, while `Screen1.scm` is consulted only by the IDE and build pipeline.

**2. Saved apps screen.**

The original `projects.html` was a bare HTML table with no styling, no search, no sort, and no date column. It also concatenated project file names into `innerHTML`, which is a real XSS sink. This PR rewrites the file to match the iOS App Library affordances. The new screen shows a `Saved Apps` title, a search input, a sort dropdown with three options (`Recently Modified`, `App Name: A-Z`, `App Name: Z-A`), a per project date label, and an empty state with text matching iOS character for character (`No Apps Downloaded` followed by `Connect to your project first to download it to your library!`). `App Name: A-Z` and `App Name: Z-A` reuse the same base wording iOS uses for its sort actions. The first label reads `Recently Modified` rather than iOS's `Recently Opened` for the same reason the date label reads `Last modified` (see below). All user controlled data now flows through `textContent` and `createElement`, so a project file named `` renders as plain text instead of executing. Sort preference is persisted under the namespaced `localStorage` key `projects_sortMode`.

**3. `ProjectsActivity.listProjects()` bridge.**

The original method dereferenced `getExternalFilesDir()` without a null check and called `list()` on the result, which itself returns null when the directory is inaccessible. Neither null was guarded. This PR adds null guards for both, switches from `list()` to `listFiles()` to obtain `File` objects with metadata, filters out hidden files and non files (which should never appear in `__projects__` but the original code would happily return them), and changes the return shape from `["name1", "name2"]` to `[{"name": "...", "lastModified": <long>}]` so `projects.html` can render the date column. `JSONObject.put` is wrapped in try and catch for the checked `JSONException`. A repository wide grep confirms `projects.html` is the only consumer of this bridge, so the contract change is safe.

**Why CheckBox1 stays in MainPanel.**

The iOS layout places `Use Legacy Connection` between the QR button and the library button. On Android, `CheckBox1` cannot move into `PhonePanel` because `Screen1.yail` line 37 sets `PhonePanel.Visible` to false whenever `PhoneStatus1.isDirect()` returns true. That happens in emulator mode and in USB direct mode. The Legacy Connection toggle controls WebRTC behavior and must remain reachable in those modes. The 48 unit spacer is the closest safe approximation of the iOS section break that does not regress the emulator path.

**Why the date label says Last modified.**

iOS tracks `lastOpened` via `SystemVariables.lastOpenedTable` persisted in UserDefaults. Android currently has no equivalent per project usage tracker. Adding one would require new SharedPreferences plumbing in `ProjectManager` and an update path in `loadProject`, which is out of scope for this PR. The Android side reads `File.lastModified` and labels it `Last modified` so the UI is honest about the data source. A follow up PR can add true `lastOpened` tracking on Android.

**Other iOS App Library features deferred to follow up.**

This PR explicitly does not port per app icons, the version mismatch warning row, or success and failure toast messages on delete.

*Testing Guidelines*

Tested on a physical Motorola Android device running the APK built from this branch, compared against a Companion built from `origin/ucr` for the before state. All 8 verification steps below passed. The emulator regression test was run on an Android emulator. The iOS Companion was also built and run on an iPhone 16 Simulator (iOS 18.6) to visually confirm that the Android changes align with the iOS layout and saved apps UX. Before and after screenshots are attached below.

**Before (origin/ucr):**

before-connect-screen
before-saved-apps-table
before-saved-apps-xss-alert

The alert dialog shows `1` -- this is `alert(1)` firing from a project file named ``. The dialog text is in Chinese because the test device's system language is Chinese. The new `projects.html` renders this filename as plain text instead of executing it.

**After (this branch):**

after-connect-screen
after-project-list
after-empty-state
after-xss-safe
after-emulator-regression

1. Build the Companion APK with Java 11. Run `JAVA_HOME=/opt/homebrew/Cellar/openjdk@11/11.0.30/libexec/openjdk.jdk/Contents/Home ant PlayApp` from the `appinventor` directory. Output is `appinventor/build/buildserver/MIT AI2 Companion.apk`.

2. Install on a physical Android device. Run `adb install -r "appinventor/build/buildserver/MIT AI2 Companion.apk"`.

3. On the connect screen, confirm the saved apps button has a clear vertical gap above it instead of touching the QR button. Compare against a Companion built from `origin/ucr` for the before state.

4. Open the saved apps screen with no projects downloaded. On a physical phone tap the saved apps button on the connect screen. On an emulator open the action bar overflow menu (three dots, top right) and tap `Load Project`, since `PhonePanel` is hidden when `PhoneStatus1.isDirect()` returns true. Confirm the empty state shows the bold heading `No Apps Downloaded` followed by the description `Connect to your project first to download it to your library!`.

5. From a connected ai2.appinventor.mit.edu session, save 3 or more projects to the library. Reopen the saved apps screen. Verify each project shows a name and a relative date. Verify the sort dropdown switches between `Recently Modified`, `App Name: A-Z`, and `App Name: Z-A`. Verify the search input filters in real time and is case insensitive. Verify Run loads the project. Verify Delete prompts for confirmation before removing the project from the list.

6. XSS regression. Save a project whose name contains HTML metacharacters such as `` and confirm the name renders as plain text instead of an image element.

7. Emulator regression. Run on an emulator and confirm `PhonePanel` is still hidden, the `Running using the emulator or USB` text is shown, and the `Use Legacy Connection` checkbox remains visible.

8. Run `JAVA_HOME=/opt/homebrew/Cellar/openjdk@11/11.0.30/libexec/openjdk.jdk/Contents/Home ant tests` and confirm the Java suites pass locally. On this machine all JUnit suites pass. The `ios-build.xml` sub-target needs an iPhone simulator destination and is skipped on a non-Mac iOS toolchain, which is unrelated to this PR.

Resolves #3855.

**Context for the changes**

- [x] I branched from `ucr`

For all other changes:

**General items:**

- [ ] My code follows the:


josmas:
Hi @zikunz can the legacy button be on its own panel? Is it the case that if you are connecting directly or through emulator, the saved apps button does not show either?

In two minds about how much sense it makes that saved apps is available in direct mode. What do you think?

AppInventorWorkerBee:

AppInventorWorkerBee:

zikunz:
> Hi @zikunz can the legacy button be on its own panel? Is it the case that if you are connecting directly or through emulator, the saved apps button does not show either?
> 
> In two minds about how much sense it makes that saved apps is available in direct mode. What do you think?

Hi @josmas, thanks for looking at this.

Yes. The legacy checkbox can have its own panel under MainPanel, which keeps it visible regardless of the connection mode. I am happy to do that if you prefer.

Yes. LibraryButton is a child of PhonePanel, so it is hidden along with that panel in direct and emulator modes. The only remaining way to reach the saved apps screen is the Load Project entry in the action bar overflow menu, which opens the same ProjectsActivity that LibraryButton would open.

I would lean toward making saved apps directly reachable in direct mode. Loading a previously downloaded project does not require a live phone connection. The iOS Companion keeps its library button visible regardless of the connection mode. Hiding the saved apps button on Android feels more like a side effect of where it sits in the tree than a deliberate design choice, and a developer in direct or emulator mode might still want to open a saved project without going through the overflow menu.

I am happy to move LibraryButton up into MainPanel and give CheckBox1 its own panel in a single follow-up commit if you agree with both. Please let me know how you would like me to proceed.

PR #3887, opened 2026-04-11
Improve Component Hydration

barreeeiroo:

When a ChartData2D, Trendline, or Map feature (Marker, LineString, Polygon, etc.) is added to its parent container through any path other than drag-and-drop (for example, double-click in palette or copy-and-paste), the required post-add initialization (`addToChart`/`addToMap`) never runs. This means the component is added to the container's children list but never actually attached to the underlying view, so the Chart does not render the data series/legend and Map features do not appear on the map until the page is refreshed.

The fix adds a `protected onChildAdded(MockComponent)` hook to `MockContainer` (no-op by default), overridden in `MockChart` and `MockMap` to call the appropriate initialization. This centralizes the logic so every `addComponent` call path goes through it. The now-redundant explicit calls in the layout `onDrop` methods and `MockChart.addDataFile()` are removed.

Canvas (Ball/ImageSprite) is not affected; those components have no post-add initialization step.

*Testing Guidelines*

1. Open a project with a Chart component.
2. Double-click ChartData2D in the palette while the Chart is selected. Verify the chart immediately renders the data series and legend without needing a page refresh.
3. Drag a ChartData2D from the palette onto the Chart. Verify it still works as before.
4. Copy a ChartData2D from one Chart and paste it into another. Verify the target chart updates.
5. Open a project with a Map component.
6. Double-click Marker in the palette while the Map is selected. Verify the marker appears on the map immediately.
7. Drag a Marker onto the Map. Verify it still works as before.
8. Save and reload a project containing Charts with data and Maps with features. Verify everything renders correctly on load.

**Context for the changes**

For all other changes:

**General items:**

- [x] My code follows the:

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockChart.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockChart.java
@@ -333,7 +333,6 @@ public final class MockChart extends MockContainer {
       // TODO: More data component support
       MockChartData2D data = new MockChartData2D(editor);
       addComponent(data);
-      data.addToChart(this);
 
       // Change the properties of the instantiated data component
       data.changeProperty("DataFileYColumn", column);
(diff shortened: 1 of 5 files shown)


josmas (approved):
LGTM, thanks!

PR #3888, opened 2026-04-12
Move Image READ_EXTERNAL_STORAGE to method-level on Picture setter

zikunz:
*Description*
The Image component declares `READ_EXTERNAL_STORAGE` as a class-level `@UsesPermissions` annotation, which causes the permission to appear in the manifest for every app that includes an Image component, even when the image is loaded from bundled assets and no storage access is needed.

~On Android 13+, `READ_EXTERNAL_STORAGE` no longer grants access to media files. The runtime code in `TiramisuUtil.requestImagePermissions` already requests `READ_MEDIA_IMAGES` on Android 13+, but `READ_MEDIA_IMAGES` was never declared in any annotation, so the system denies the permission request and the image cannot be loaded from external storage.~

**Update:** After feedback from Anke on the [community forum thread](https://community.appinventor.mit.edu/t/the-image-component-incorrectly-requires-read-permission/171831/5), the `READ_MEDIA_IMAGES` portion was removed from this PR. Google Play has required prior approval for `READ_MEDIA_IMAGES` declarations since May 2025, so adding it to the manifest would cause compliance issues. This PR now focuses solely on moving `READ_EXTERNAL_STORAGE` from class-level to method-level.

This PR makes ~two changes~ one change to `Image.java`:

1. Moves `READ_EXTERNAL_STORAGE` from the class-level `@UsesPermissions` to a method-level annotation on the `Picture` property setter. This follows the pattern established by ~`Player.java`~ `VideoPlayer.java` for its `Source` property. With this change, the permission only appears in the manifest when the `Picture` property is actually used (set in the Designer or via blocks).

2. ~Adds `READ_MEDIA_IMAGES` to the same method-level annotation for Android 13+ compatibility. This makes the manifest declaration match the runtime behavior that `TiramisuUtil` already implements.~

PR #2773 addresses a related but different aspect of this issue. That PR refines the runtime logic for when to request the permission. This PR addresses the manifest-level declaration and does not touch the runtime logic.

~*Testing Guidelines*~

1. ~Build the project and verify `simple_components_build_info.json` shows `READ_EXTERNAL_STORAGE` ~and `READ_MEDIA_IMAGES`~ under `conditionals.permissions.Picture` rather than in the top-level `permissions` array.~

2. ~Build a test APK containing an Image component with `Picture` set to an external storage path via blocks (e.g., `file:///storage/emulated/0/Download/photo.jpg`). Install on an Android 14 (API 34) emulator. Click the button to load the image. The system should display a `READ_MEDIA_IMAGES` permission dialog. After granting, the image should load successfully.~

3. ~Repeat step 2 using an APK built from the unmodified `ucr` branch. The app should display "Error 908: The permission READ_MEDIA_IMAGES has been denied" because the permission is not declared in the manifest.~

4. ~Build the Companion APK (`ant PlayApp`). Verify the Companion launches and runs without errors.~

**Updated Testing Guidelines**

1. Build the project and verify `simple_components_build_info.json` shows `READ_EXTERNAL_STORAGE` under `conditionals.permissions.Picture` rather than in the top-level `permissions` array.

2. Build a test APK containing an Image component with `Picture` set to an external storage path via blocks (e.g., `file:///storage/emulated/0/Download/photo.jpg`). Install on an Android 14 (API 34) emulator. Verify the behavior is unchanged from the `ucr` branch.

3. Build the Companion APK (`ant PlayApp`). Verify the Companion launches and runs without errors.

**Test setup**

The test project contains a Button and an Image component. The Button's Click handler sets `Image1.Picture` to `file:///storage/emulated/0/Download/mitappinventor.png` (a file pushed to the emulator's external storage via `adb push`). The Image's `Picture` property is not set in the Designer. Both APKs were built using a local buildserver and installed on an Android 14 (API 34) emulator.

**Manifest comparison**

Before (ucr branch, no fix):
```
uses-permission: name='android.permission.INTERNET'
uses-permission: name='android.permission.READ_EXTERNAL_STORAGE'
```

After (this branch):
```
uses-permission: name='android.permission.INTERNET'
uses-permission: name='android.permission.READ_EXTERNAL_STORAGE'
```

The manifest is the same when the `Picture` property is used. The improvement is visible when `Picture` is NOT used -- in that case, `READ_EXTERNAL_STORAGE` no longer appears in the manifest.

> Note: The before/after screenshots below were taken with an earlier revision that included `READ_MEDIA_IMAGES`. That permission has since been removed from this PR per community feedback. The current revision does not change behavior for apps that use the `Picture` property -- the improvement is that apps that do NOT use `Picture` no longer get `READ_EXTERNAL_STORAGE` in their manifest.

**Earlier test (with READ_MEDIA_IMAGES, since removed):**

~`READ_MEDIA_IMAGES` is now present in the manifest because the `Picture` block is used in the project, triggering the conditional permission.~

**Before (ucr branch, no fix):**

**After (this branch):**

Relates to #2700.

Reported on the community forum: https://community.appinventor.mit.edu/t/the-image-component-incorrectly-requires-read-permission/171831

**Context for the changes**

- [x] I branched from `ucr`

For all other changes:

**General items:**

- [ ] My code follows the:

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/Image.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/Image.java
@@ -6,7 +6,7 @@
 
 package com.google.appinventor.components.runtime;
 
-import static android.Manifest.permission.READ_EXTERNAL_STORAGE;
+import android.Manifest;
 
 import com.google.appinventor.components.annotations.Asset;
 import com.google.appinventor.components.annotations.DesignerComponent;
@@ -51,7 +51,6 @@ import java.io.IOException;
     "Designer or in the Blocks Editor.",
     iconName = "images/image.png")
 @SimpleObject
-@UsesPermissions(permissionNames = "android.permission.INTERNET")
 public final class Image extends AndroidViewComponent {
 
   private final ImageView view;
@@ -149,7 +148,7 @@ public final class Image extends AndroidViewComponent {
   @DesignerProperty(editorType = PropertyTypeConstants.PROPERTY_TYPE_ASSET,
       defaultValue = "")
   @SimpleProperty
-  @UsesPermissions(READ_EXTERNAL_STORAGE)
+  @UsesPermissions({Manifest.permission.INTERNET, Manifest.permission.READ_EXTERNAL_STORAGE})
   public void Picture(@Asset String path) {
     final String tempPath = path == null ? "" : path;
     if (TiramisuUtil.requestImagePermissions(container.$form(), path,
--- a/appinventor/components/src/com/google/appinventor/components/runtime/Image.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/Image.java
@@ -7,7 +7,6 @@
 package com.google.appinventor.components.runtime;
 
 import static android.Manifest.permission.READ_EXTERNAL_STORAGE;
-import static android.Manifest.permission.READ_MEDIA_IMAGES;
 
 import com.google.appinventor.components.annotations.Asset;
 import com.google.appinventor.components.annotations.DesignerComponent;
@@ -150,7 +149,7 @@ public final class Image extends AndroidViewComponent {
   @DesignerProperty(editorType = PropertyTypeConstants.PROPERTY_TYPE_ASSET,
       defaultValue = "")
   @SimpleProperty
-  @UsesPermissions({READ_EXTERNAL_STORAGE, READ_MEDIA_IMAGES})
+  @UsesPermissions(READ_EXTERNAL_STORAGE)
   public void Picture(@Asset String path) {
     final String tempPath = path == null ? "" : path;
     if (TiramisuUtil.requestImagePermissions(container.$form(), path,
--- a/appinventor/components/src/com/google/appinventor/components/runtime/Image.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/Image.java
@@ -6,7 +6,9 @@
 
 package com.google.appinventor.components.runtime;
 
-import android.Manifest;
+import static android.Manifest.permission.READ_EXTERNAL_STORAGE;
+import static android.Manifest.permission.READ_MEDIA_IMAGES;
+
 import com.google.appinventor.components.annotations.Asset;
 import com.google.appinventor.components.annotations.DesignerComponent;
 import com.google.appinventor.components.annotations.DesignerProperty;
@@ -50,8 +52,7 @@ import java.io.IOException;
     "Designer or in the Blocks Editor.",
     iconName = "images/image.png")
 @SimpleObject
-@UsesPermissions(permissionNames = "android.permission.INTERNET," +
-    "android.permission.READ_EXTERNAL_STORAGE")
+@UsesPermissions(permissionNames = "android.permission.INTERNET")
 public final class Image extends AndroidViewComponent {
 
   private final ImageView view;
@@ -149,6 +150,7 @@ public final class Image extends AndroidViewComponent {
   @DesignerProperty(editorType = PropertyTypeConstants.PROPERTY_TYPE_ASSET,
       defaultValue = "")
   @SimpleProperty
+  @UsesPermissions({READ_EXTERNAL_STORAGE, READ_MEDIA_IMAGES})
   public void Picture(@Asset String path) {
     final String tempPath = path == null ? "" : path;
     if (TiramisuUtil.requestImagePermissions(container.$form(), path,
--- a/appinventor/components/src/com/google/appinventor/components/runtime/Image.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/Image.java
@@ -7,7 +7,6 @@
 package com.google.appinventor.components.runtime;
 
 import android.Manifest;
-
 import com.google.appinventor.components.annotations.Asset;
 import com.google.appinventor.components.annotations.DesignerComponent;
 import com.google.appinventor.components.annotations.DesignerProperty;


zikunz:
After posting this PR, Anke pointed out on the [community forum thread](https://community.appinventor.mit.edu/t/the-image-component-incorrectly-requires-read-permission/171831/5) that Google has required prior approval for `READ_MEDIA_IMAGES` declarations since May 2025. This means adding `READ_MEDIA_IMAGES` to the manifest (as this PR does) could cause issues for Play Store submissions.

There is a tradeoff here. Without `READ_MEDIA_IMAGES` in the manifest, compiled apps get Error 908 and cannot load images from external storage on Android 13+. With it, the image loading works but the app may need a Google Play permissions declaration form.

If the preferred direction is to remove all storage permission declarations from the component annotations entirely and leave permission handling to the app developer, I am happy to update this PR accordingly.

zikunz:
Updated based on community feedback from [Anke's response](https://community.appinventor.mit.edu/t/the-image-component-incorrectly-requires-read-permission/171831/5).

Changes in this revision:

- Removed `READ_MEDIA_IMAGES` from the method-level `@UsesPermissions` annotation. Google Play has required prior approval for `READ_MEDIA_IMAGES` declarations since May 2025, so adding it to the manifest would cause Play Store compliance issues for users.
- The annotation now declares only `READ_EXTERNAL_STORAGE` on the `Picture` setter, matching the existing pattern in `VideoPlayer.java` (line 199) where `@UsesPermissions(READ_EXTERNAL_STORAGE)` is used on the `Source` setter without `READ_MEDIA_VIDEO`.

The primary value of this PR remains: `READ_EXTERNAL_STORAGE` moves from an unconditional class-level declaration to a conditional method-level declaration on the `Picture` setter. Apps that use Image without setting Picture from external storage will no longer get an unnecessary permission in their manifest.

josmas:

AppInventorWorkerBee:

AppInventorWorkerBee:

appinventor/components/src/com/google/appinventor/components/runtime/Image.java
```diff
@@ -6,7 +6,8 @@
 
 package com.google.appinventor.components.runtime;
 
-import android.Manifest;
+import static android.Manifest.permission.READ_EXTERNAL_STORAGE;
```
josmas: This works but the import seem inconsistent, using a string for Internet and a constant for storage. I think the webviewer uses this slightly differently. Please have a look.

josmas:
Hey @zikunz thanks for the PR. Please have a look at the comment above.
Also wondering if we still need Internet at the class level; are there any other properties that would need it other than Picture?

appinventor/components/src/com/google/appinventor/components/runtime/Image.java
zikunz: I get it! I switched to `import android.Manifest` with `Manifest.permission.X` constants, matching the `WebViewer.java` pattern. I also moved INTERNET from class level to the `Picture` setter since that is the only property that needs it.

zikunz:
> Hey @zikunz thanks for the PR. Please have a look at the comment above. Also wondering if we still need Internet at the class level; are there any other properties that would need it other than Picture?

Hi @josmas, thank you for the review! I looked at `WebViewer.java` and switched to the `Manifest.permission.X` style with a regular `import android.Manifest` to keep the imports consistent.

For the INTERNET question, I went through every property and method in Image. Only the `Picture` setter performs network access (through `MediaUtil.getBitmapDrawable`, which can fetch URLs). The other properties are all local view operations. I moved INTERNET from the class level to the `Picture` setter alongside `READ_EXTERNAL_STORAGE` so both permissions are conditional on `Picture` being used. Please let me know if you would prefer keeping INTERNET at the class level instead.

appinventor/components/src/com/google/appinventor/components/runtime/Image.java
```diff
@@ -7,6 +7,7 @@
 package com.google.appinventor.components.runtime;
 
 import android.Manifest;
+
```
josmas: do we need the whitespace here?

josmas (dismissed):
LGTM, thanks!

josmas:
@zikunz thanks for the PR. This one looks good to me but I think we might have a related issue with READ_MEDIA_IMAGES as you pointed out earlier. Could you have a look into the call to TiramisuUtil.java for permissions? it looks like it will silently fail to load external images for newer devices (and likely other components are broken too like the videoplayer).

appinventor/components/src/com/google/appinventor/components/runtime/Image.java
zikunz: Thank you for pointing this out, removed in 5b7021f2effd2d9cd4a8d1662cac2abac72c59c4.

zikunz:
> @zikunz thanks for the PR. This one looks good to me but I think we might have a related issue with READ_MEDIA_IMAGES as you pointed out earlier. Could you have a look into the call to TiramisuUtil.java for permissions? it looks like it will silently fail to load external images for newer devices (and likely other components are broken too like the videoplayer).

Thanks for the review! On it, I will look into the other TiramisuUtil callers (VideoPlayer, Sound, Player) as well and report back with findings.

zikunz:
> @zikunz thanks for the PR. This one looks good to me but I think we might have a related issue with READ_MEDIA_IMAGES as you pointed out earlier. Could you have a look into the call to TiramisuUtil.java for permissions? it looks like it will silently fail to load external images for newer devices (and likely other components are broken too like the videoplayer).

The root cause is that TiramisuUtil.requestFilePermissions unconditionally asks for READ_MEDIA_IMAGES, READ_MEDIA_VIDEO, or READ_MEDIA_AUDIO on Android 13+, but the manifest may not declare them. Image and VideoPlayer specifically cannot declare READ_MEDIA_IMAGES or READ_MEDIA_VIDEO due to the Google Play Photo and Video Permissions policy (per the earlier Anke thread). When the app has not declared the permission, the system silently denies the request.

A matching conditional pattern already exists in the codebase. FileOperation.java, PhoneCall.java, and SUtil.java all use form.doesAppDeclarePermission(perm) to ask only when the app has actually declared the permission, and TiramisuUtil does not follow that pattern.

Image (this PR) and VideoPlayer are both affected, and both can be fixed by switching TiramisuUtil to the conditional pattern. Adding READ_MEDIA_IMAGES or READ_MEDIA_VIDEO to @UsesPermissions is blocked by the Play policy.

Sound has no method-level @UsesPermissions on Source at all, and adding @UsesPermissions({READ_EXTERNAL_STORAGE, READ_MEDIA_AUDIO}) on Source would match Player.java and remain allowed under the Play policy, since READ_MEDIA_AUDIO is not covered there.

Player is already correct.

I am happy to split the scope however you prefer, and I see at least two possible options.

(a) I update TiramisuUtil in this PR, which would fix Image and VideoPlayer together, and I would also update the PR title and description to reflect the wider scope. Sound would go in a separate PR.

(b) I keep this PR focused on Image only, and open separate PRs for TiramisuUtil (together with VideoPlayer) and for Sound.

Please let me know which you prefer.

josmas:
Hey @zikunz I just talked to @ewpatton and he pointed out that this is likely not a fix.

The Picture property will always be used, even from the designer, so moving the permission to the property does not fix the problem.

I believe what the community report is saying is that if the app is only going to use images that are internal to the app, the permission is not needed (it is only needed for external locations).
Possible solutions are to let the user decide (what Anke proposes) or to somehow figure out at build time if the assets are external and only then add the permission. Let me know if that makes sense.

PR #3917, opened 2026-05-02
Lambda procedure doc

daki7711:
Documentation for the new anonymous procedure blocks


daki7711:
Working more on it but made a PR for at least feedback on structure of the files

SusanRatiLane (dismissed):
This looks good to me. I think the old named procedure blocks and the new anonymous/higher-order blocks should be under two distinct headers with the explanation of higher-order procedures under that header.

jisqyv (approved):
LGTM

PR #3921, opened 2026-05-04
Sanitize Gallery Imports

barreeeiroo:

Fixes importing Gallery projects with parenthesis or non-alphanumeric characters (and `_`).

*Description*

Sanitize Gallery project names before importing.

Fixes

*Testing Guidelines*

Fixes # .

Resolves # .

**Context for the changes**

- [ ] I branched from `ucr`

For all other changes:

**General items:**

- [ ] My code follows the:

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/server/project/youngandroid/YoungAndroidProjectService.java
+++ b/appinventor/appengine/src/com/google/appinventor/server/project/youngandroid/YoungAndroidProjectService.java
@@ -79,6 +79,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.StringReader;
 import java.net.ConnectException;
+import java.text.Normalizer;
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
 import java.net.URL;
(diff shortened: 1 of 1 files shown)


appinventor/appengine/src/com/google/appinventor/server/project/youngandroid/YoungAndroidProjectService.java
```diff
@@ -982,6 +983,27 @@ private String verifyProjectName(String userId, String projectName) {
     }
   }
 
+  /**
+   * Mirrors the client-side rules in {@link
+   * com.google.appinventor.client.youngandroid.TextValidators#isValidIdentifier(String)} so that
```
ewpatton, line 988: I wonder if it would make more sense to pull the TextValidators class into the shared package so the logic would be available on both sides?

barreeeiroo: In that case, I'd rather propose something else: automatically sanitize imported AIA files. This is a change I recently made in Kodular too:

> Upload flow shows a live preview of the final project name (after automatic sanitization/deduplication) and uses that resolved name when importing.

Basically, when importing an AIA, I do check if the project exists, and if so, just append a suffix to ensure uniqueness. It also previews the project name when sanitizing it or modifying it:

image
image
image

The main benefits:

* Users can import AIAs with spaces without manually renaming, as they get sanitized automatically. The same for other characters.
* Users can re-import existing projects, as a suffix is automatically appended.

The only tradeoff is a breaking change in terms of behaviour:

* If the unique name check was intentionally there, now they can import "duplicated projects" with different suffix.

Although this tradeoff to me sounds like something more beneficial as it can save time for a developer and avoid them renaming the AIA manually. I'm assuming the "let me reimport the same project" use case is more frequent than "I forgot I already had this project imported".

---

If it sounds good, I'll implement all that logic in a centralized place, for both direct AIA import and gallery import...

PR #3930, opened 2026-05-09
Advisor AI Agent

barreeeiroo:
# AI Agent for App Inventor (Advisor)

> **DISCLAIMER:** This PR is a "massive" work in progress. There is still plenty of work remaining before calling it "stable".

Implements an AI agent for App Inventor in a chat-style interface. The agent is **read-only**: it inspects the user's project and live runtime state, and answers in natural language, but does not modify the workspace. It can read live Companion runtime state to help with debugging, and consume tutorial content to guide the user better. Users can BYOK or fall back to server-provided defaults.

Full design and contributor docs: [CONTRIBUTING_AI.md](https://github.com/barreeeiroo/appinventor-sources/blob/ai-agent/CONTRIBUTING_AI.md)

## Design Overview

The agent is an editor-time feature inside the existing GWT client and Java App Engine server. It is not a runtime component of generated apps. Three principles shape the design:

- **Read-only by protocol.** The LLM is offered only read tools (`lookup_component`, `lookup_screen`, `read_*` Companion tools). Write tools are absent from the schema sent to the provider, and a server-side `ModeEnforcer` rejects any write attempt before it reaches the client. Read-only is enforced at the protocol layer, not by trusting model compliance.
- **Modular context.** The system prompt is assembled at request time from a chain of `ContextModule` instances (project, screen, catalog, mode, tutorial, Companion, etc.), each conditional and budget-aware.
- **Streaming via Memcache polling.** Tokens buffered server-side, consumed by a 250 ms client poll. No new transport on top of GWT-RPC.

## Capabilities

- **Guided project inspection.** `lookup_component` returns a component type's schema (properties, events, method signatures), including third-party extensions. `lookup_screen` returns the component tree and block YAIL of a named screen, exposing the student's actual instances, property values, and event wiring. The agent chains both to give specific, project-grounded answers.
- **Companion runtime reads.** When the App Inventor Companion is connected and the user enables the per-session share toggle, the agent can read component property values, global variable values, and runtime log slices through `CompanionBridge`. Hard-coded Scheme templates, regex-validated identifiers, per-turn (10) and per-minute (30) budgets. No path for the LLM to inject Scheme into the device.
- **Tutorial integration.** When a project's tutorial URL is set, `TutorialModule` fetches the content (cached 8 h) and instructs the agent to act as a tutor for *that lesson*: hints before solutions, locating the user's current step against live project state.
- **Contextual entry points.** Right-click "Explain Block" on any block, and "Ask AI about this error" on runtime errors, with the failing block's referenced symbols pre-attached via `BlockYailSymbolScanner` and resolved against live values from the device.
- **BYOK.** Users can supply their own provider, model, and key, or fall back to whatever the server admin configured.
- **Conversation persistence.** Chats retained per-project until the user deletes them or the project is removed. Memcache state has a 24 h TTL with rehydration from Datastore.

## Provider System

A single `LLMProvider` interface unifies stateless and stateful backends under one tool-use loop. Stateless providers (Anthropic, Bedrock, MiniMax, OpenRouter, Ollama, compatible) replay full history on each call. Stateful providers (OpenAI, Gemini, Vertex) maintain server-side state via `previous_response_id` or session refs.

Supported providers: `anthropic`, `anthropic-compatible`, `openai`, `openai-compatible`, `gemini`, `vertex`, `bedrock`, `ollama`, `minimax`, `openrouter`. Reasoning/thinking effort is provider-mapped via `ai.agent.reasoning.effort`.

## Configuration

Runtime provider/credentials in `appengine-web.xml`:

```xml
  <property name="ai.agent.provider" value="anthropic" />
  <property name="ai.agent.model" value="" />
  <property name="ai.agent.api.key" value="YOUR_KEY" />
```

Feature toggles (master switch, debug logging, editing modes, orchestration, etc.) live in `AppInventorFeatures.java` as compile-time constants: flipping any requires a rebuild, no runtime override.

Minimum to enable when using server defaults: set `AppInventorFeatures.aiAgentAvailable()` to true, then `ai.agent.provider`, `ai.agent.model`, `ai.agent.api.key`. Extra flags per provider: Bedrock needs AWS creds, Vertex needs GCP project plus service account, Ollama and compatible providers need `ai.agent.base.url`. Flip `aiAgentDebugEnabled()` to false for production.

Full provider matrix and feature flag table in [CONTRIBUTING_AI.md](https://github.com/barreeeiroo/appinventor-sources/blob/ai-agent/CONTRIBUTING_AI.md).


ewpatton:
Hmm...

Given how App Engine scaling works and how LLMs work, I think it would be better if we can deploy the server for interacting with the LLMs as a separate piece of infrastructure. Alternatively, we may want to have the communications go through the ChatBot proxy so we can give some free credits and then let people provide an API key only if they need more than a base level of credits.

barreeeiroo:
> it would be better if we can deploy the server for interacting with the LLMs as a separate piece of infrastructure

This sounds interesting, but I think there are a couple of things to keep in mind on how it currently works:

* As of now, some tools resolve server side. For example, `lookup_screen` will just fetch the state of a non-active screen (as the currently active is passed as part of the context) from Objectify, and return it back. If we want to decouple the LLM server, then we would have to make all tools resolve client side so we can remove the dependency of server-side tools (as it won't have access to Objectify).
* The LLM server currently uses Objectify to persist conversations and message history. Even though there's no hard requirement to use Datastore, we will still have to use some persistance layer for these.
* For authentication, we currently link conversation with User ID, and it's hard-coupled to the session cookies. We will need to find some way to let this extra server understand the session cookie, so we don't have to pass an arbitrary User ID which may be tampered.
* For BYOK, we could fully move this to the standalone service and remove the dependency from UserSettings. So this is actually a benefit.
* For extensions, we resolve the available components server side. Trivial to solve by providing client-side, but it's another refactor.

Then there are other couple of minor changes as well, but those are the main refactors to do in case we want to fully decouple it.

I did not consider things like letting the server call Datastore/Memcache APIs directly as kinda defeats the purpose of hosting it separately... If we go through this path, I'd rather have a fully separate service which doesn't have any dependency on `appengine/` besides the API contracts with the client.

> we may want to have the communications go through the ChatBot proxy

This is also possible, and it's not dependent on the potential refactor mentioned earlier. However, we need to make sure we can support tool calls in that proxy, as it's how the LLM can discover the available components, properties and blocks, along with other screens (and potentially edits).

---

PS: Not sure why those tests are failing in the CI build. They were passing locally; I'll try to sort it out this next week. In any case, it's fine to build and test locally as it is right now.

PR #3931, opened 2026-05-10
Fix MockMap TypeError on import

barreeeiroo:

Avoids crashing App Inventor when importing projects with Map component. This seems to be more like a timing issue though, I couldn't build a "dummy AIA" to replicate... I've experience this when judging Appathon projects.

*Description*

It seems like it sometimes tries to call the `addToMap` method before the Leaflet library initialized the map. When that happens, it throws a NPE when calling `addLayer`.

*Testing Guidelines*

Reach out for an AIA that causes this bug. After the fix, it doesn't happen again.

Fixes # .

Resolves # .

**Context for the changes**

- [ ] I branched from `ucr`

For all other changes:

**General items:**

- [ ] My code follows the:

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockMap.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/editor/simple/components/MockMap.java
@@ -155,7 +155,7 @@ public final class MockMap extends MockContainer {
 
   @Override
   protected void onChildAdded(MockComponent component) {
-    if (component instanceof MockMapFeature) {
+    if (component instanceof MockMapFeature && mapInstance != null) {
       ((MockMapFeature) component).addToMap(this);
     }
   }


josmas:
@barreeeiroo if it is a timing issue, and we skip adding the map, the NPE is gone but the maps does not load? Would it look like it is broken anyway?

barreeeiroo:
@josmas that's what I was thinking as well. However, I think the issue might be when the map is in a non-active Screen. I did test the broken project post-fix, and the Map Mock was rendering correctly. I did also test for an active Screen one (basically, importing an AIA where there was a Map in both the first and non-first Screen), and they also rendered correctly.

In any case, even though I couldn't build a simple AIA to reproduce this, my way of thinking for the fix is that it's better to have a non working Map Mock (despite I never observed it), than Ode client crashing.

josmas:
Yeah, agree, my only concern is silently masking a race condition that would be more difficult to track in the long term. But I do agree on removing the NPE. Will approve now.

josmas (approved):
LGTM, thanks!

PR #3935, opened 2026-05-12
Fix: a small listener bug in setDivider() - Android

### LINKED ISSUE #2597: ListView separator line between items.

prabhasg5:

Fixes a listener leak in `ListView.setDivider()`. The method was calling `removeLayoutChangeListener()` on the newly created `DividerItemDecoration` instead of the existing one being replaced. As a result, the old decoration's `OnLayoutChangeListener` — which gets registered on the `RecyclerView` was never unregistered before the old decoration was discarded. Every subsequent call to `setDivider()` (triggered by changing `DividerColor`, `DividerThickness`, or `ElementMarginsWidth`) would leave another orphaned listener registered on the `RecyclerView`.

Fixes #2597

- [x] Google Java style guide

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/ListView.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/ListView.java
@@ -1305,16 +1305,15 @@ public final class ListView extends AndroidViewComponent {
    * Sets new dividers or margins in RecyclerView
    */
   private void setDivider() {
-    DividerItemDecoration dividerDecoration = new DividerItemDecoration();
-    dividerDecoration.removeLayoutChangeListener();
     for (int i = 0; i < recyclerView.getItemDecorationCount(); i++) {
       RecyclerView.ItemDecoration decoration = recyclerView.getItemDecorationAt(i);
       if (decoration instanceof DividerItemDecoration) {
+        ((DividerItemDecoration) decoration).removeLayoutChangeListener();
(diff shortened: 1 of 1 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

ewpatton:

SusanRatiLane (changes requested):
This looks fine and tests out. However, it's a change to the Android companion. Changes to the Android Companion need to branch from ucr instead of master. That's an unusual and unintuitive part of our repository.

prabhasg5:
oh is it? ok then i will just close the PR and cherry pick the same commit and place it in branch pulled from ucr and make a PR again. does that make sense?

SusanRatiLane:
> oh is it? ok then i will just close the PR and cherry pick the same commit and place it in branch pulled from ucr and make a PR again. does that make sense?

That seems like exactly the right thing.

prabhasg5:
Closing in favour of #3943 which branches from ucr

ewpatton:
Note that in the future, so long as master and ucr are in sync, you can just change the base branch in the PR rather than opening a new PR.

PR #3937, opened 2026-05-12
Add runtime i18n prototype for app text

Akashgite1:

*Description*
This is a draft PR for early architectural feedback on a runtime-only MVP for app-level internationalisation.
The PR adds a small `I18nUtil` runtime helper that loads an `i18n.json` file from app assets, resolves strings marked with the `@i18n:` prefix using the device locale, and falls back safely when translations are missing.

For this prototype, the integration is intentionally limited to:

- `Button.Text`
- `Label.Text`
- `TextBox.Hint`

Normal strings remain unchanged, so existing projects should continue to work without modification.

Example:

```text
@i18n:screen1_button1_text

with i18n.json:

{
  "defaultLanguage": "en",
  "translations": {
    "screen1_button1_text": {
      "en": "Submit",
      "hi": "जमा करें"
    }
  }
}

```
The runtime resolves the value based on the device locale.

This PR does not yet include:

- Designer-generated translation keys
- Translation Editor UI
- Blockly translated-text blocks
- Build-time Android strings.xml generation
- Runtime UI refresh after device language changes
Known limitation: the locale is read at app startup. If the device language changes while the app is already running, existing UI does not refresh automatically yet.

Testing Guidelines

I tested this manually with a small sample app containing:

- Button1.Text = @ i18n:screen1_button1_text
- Label1.Text = @ i18n:screen1_label1_text
- TextBox1.Hint = @ i18n:screen1_textbox1_hint

and an uploaded i18n.json media asset.

Manual checks performed:

1. Button.Text resolves translated values
2. Label.Text resolves translated values
3. TextBox.Hint resolves translated values
4. Plain strings remain unchanged
5. TextBox.Text is intentionally not translated
6. Runtime updates from Blocks work for Button.Text, Label.Text, and TextBox.Hint
7. Changing language while the app is already running does not refresh UI automatically, which is expected for this MVP

This is currently a draft for feedback before expanding the implementation.

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/util/I18nUtil.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/util/I18nUtil.java
@@ -1,0 +1,221 @@
+// -*- mode: java; c-basic-offset: 2; -*-
+// Copyright 2025 MIT, All rights reserved
+// Released under the Apache License, Version 2.0
+
+package com.google.appinventor.components.runtime.util;
+
+import android.content.res.AssetManager;
+import android.util.Log;
+
+import com.google.appinventor.components.runtime.Form;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.io.ByteArrayOutputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Locale;
+import java.util.Map;
+
+/**
+ * Minimal runtime internationalization helper for App Inventor applications.
+ *
+ * MVP behavior:
+ * - Loads i18n.json from app assets if present.
+ * - Resolves strings prefixed with @i18n:.
+ * - Falls back safely when file/key/language is missing.
+ *
+ * This is intentionally runtime-only. Designer key generation, Blockly
+ * integration,
+ * and build-time strings.xml generation should be added in later phases.
+ */
+public final class I18nUtil {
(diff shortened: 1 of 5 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

appinventor/components/src/com/google/appinventor/components/runtime/util/I18nUtil.java
```diff
@@ -0,0 +1,221 @@
...
+
+            parse(json);
+            Log.i(LOG_TAG, "Loaded " + translations.size() + " translation keys from " + I18N_FILE);
+        } catch (FileNotFoundException e) {
+            // Backward compatibility: most existing apps will not have i18n.json.
+            Log.i(LOG_TAG, "No " + I18N_FILE + " found. Internationalization disabled.");
+        } catch (IOException e) {
+            Log.w(LOG_TAG, "Unable to read " + I18N_FILE + ". Internationalization disabled.", e);
+        } catch (JSONException e) {
+            Log.w(LOG_TAG, "Invalid " + I18N_FILE + ". Internationalization disabled.", e);
```
ewpatton, line 83: We may want to bubble up these exceptions in a way that makes sense to the user. Right now, if something goes wrong, i18n will silently fail and there will be no feedback on what to fix.

Akashgite1: This PR is just for feedback and my own understanding; we are not supposed to merge this one

PR #3943, opened 2026-05-15
corrected a small listener bug in setDivider()

### LINKED ISSUE #2597: ListView separator line between items.

prabhasg5:

Fixes a listener leak in `ListView.setDivider()`. The method was calling `removeLayoutChangeListener()` on the newly created `DividerItemDecoration` instead of the existing one being replaced. As a result, the old decoration's `OnLayoutChangeListener` — which gets registered on the `RecyclerView` when horizontal orientation and margins are both active — was never unregistered before the old decoration was discarded. Every subsequent call to `setDivider()` (triggered by changing `DividerColor`, `DividerThickness`, or `ElementMarginsWidth`) would leave another orphaned listener registered on the `RecyclerView`.

Fixes #2597

- [x] I branched from `ucr`
- [x] Google Java style guide

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/ListView.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/ListView.java
@@ -1305,16 +1305,15 @@ public final class ListView extends AndroidViewComponent {
    * Sets new dividers or margins in RecyclerView
    */
   private void setDivider() {
-    DividerItemDecoration dividerDecoration = new DividerItemDecoration();
-    dividerDecoration.removeLayoutChangeListener();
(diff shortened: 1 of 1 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

SusanRatiLane (approved):
Looks good

PR #3946, opened 2026-05-19
Fix iOS unit test memory leak from EventDispatcher delegate retention.

vanshapple:


AppInventorWorkerBee:

AppInventorWorkerBee:

appinventor/schemekit/src/SCMInterpreter.h
```diff
@@ -12,7 +12,7 @@
 
 @interface SCMInterpreter : NSObject
 
-+ (void)setDefault:(nonnull SCMInterpreter *)defaultInterpreter;
++ (void)setDefault:(nullable SCMInterpreter *)defaultInterpreter;
```
ewpatton, line 15: Why would the interpreter ever be null? I feel like nonnull is still appropriate here.

ewpatton: Never mind. I see what you're trying to do here w.r.t. clearing out the entire interpreter state. I'm not 100% sure that will actually work given a number of strong references that are maintained there, but we can consider it.

ewpatton:

appinventor/components-ios/tests/AppInventorTestCase.swift
```diff
@@ -168,6 +168,23 @@ class AppInventorTestCase: XCTestCase {
...
+
+  /// Regression guard: dispatch delegates must not accumulate when tests finish.
+  func testDispatchDelegateReleasedAfterTearDown() {
+    let baseline = EventDispatcher.registeredDispatchDelegateCount
+    expectToReceiveEvent(on: form, named: "BackPressed")
+    XCTAssertGreaterThan(EventDispatcher.registeredDispatchDelegateCount, baseline)
+    form.checkerMap.removeAll()
+    expectations.removeAll()
+    releaseTestResources(form: form, interpreter: interpreter)
+    XCTAssertEqual(baseline, EventDispatcher.registeredDispatchDelegateCount)
```
ewpatton, line 188: I don't understand this addition. It's unclear how the form would receive the BackPressed event in this test, and its addition causes all of the other subclasses to fail since this fails. Make sure you run the tests and they all pass before pushing an update to this PR.

ewpatton (changes requested):

appinventor/schemekit/src/SCMInterpreter.h
vanshapple: Okay

PR #3950, opened 2026-05-21
Add tolerance for math_compare block

### LINKED ISSUE #3948: Add a compare with tolerance block

Floating point numbers present a challenge for folks expecting precision on decimal numbers. The equality operators in the blocks generally won't work for things that cannot be expressed as a sum of powers of two. Other languages have an operator that compares floats with a given tolerance. It's possible to implement such an operator using procedures, but it might be better to just provide it as a built-in block.

Working with any floating point data.

It can be but requires defining an entire procedure rather than using a single math block like a beginning might need.

Understanding floating point numbers is important in a CS context.

yashsrv:
This PR solves the famous issue users often deal with, floating numbers comparison.

0.1 + 0.2 ≠ 0.3 in floating-point arithmetic because real numbers like 0.1 and 0.2 can't be represented exactly in binary, the stored values are approximations, so their sum accumulates a tiny error. 

To handle this, a tolerance block is used:

1. Add tolerance input using mutator

image

3. When tolerance is user-provided, it uses absolute comparison

image

4. When default tolerance is used, it uses relative comparison (see: https://floating-point-gui.de/errors/comparison/)

image

This approach always return correct result, the tolerance is adaptive here!

5. Tolerance do not alter the result when inputs are non-numeric

image

*Testing Guidelines*
Set up your blocks exactly as above and test the implementation.

Resolves #3948 

**Context for the changes**

- [x] I branched from `ucr`

For all other changes:

**General items:**

- [ ] I have updated the relevant documentation files under docs/html/reference/blocks/math.html
- [x] My code follows the:


AppInventorWorkerBee:

AppInventorWorkerBee:

daki7711:
I can try for myself and see, but I didn't want to wait for ant to compile.... LOL

does it have both of the blocks? because I think there should be a mutator where the user can toggle between the version that has tolerance and the version that hasn't.

The version that hasn't should be the default.

yashsrv:
> does it have both of the blocks?

Yes

> because I think there should be a mutator where the user can toggle between the version that has tolerance and the version that hasn't.

This sounds good, should I move with this approach?

daki7711:
I personally think the mutator is the correct approach, adding another block will bloat our already busy blocks.
But let's give it a day or two for people to have their opinions here and then proceed

daki7711:
@yashsrv I think at this point the general consensus is the tolerance should be hidden by a mutator. 
There should be a default tolerance value so that basic arithmetic such as 0.1 + 0.2 = 0.3 works (we need to test out all >, <, >=, <= ...) 

But then the user can click the mutator to tinker the tolerance value.

When you get the chance, could you update this fix?

yashsrv:
> When you get the chance, could you update this fix?

Sure, I'll add a commit soon :)

yashsrv:
Hi @daki7711, please let me know if this looks right?

PS: I'll create changes in docs if it's a Yes!

yashsrv:
@daki7711 Hi, there's a concern regarding operator EQ & NEQ. First, these are the internal operations for different operators:

image

The issue is that EQ and NEQ compute abs(arg0 - arg1) internally, so they implicitly require numeric types, but the current signature says "any any", so non-numeric inputs can slip through and fail at runtime.

We can mitigate this by changing the signature to "number number", but this can break existing projects that passes non-numeric values to EQ/NEQ.

OR, we can remove the default tolerance, tolerance should be enabled only when it's added through mutator. This avoids disrupting existing projects.

daki7711:
@yashsrv sorry in advance, but after looking at the mutator I think it's a little awkward that users added the tolerance just like adding another variable. Considering tolerance should be rather added or not, it should work more like blocks "External input" vs "Internal input" (you can see an example by dragging in the "+" block and right clicking the block. So in the same sense, when you right click the '=' block there should be "adjust tolerance" where your tolerance pops up. Let me know if that makes sense.

Once we get to that point, maybe we can put this on ai2-test for people to try it out. I saw the discussion on the community site

daki7711:
> @daki7711 Hi, there's a concern regarding operator EQ & NEQ. First, these are the internal operations for different operators:
> 
> image
> The issue is that EQ and NEQ compute abs(arg0 - arg1) internally, so they implicitly require numeric types, but the current signature says "any any", so non-numeric inputs can slip through and fail at runtime.
> 
> We can mitigate this by changing the signature to "number number", but this can break existing projects that passes non-numeric values to EQ/NEQ.
> 
> OR, we can remove the default tolerance, tolerance should be enabled only when it's added through mutator. This avoids disrupting existing projects.

For this concern, can we check if it's "number, number" and apply the default tolerance if so?
Cause I do agree with one of the community comments that 0.1 + 0.2 =0.3 should be true.

yashsrv:
@daki7711 Hi, lets clear the UX idea properly, please correct me if I'm wrong.

1. Remove mutator completely from math_compare block
2. Add a contextmenu item -> "Show Tolerance" / "Hide Tolerance", which should appear when user right clicks on the math_compare block
3. After clicking on it, the block UI should update like this where user can adjust the tolerance (prefilled with default tolerance)

math_compare_tolerance

> can we check if it's "number, number" and apply the default tolerance if so?

Yes, this is doable, will implement it too

daki7711:
@yashsrv yes, exactly

Thank you very much, these seem like small changes but sometimes require a lot of back and forth.

yashsrv:
> @yashsrv yes, exactly
> 
> Thank you very much, these seem like small changes but sometimes require a lot of back and forth.

True, but after community discussion and your ideas, now I have a clear path. I'll commit soon :)

mark-friedman (changes requested):
I think I would object to that PR  as it currently stands.  For one, it changes the semantics of existing programs, since it doesn't have an upgrader.  I also think that its approach of having a single default tolerance is too simplistic.  For example, see https://floating-point-gui.de/errors/comparison/

yashsrv:
> I also think that its approach of having a single default tolerance is too simplistic. For example, see https://floating-point-gui.de/errors/comparison/

I get the point here, I actually tested this case, even they both differ by 100% they are considered equal because the tolerance is too large for these small numbers.

blocks(1)




> For one, it changes the semantics of existing programs, since it doesn't have an upgrader.

Even if I address this concern https://github.com/mit-cml/appinventor-sources/pull/3950#issuecomment-4529869225, will I need to add an upgrader? Could you please tell more about upgrader thing as I am not very clear about it.

yashsrv:
@mark-friedman Hi, I've patched a fix:

Screenshot 2026-05-27 at 11 30 59 AM

❌ _This is a wrong result coming from my old code, as the single default tolerance (0.0000001) is too large for these numbers_

Screenshot 2026-05-27 at 11 33 20 AM

✅ _This is the right result after the patch_

daki7711:
@mark-friedman can you review these changes?

I think it's at least worth putting it into ai2-test so we can get community feedback

yashsrv:
@daki7711 I've pushed the new changes, as we discussed, this now has a customContextMenu to toggle tolerance input + now it supports non-numeric inputs for EQ/NEQ operator (no runtime error)

_PR description has been updated_

mark-friedman:
> So in the same sense, when you right click the '=' block there should be "adjust tolerance" where your tolerance pops up.

I disagree.  Mutators (and the mutator UI) are the accepted way in Blockly and App Inventor to add/remove inputs.  Context menu items can be used to change the way that a block is rendered (e.g. internal vs. external inputs) but it shouldn't be used to add/remove inputs.

appinventor/blocklyeditor/src/generators/yail/math.js
```diff
@@ -31,19 +31,102 @@ AI.Yail.forBlock['math_compare'] = function(block, generator) {
...
+        + AI.Yail.YAIL_OPEN_COMBINATION
+        + (mode === 'EQ' || mode === 'NEQ' ? 'any any' : 'number number')
+        + AI.Yail.YAIL_CLOSE_COMBINATION + AI.Yail.YAIL_SPACER;
+    code = code + AI.Yail.YAIL_DOUBLE_QUOTE + operator2
+        + AI.Yail.YAIL_DOUBLE_QUOTE + AI.Yail.YAIL_CLOSE_COMBINATION;
+    return code
+  };
+  var nearlyEqual = function() {
+    var code;
+    var operator1 = '<';
```
mark-friedman: Since this is a constant (i.e. '<') just name it something like `lessThan` or even just not have it be a variable at all and use '<' in the code below.

appinventor/blocklyeditor/src/generators/yail/math.js
```diff
@@ -31,19 +31,102 @@ AI.Yail.forBlock['math_compare'] = function(block, generator) {
...
+    var abs0 = '(abs ' + arg0 + ')';
```
mark-friedman: To be consistent with the rest of the code, you should use the `AI.Yail.YAIL_` constants rather than '(' and ')'.  In retrospect I think that those constants make the code harder to read and parse, but at this point we should accept it and be consistent.

appinventor/blocklyeditor/src/generators/yail/math.js
```diff
@@ -31,19 +31,102 @@ AI.Yail.forBlock['math_compare'] = function(block, generator) {
...
+    var abs1 = '(abs ' + arg1 + ')';
+    var diff = '(abs (- ' + abs0 + ' ' + abs1 + '))';
+
+    // Min Threshold: tol * Number.MIN_VALUE
+    var minThreshold = '(* ' + tol + ' ' + Number.MIN_VALUE + ')';
```
mark-friedman: I'm not sure that it is correct to use Java's `Number.MIN_VALUE` here.  This code is being executed by Scheme, which may have a different value for the minimum value.  This suggests that perhaps the threshold value should actually be computed in the Scheme code in the two runtime.scm files.  My suggestion would be to define a `yail-compare` Scheme procedure in those runtime.scm files (look for the "Math implementation" section of those files.

appinventor/blocklyeditor/src/generators/yail/math.js
```diff
@@ -31,19 +31,102 @@ AI.Yail.forBlock['math_compare'] = function(block, generator) {
...
-  code = code + AI.Yail.YAIL_SPACER + AI.Yail.YAIL_QUOTE
-      + AI.Yail.YAIL_OPEN_COMBINATION + (mode == "EQ" || mode == "NEQ" ? "any any" : "number number" )
-      + AI.Yail.YAIL_CLOSE_COMBINATION + AI.Yail.YAIL_SPACER;
-  code = code + AI.Yail.YAIL_DOUBLE_QUOTE + operator2
-      + AI.Yail.YAIL_DOUBLE_QUOTE + AI.Yail.YAIL_CLOSE_COMBINATION;
+  var arg0 = generator.valueToCode(block, 'A', order) || 0;
+  var arg1 = generator.valueToCode(block, 'B', order) || 0;
+  var tol = block.toleranceVisible_ === true
+      ? generator.valueToCode(block, 'TOL', order) || 0
+      : 0.0000001; // Default tolerance.
```
mark-friedman: Isn't the default tolerance being calculated below?

appinventor/blocklyeditor/src/blocks/math.js
```diff
@@ -133,6 +133,67 @@ Blockly.Blocks['math_compare'] = {
       var mode = thisBlock.getFieldValue('OP');
       return Blockly.Blocks.math_compare.TOOLTIPS()[mode];
     });
+    this.toleranceVisible_ = false;
```
mark-friedman: I don't think that we should be using a context menu item for this.

mark-friedman (changes requested):
Note especially my comments about the use of a context menu and about doing most of the tolerance handling in the Scheme code

mark-friedman:
> I think it's at least worth putting it into ai2-test so we can get community feedback

Sure, once the issues I've raised in m review comments are resolved.

appinventor/blocklyeditor/src/generators/yail/math.js
yashsrv: Done, removed this function and switched to yail-compare scheme procedure

appinventor/blocklyeditor/src/generators/yail/math.js
yashsrv: Fixed

appinventor/blocklyeditor/src/generators/yail/math.js
yashsrv: Done

appinventor/blocklyeditor/src/generators/yail/math.js
yashsrv: This is required to calculate diff in yail-compare

yashsrv:
@mark-friedman I have moved the complete tolerance handling to yail-compare procedure. Also, now the math_compare block has a mutator

image

appinventor/blocklyeditor/src/generators/yail/math.js
```diff
@@ -28,32 +28,38 @@ AI.Yail.forBlock['math_compare'] = function(block, generator) {
...
-  var operator2 = prim[1];
-  var order = prim[2];
+  var operator = prim[0];
+  var order = prim[1];
   var argument0 = generator.valueToCode(block, 'A', order) || 0;
   var argument1 = generator.valueToCode(block, 'B', order) || 0;
-  var code = AI.Yail.YAIL_CALL_YAIL_PRIMITIVE + operator1
+  var tol = block.itemCount_ === 1
+      ? generator.valueToCode(block, 'TOL0', order) || 0
+      : 0.0000001; // Default tolerance to calculate diff in yail-compare.
```
mark-friedman, line 37: I believe that this 0.0000001 value will end up being unused if itemCount is not 1 and the default, relative, computation will be used.  Maybe put a comment here specifying that.

appinventor/blocklyeditor/src/versioning.js
```diff
@@ -2411,7 +2411,10 @@ Blockly.Versioning.AllUpgradeMaps =
     37: "noUpgrade",
 
     // AI2: Added Anonymous procedure Blocks
-    38: "noUpgrade"
+    38: "noUpgrade",
+
+    // AI2: Add tolerance mutator for math_compare block.
+    39: "noUpgrade"
```
mark-friedman, line 2417: You will need an upgrader to avoid preexisting code having new semantics.  I believe that the upgrader will need to change preexisting code to provide an explicit tolerance of 0.

mark-friedman (changes requested):
It would also be nice to add some tests to [YailEvalTest.java](https://github.com/mit-cml/appinventor-sources/blob/master/appinventor/buildserver/tests/com/google/appinventor/buildserver/YailEvalTest.java)

ewpatton:
One question is "what do we mean by tolerance?" Depending on the domain that can mean different things. For example, in woodwork, a 1 mm tolerance might be a big deal, but 1 mm doesn't mean squat for astronomers measuring the size of planets. Regardless of units, if we have two very small floating point numbers or two very large floating point numbers (abs(mantisaa) > 16, say), absolute thresholds may be less than optimal.

Therefore, we might want to consider absolute tolerances versus relative tolerances. What might the block look like if we consider that?

ewpatton:
@yashsrv Just to be clear, there are other ways to provide mutations versus what was implemented here. The use of the container block with one or more "tolerance" entries doesn't make much sense to me. Instead, we might want to have a checkbox on the "tolerance" container (the name should be adjusted). If we have to do absolute vs relative tolerances that may complicate things but that's a separate issue.

mark-friedman:
> absolute tolerances versus relative tolerances

Wasn't that addressed in some of my comments and subsequent commits?  Or are you saying that you actually do want absolute tolerances?

ewpatton:
I guess my point is that either can be relevant depending on the context. If we only do absolute, or only do relative, either is a failure of imagination.

mark-friedman:
> I guess my point is that either can be relevant depending on the context. If we only do absolute, or only do relative, either is a failure of imagination.

Fair enough.

yashsrv:
> Instead, we might want to have a checkbox on the "tolerance" container (the name should be adjusted)

Hi @ewpatton, I agree with the mutator container having a checkbox instead of drag & drop UI, does this UI make sense? could you please suggest a name also?

image

appinventor/blocklyeditor/src/generators/yail/math.js
yashsrv: > default, relative, computation will be used

Yes but that uses `nearly-equal?` and it requires `tol` value, could you please let me know how it will end up being unused?

```
(define (nearly-equal?)
    (let* ((abs1 (abs x1))
           (abs2 (abs x2))
           (diff (abs (- abs1 abs2)))
           (sum  (+ abs1 abs2)))
      (if (or (= x1 0)
              (= x2 0)
              (< sum java.lang.Float:MIN_NORMAL))
          (< diff (* tol java.lang.Float:MIN_NORMAL))
          (< (/ diff (min sum java.lang.Float:MAX_VALUE))
             tol))))
```

yashsrv:
> Therefore, we might want to consider absolute tolerances versus relative tolerances. What might the block look like if we consider that?

absolute vs relative blocks

Absolute

When the user doesn't provide a tolerance, we fall back to `nearly-equal?` (the 2nd block above) which already uses a relative comparison that scales with magnitude. The absolute comparison only applies when the user explicitly provides a tolerance value.

appinventor/blocklyeditor/src/generators/yail/math.js
mark-friedman: Ah, I see now.  Thanks for the clarification.

PR #3951, opened 2026-05-21
Nested More Category for Lambda Procedures

daki7711:

This is an attempt to clean our blocks category.

Screenshot 2026-05-21 at 4 28 28 PM

As you can see above, the regular "Procedure" Blocks will only contain the original Value/ Statement named procedure blcoks. However, if you click on the More category, which is nested under the Procedure category, you will able to see see the anonymous procedure blocks.

We plan to do something similar for all categories.

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
@@ -1094,7 +1094,7 @@ public interface OdeMessages extends Messages, ComponentTranslations {
   @Description("Label on built-in-Procedures-blocks branch of block selector tree")
   String builtinProceduresLabel();
 
-  @DefaultMessage("More")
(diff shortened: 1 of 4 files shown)


daki7711:
This will make https://github.com/mit-cml/appinventor-sources/pull/3911 obsolete @jisqyv

yashsrv:
This is a nice feature to have, I also have this idea to combine similar blocks, like having only one trigonometry block for all of these: 

image

To help users understand what does the block do, we can append a field "trigonometry". By this way, we can reduce the number of blocks in flyout. Let me know your views on this :)

daki7711:
I agree, I'm working with our team inside about grouping our blocks and nest them under each category. so the blocks in flyout is less busy and also easier for user to find the block they want.

jisqyv (approved):
LGTM

PR #3954, opened 2026-05-26
fix iOS Layout autosize, and ListView UI bugs

daki7711:

App Inventor has been having this long annoying bug when auto-sizing things in iOS. Especially when it comes to our Arrangements and ListView. This PR is an attempt to aliign autosize behavior close to the android behaviors as much as possible.

Problem
The iOS renderer was not consistently treating App Inventor Automatic as Android-style wrap_content.
On Android, arrangements and ListView naturally measure their children/items and then size themselves. On iOS, several components were either reporting no useful intrinsic size, keeping stale constraints, or sizing against the wrong parent. That caused cases like:
- ListView inside HorizontalArrangement measuring too small/zero, so items did not appear.
- ListView automatic height expanding too much, then later being too tight because it assumed 44pt × rows.
- Bigger text, like FontSize = 22, needing taller rows than the old hard cutoff allowed.
- Percent sizes and automatic sizes behaving differently depending on arrangement type.

Critical Changes
The biggest fix is in ListView.swift (line 18). I added a root view with a real intrinsic size, so arrangements can measure ListView properly. Then I made vertical automatic height come from item count and row height instead of “fill available space” or a fixed 44pt guess.

For the cutoff bug specifically, ListView.swift (line 1027) now computes row height from the configured fonts plus padding, and ListView.swift (line 972) makes the table rows use the same height. With three rows at FontSize = 22, the test now lays out at 171pt, not the old 132pt.

For arrangements, LinearView.swift (line 527), AbsoluteArrangement.swift (line 53), and RelativeLayout.swift (line 148) now report intrinsic sizes based on visible children. I also cleaned up stale constraint handling so hidden/removed/replayed components do not leave old sizing constraints behind.

For percentages, AbsoluteArrangement.swift (line 150) and TableArrangement.swift (line 102) now size percentage dimensions against the form scale frame where appropriate, which is closer to how App Inventor users expect percentage sizing to behave.

Added regression coverage in ListViewTests.swift (line 40) and LinearViewTests.swift (line 81).

**General items:**

- [x] My code follows the:

### DIFF
--- a/appinventor/components-ios/src/AbsoluteArrangement.swift
+++ b/appinventor/components-ios/src/AbsoluteArrangement.swift
@@ -25,12 +25,16 @@ open class AbsoluteArrangement: ViewComponent, ComponentContainer, AbstractMetho
 
   // Layout
   private var viewLayout: RelativeLayout
+  private var _widthConstraints = [ObjectIdentifier: NSLayoutConstraint]()
+  private var _heightConstraints = [ObjectIdentifier: NSLayoutConstraint]()
   
   // MARK: - Constants
   private let NOT_VALID: Int = -1
   
   // MARK: - Custom View for Absolute Layout
   private class AbsoluteView: UIView {
+    weak var contentView: UIView?
+
     override init(frame: CGRect) {
       super.init(frame: frame)
       translatesAutoresizingMaskIntoConstraints = false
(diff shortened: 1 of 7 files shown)


daki7711:
This is a nice [test aia](https://drive.google.com/file/d/1vlwHvApEOmtenHyxW1cAhOUiFoafbJ4l/view?usp=sharing), there are four screens where our current iOS would differ how it looks in android. With this PR, the screen behavior reflects android much closer in iOS devices.

Comparing the UI from that aia can be seen [here](https://docs.google.com/presentation/d/1wW7JAHmhpqE7EeDic_DYUYMJ3H7IDXkv2JQUoG3vWII/edit?slide=id.g3a8cce00295_0_0#slide=id.g3a8cce00295_0_0)

PR #3955, opened 2026-05-28
mockup UI fixes (Designer UI)

daki7711:

- Previously, the Mockup UI didn't show the hint in textbox. This PR fixes that bug so that the hint will show up inside textbox 

- Also fix Slider not enlarging in the Designer

- LinearProgress Color was messed up, this PR cleans it


daki7711:
Screenshot 2026-06-03 at 1 22 31 PM

Here is before this PR change 
The Slider's Width is set to fill the parent but in the Designer the Slider still remains short.
For the textbox, you can't see the hint text.

Screenshot 2026-06-03 at 1 22 40 PM

This is after the changes

daki7711:
Screenshot 2026-06-03 at 7 01 18 PM


Before <Note that when you first drag in LinearProgress it is fine, but when you explicitly check default color in the properties, it changes to white/grey as you above

Screenshot 2026-06-03 at 7 04 00 PM

After

SusanRatiLane (approved):
LGTM

PR #3956, opened 2026-05-28
[iOS] fix TextToSpeech country property

daki7711:

The TextToSpeech Country property was not compatible with the Country codes in our designer. This PR fixes it

Here is a good text [example](https://drive.google.com/file/d/1BD51lKuyRN24mXaTOm_19sK7oqQC9jBR/view?usp=sharing)


SusanRatiLane (approved):
LGTM

PR #3958, opened 2026-05-28
[iOS] UI Discrepancy Fix

### LINKED ISSUE #3957: [Bug] Label BackgroundColor None displays as white instead of transparent on iOS Companion
Description
There is an inconsistency in how the Label component handles transparent backgrounds on the iOS Companion app.

- Setting the BackgroundColor to Default correctly results in a transparent background.

- Setting the BackgroundColor to None incorrectly results in an opaque white background.

Expected behavior is that None should render as transparent/clear, or at the very least, mirror the behavior of Default.

daki7711:
fixes issue #3957 but also other default color issues on the iOS side

builds on top of PR https://github.com/mit-cml/appinventor-sources/pull/3954


SusanRatiLane:
@daki7711 Should this still be in draft?

SusanRatiLane (approved):
LGTM

PR #3969, opened 2026-06-03
Fix account deletion logic

### LINKED ISSUE #3584: Deleting account w/o emptying trash

yashsrv:

1. Add getMyLiveProjectsCount method that returns correct projects count including nested projects in folders. Previous method considered only those folders' projects which are expanded, returning wrong result.
2. Add getLiveProjects method that returns list of project which are not in trash + available in any folder. This helps in counting live projects before account deletion.
3. Fix deleteAccount logic in ObjectStorageIo by adding some methods to check that all of the project IDs are in the trash folder. Previously, it was using an older flag which is obsolete now.

*Description*
_Fixed now_

https://github.com/user-attachments/assets/1def777a-f1b4-4754-91d0-9801f4aff310

*Testing Guidelines*

1. Create project/project folder with a project and move it to Trash.
2. Try deleting account, your account should get deleted without emptying the Trash.

Fixes #3584 

**Context for the changes**

- [ ] I branched from `ucr`

For all other changes:

**General items:**

- [x] My code follows the:

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/server/storage/ObjectifyStorageIo.java
+++ b/appinventor/appengine/src/com/google/appinventor/server/storage/ObjectifyStorageIo.java
@@ -683,24 +683,6 @@ public class ObjectifyStorageIo implements StorageIo {
   }
 
   @Override
-  public void setMoveToTrashFlag(final String userId, final long projectId, final boolean flag) {
-    try {
-      runJobWithRetries(new JobRetryHelper() {
-        @Override
-        public void run(Objectify datastore) {
-          ProjectData projectData = datastore.find(projectKey(projectId));
-          if (projectData != null) {
-            projectData.projectMovedToTrashFlag = flag;
-            datastore.put(projectData);
-          }
-        }
-      }, true);
-    } catch (ObjectifyException e) {
-      throw CrashReport.createAndLogError(LOG, null, collectUserErrorInfo(userId), e);
-    }
-  }
-
-  @Override
   public List<Long> getProjects(final String userId) {
     final List<Long> projects = new ArrayList<Long>();
     try {
@@ -863,7 +845,7 @@ public class ObjectifyStorageIo implements StorageIo {
     } else {
       return new UserProject(projectId, projectData.t.name,
           projectData.t.type, projectData.t.dateCreated,
-          projectData.t.dateModified, projectData.t.dateBuilt, projectData.t.projectMovedToTrashFlag);
+          projectData.t.dateModified, projectData.t.dateBuilt);
     }
   }
 
@@ -895,7 +877,7 @@ public class ObjectifyStorageIo implements StorageIo {
       for (ProjectData projectData : projectDatas.t.values()) {
         uProjects.add(new UserProject(projectData.id, projectData.name,
             projectData.type, projectData.dateCreated,
-            projectData.dateModified, projectData.dateBuilt, projectData.projectMovedToTrashFlag));
+            projectData.dateModified, projectData.dateBuilt));
       }
       return uProjects;
     }
--- a/appinventor/appengine/src/com/google/appinventor/server/storage/ObjectifyStorageIo.java
+++ b/appinventor/appengine/src/com/google/appinventor/server/storage/ObjectifyStorageIo.java
@@ -109,6 +109,8 @@ import java.util.stream.Collectors;
 
 import javax.annotation.Nullable;
 
+import org.json.*;
+
 /**
  * Implements the StorageIo interface using Objectify as the underlying data
  * store.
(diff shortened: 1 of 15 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

SusanRatiLane:

appinventor/appengine/src/com/google/appinventor/server/storage/ObjectifyStorageIo.java
```diff
@@ -2812,18 +2814,19 @@ public void run(Objectify datastore) {
...
+    List<Long> trashProjectIds = getTrashProjectIds(userId);
     // We iterate over the projects in two loops The first loop is
     // just to determine that all remaining projects are in the trash.
     // The second loop actually removes such projects.  We do it this
     // way so that no projects are removed if any projects
     // exist. Otherwise some trashed projects may get removed before
     // we discover a live project.
     for (long projectId : projectIds) {
-      ProjectData projectData = getProject(projectId);
-      if (!projectData.projectMovedToTrashFlag) {
```
SusanRatiLane, line 2823: So, is the issue that the projectMovedToTrashFlag is not reliable? Can it be fixed? Where else is it used? 

It should either be fixed or removed.

appinventor/appengine/src/com/google/appinventor/server/storage/ObjectifyStorageIo.java
```diff
@@ -2812,18 +2814,19 @@ public void run(Objectify datastore) {
...
-        return false;           // Have a live project
+      if (!(trashProjectIds).contains(projectId)) {
```
SusanRatiLane, line 2807: All we really need to do here is verify that there are no live projects, correct? We shouldn't need to create a list of trashed projects and then loop through each to see if it's in the list. We're just doing the same check the same check as in DeleteAccountAction. We might not need to check here at all, though it's important to be really sure there are no active projects associated with an account that is being deleted.

SusanRatiLane (changes requested):

appinventor/appengine/src/com/google/appinventor/server/storage/ObjectifyStorageIo.java
yashsrv: When we delete a project, it calls `moveItemsToFolder`, which moves the projects to `*trash*` folder, it doesn't update the `projectMovedToTrash` flag at all. 

Also, when we restore a project, it only changes its folder. So yes it's not reliable anymore and I believe it should be removed with all the references.

appinventor/appengine/src/com/google/appinventor/server/storage/ObjectifyStorageIo.java
yashsrv: > All we really need to do here is verify that there are no live projects, correct?

Yes

This is added as an extra check just in case we get the globalFolder as null in `getMyLiveProjectsCount`
I believe we should keep it to prevent accidental account deletion with live projects.

Let me know if I should remove this check completely and rely only on `getMyLiveProjectsCount`

yashsrv:
@SusanRatiLane I've removed the `projectMovedToTrash` flag from the codebase with all references. 

I also noticed one bug while writing this another commit, users can't move a project to Trash from designer view because the method it uses is obsolete. I've replaced it with the correct method.

_Fixed now_

https://github.com/user-attachments/assets/0ecc8a04-fe82-4d6c-a836-544ac6c1b469

appinventor/appengine/src/com/google/appinventor/server/storage/ObjectifyStorageIo.java
SusanRatiLane: Under what circumstances might we get a null folder?

If you are creating a list of trashed projects, can you just check to see if it is empty or not? I'm unclear why we need the loop.

SusanRatiLane:
> @SusanRatiLane I've removed the `projectMovedToTrash` flag from the codebase with all references.
> 
> I also noticed one bug while writing this another commit, users can't move a project to Trash from designer view because the method it uses is obsolete. I've replaced it with the correct method.
> 
> _Fixed now_
> REC-20260606134801.mp4

This fix does not work in my testing. Are you sure it was pushed with the rest of your fixes?

So far, my testing of the rest of your work looks great.

SusanRatiLane:
> This fix does not work in my testing. Are you sure it was pushed with the rest of your fixes?
> 
> So far, my testing of the rest of your work looks great.

Sorry. It appears that this is on my end. Your work checks out. Thanks.

SusanRatiLane (approved):
LGTM

PR #3970, opened 2026-06-03
TinyDB iOS: enable PRAGMA fullfsync and surface silent failures

### LINKED ISSUE #3967: iOS issues with TinyDB
**Describe the bug**

On my iPhone, when I save image and text in a gallery via TinyDB, everything works fine for a while but when I check the next day, all the gallery images and text are gone.

[ImageTextGallery.aia.zip](https://github.com/user-attachments/files/28482000/ImageTextGallery.aia.zip)

**Affects**

- [ ] Designer
- [ ] Blocks editor
- [ ] Projects Explorer
- [ ] Android Companion
- [ ] iOS Companion
- [ ] iOS Compiled IPA
- [ ] Buildserver
- [ ] Debugging

**Expected behavior**

The image and text should stay put until the user intentionally deletes them.  TinyDB should be persistent.

**Steps to reproduce**

Try the file given, take some photos and write some text and save them to the gallery.  Check that they are there the next day.  Interestingly @daki7711 said this worked for him the next day so it may be dependent on iOS.  Mine is iPhone 16 with version 26.5.

Urgency: We are hoping to use the ImageTextGallery app with Kentucky students on their journey to MIT Summit.  They will take photos and record memories.  If not resolved by last week of June, it will mess up with the plans.

apexpdl:

Addresses #3967 — iOS `TinyDB` data disappears overnight on iOS 26.5.

*Description*
iOS's `fsync()` doesn't actually flush to disk: the OS returns success while the bytes are still in a write buffer. The only call that forces a real flush on iOS is `F_FULLFSYNC`, exposed in SQLite as `PRAGMA fullfsync`, and it's off by default for performance.
TinyDB opens its SQLite `Connection` with all defaults, so every write sits in an unflushed buffer. Foreground reads find it in the buffer and look fine. But once the app is suspended and an OS-level event happens overnight (low-battery shutdown, force restart, memory reclaim, OS update), the buffer can be discarded before reaching disk — and the data is silently gone the next day, matching the report.

Fix is one line right after opening the connection:

try db.run("PRAGMA fullfsync = 1")

This is Apple's documented recommendation for durable SQLite on iOS ([QA1809](https://developer.apple.com/library/archive/qa/qa1809/_index.html), [PRAGMA fullfsync](https://sqlite.org/pragma.html#pragma_fullfsync)).

Also cleans up the silent-failure paths that hid the bug — try? → try/catch with error.localizedDescription, guard let db = _database in every method, and removed the ! force-unwrap on pathForPrivateAsset. No behaviour change on the happy path.

*Testing Guidelines*

1. Build the iOS Companion on an iPhone running iOS 26.x.
2. Load ImageTextGallery.aia from the issue, save several entries.
3. Background the app, lock the device, leave overnight.
4. Before this PR: entries are gone the next day.
5. After this PR: entries persist.
6. Regression check: store/get/clear/namespace all behave identically; no new logs during normal use.

Fixes #3967

**Context for the changes**

- [x] I branched from `ucr`

For all other changes:

**General items:**

- [ ] My code follows the:

### DIFF
--- a/appinventor/components-ios/src/TinyDB.swift
+++ b/appinventor/components-ios/src/TinyDB.swift
@@ -17,16 +17,27 @@ open class TinyDB: NonvisibleComponent {
 
   public override init(_ parent: ComponentContainer) {
     let assetmgr = parent.form?.application?.assetManager
-    let sqlitedb = (assetmgr?.pathForPrivateAsset("TinyDb1.sqlite"))!
+    let sqlitedb = assetmgr?.pathForPrivateAsset("TinyDb1.sqlite") ?? ""
     do {
-      _database = try? Connection(sqlitedb)
+      _database = try Connection(sqlitedb)
+    } catch {
+      NSLog("TinyDB: Unable to open database at \(sqlitedb): \(error.localizedDescription)")
     }
     super.init(parent)
-    do {
-      _ = try? _database.run(_table.create(ifNotExists: true) { t in
-        t.column(_key, primaryKey: true)
-        t.column(_value)
-      })
+    if let db = _database {
+      do {
+        try db.run("PRAGMA fullfsync = 1")
+      } catch {
+        NSLog("TinyDB: Unable to enable fullfsync: \(error.localizedDescription)")
+      }
+      do {
+        try db.run(_table.create(ifNotExists: true) { t in
+          t.column(_key, primaryKey: true)
+          t.column(_value)
+        })
+      } catch {
+        NSLog("TinyDB: Unable to create table \(_namespace): \(error.localizedDescription)")
+      }
     }
   }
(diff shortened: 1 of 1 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

ewpatton (commented):
I'm going to put this up on TestFlight, but as far as the NSLog statements we may want to consider turning those into proper errors that get dispatched via the Form's ErrorOccurred event. Otherwise, the developer doesn't know something has gone wrong and their app may not be saving data.

ewpatton:

ewpatton (approved):
LGTM

PR #3973, opened 2026-06-03
[iOS] fix bypassing extension checking when switching projects

### LINKED ISSUE #3959: Extension validation bypassed on iOS Companion when switching projects without disconnecting
Description
There is a consistency bug in how the iOS Companion app handles unsupported extensions.

If you connect directly to a project (.aia) containing an unsupported extension, the Companion correctly throws an error and refuses to connect. However, if you first connect to a valid project that does not have the extension, and then switch to the project with the unsupported extension via the browser without explicitly disconnecting the Companion first, the validation is bypassed. The app connects and attempts to run, leading to undefined behavior or crashes later on.

Steps to Reproduce
Open a project Project A (which contains no extensions and works perfectly) in MIT App Inventor.

Connect the iOS Companion app to Project A. Verify it connects successfully.

Without disconnecting or closing the Companion app, switch projects in the browser to Project B (which contains an extension not supported by iOS).

Observe that the iOS Companion automatically refreshes and loads Project B without throwing the expected "Extension not supported" error.

Expected Behavior
When the Companion receives the asset/block payload for Project B, it should detect the unsupported extension, terminate the connection, and display the appropriate error message—matching the behavior of a fresh connection.

Actual Behavior
The Companion reuses the existing active connection lifecycle, bypassing the initial extension check, and attempts to render a project it cannot actually support.

Environment / Context
Device: iPhone / iPad

App: iOS Companion App

Scope: Extension Lifecycle / Project Switching State

daki7711:
fixes #3959

### DIFF
--- a/appinventor/blocklyeditor/src/replmgr.js
+++ b/appinventor/blocklyeditor/src/replmgr.js
@@ -1496,14 +1496,7 @@ Blockly.ReplMgr.getFromRendezvous = function() {
                     // Newer iOS versions will report the extensions that they support
                     top.ALLOWED_IOS_EXTENSIONS = JSON.parse(json.extensions);
                 }
-                if (!(rs.android) && Blockly.ReplMgr.hasDisallowedIosExtensions()) {
-                    rs.dialog.hide();
-                    top.ReplState.state = Blockly.ReplMgr.rsState.IDLE;
-                    top.BlocklyPanel_indicateDisconnect();
-                    rs.connection = null;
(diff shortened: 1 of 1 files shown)


SusanRatiLane:
I may not be testing the thing you fixed here.

I added an unsupported extension to a project and then verified that I could not connect the iOS Companion. Then I connected to a different project, backed out to the project list, and then loaded the project that had previously been rejected. The project with the unsupported extension did load. Was the goal for this to throw the same error as it would if I attempted to connect directly?

SusanRatiLane (changes requested):
Bug found in testing, described in earlier comment.

PR #3977, opened 2026-06-09
fix: add cleanup for EvConnection observers and streams

vanshapple:
## Summary

Adds explicit cleanup for resources managed by `EvConnection`.

## Changes

* Remove NotificationCenter observers during deallocation.
* Close active input/output streams.
* Remove streams from the run loop.
* Clear stream delegates.
* Clean up existing sessions before creating new connections.

## Motivation

While investigating iOS memory growth and lifecycle management, I noticed that `EvConnection` acquires several long-lived resources (observers, streams, and sessions) but does not currently provide corresponding cleanup logic.

This change improves resource management and reduces the risk of retained objects or callbacks occurring after the object lifecycle has ended.

## Notes

This PR focuses only on lifecycle cleanup in `EvConnection` and does not claim to fully resolve the broader memory-growth investigation.

### DIFF
--- a/appinventor/components-ios/src/EvConnection.swift
+++ b/appinventor/components-ios/src/EvConnection.swift
@@ -61,6 +61,8 @@ class EvConnection: NSObject, StreamDelegate {
   }
 
   func connect(_ accessory: EAAccessory) {
+    cleanupSession()  // Clean up previous session before establishing new connection
+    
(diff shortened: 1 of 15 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

PR #3981, opened 2026-06-10
Cindy listview ios newmethod

cindyloo:
listVIew on iOS needed a tableView and a collectionView for vertical and horizontal orientations. It also need a number of small design updates specifically elementColor in order to pass the testing harness


ewpatton:
Pulling in the other iOS ListView functionality was done in a separate PR so we can close this one now.

Issue #3984, opened 2026-06-11
Clock date parsing returns wrong year on non-Gregorian calendar locales


apexpdl:
**Describe the bug**

On a Thai-language Android or iPhone, the Clock component reads dates with the wrong year. Calling `Clock.MakeInstant("06/10/2026")` and asking for the year back returns `1483` instead of `2026`. This breaks any app that saves and reloads dates, runs countdowns, calculates ages, or schedules alarms.

This is the same class of bug fixed in #3891 (Chart date formatters), but in the Clock component, which has much broader reach since it's used in nearly every tutorial that touches time.

`appinventor/components/src/com/google/appinventor/components/runtime/util/Dates.java` creates `SimpleDateFormat` instances without a `Locale` in three places:

- Line 148, `tryParseDate` — used by `Clock.MakeInstant`, `Clock.MakeInstantFromMillis`, and `Clock.MakeInstantFromParts` when parsing a string.
- Line 202, `FormatDateTime`.
- Line 222, `FormatDate`.

On devices set to a Thai locale, `SimpleDateFormat` resolves against the Buddhist calendar. Parsing is the dangerous side: `Clock.MakeInstant("06/10/2026")` reads `2026` as a Buddhist Era year, which is Gregorian year `1483`. Any app that parses a date string (saved data, hardcoded dates, user input) silently gets an instant about 543 years in the past. Format has the mirror problem — `Clock.FormatDate` produces a Buddhist-year string, which then round-trips wrong.

The same root cause exists on iOS. `appinventor/components-ios/src/Clock.swift` line 16 sets `date.locale = Locale.preferredLanguage`, so a Thai-language iPhone hits the same wrong-calendar parsing.

Note this should **not** affect the `WeekdayName` / `MonthName` style functions, which intentionally use the device locale for display names and should keep their current behavior.

Also worth flagging: `tryParseDate` swallows `ParseException` silently (`catch (ParseException e) {}` on line 149), so when parsing fails it produces an `IllegalArgumentException` with no underlying detail. I'll surface the underlying error in the same PR so any remaining locale-related parse failures are debuggable.

**Affects**

- [x] Android Companion
- [x] iOS Companion
- [x] Android Compiled APK/AAB
- [x] iOS Compiled IPA

**Expected behavior**

`Clock.MakeInstant` with one of the documented format strings (e.g. `MM/dd/yyyy`) should produce the same instant on every device, regardless of locale. Format and parse should round-trip on every locale.

[...]

PR #3891, opened 2026-04-17
Pass Locale.US to Chart SimpleDateFormat constructors

apexpdl:
**What does this PR accomplish?**

*Description*

Chart-related `SimpleDateFormat` constructors in `AxisChartView`, `BarChartDataModel`, and `PointChartDataModel` are called without a `Locale` argument, so they use `Locale.getDefault()` at runtime. On Thai-locale devices the `y` pattern character resolves against `BuddhistCalendar`, producing Buddhist-year labels (e.g. `2569` instead of `2026`). On locales with non-Western digit shaping (Arabic, Persian, Myanmar) the numeric pattern characters render in the local script, which looks inconsistent next to the numeric axis tick values rendered by MPAndroidChart.

This PR passes `Locale.US` explicitly to all six affected `SimpleDateFormat` constructors so the ISO-style format strings are deterministic across all device locales:

- `AxisChartView.java`: `yyyy-MM-dd` and `HH:mm:ss` for axis label formatting.
- `BarChartDataModel.java`: `yyyy-MM-dd` and `HH:mm:ss` for parsing user-provided tuple strings.
- `PointChartDataModel.java`: `dd/MM/yyyy` and `HH:mm:ss` for parsing user-provided tuple strings.

This PR is Android-only. The iOS runtime under `components-ios/` was not examined and may need a follow-up if the same pattern exists there.

*Testing Guidelines*

1. Set an Android device or emulator locale to Thai (Settings → System → Languages → Thai).
2. In App Inventor Designer, create a chart with date-type X values.
3. Run the app in Companion.

Before this PR: axis date labels render with Buddhist-year numbers (e.g. `2569-04-17` instead of `2026-04-17`).

After this PR: axis date labels render with Gregorian-year Western digits regardless of device locale.

**Context for the changes**

This PR touches chart-rendering runtime code that runs on the Android companion, so it targets the `ucr` branch per the contributing guidelines.

If your code changes how something works on the device (i.e., it affects the companion):

- [ ] I have made no changes that affect the companion
- [x] I branched from `ucr`
- [x] My pull request has `ucr` as the base

For all other changes:

- [x] I have made no changes that affect the master branch
- [ ] I branched from `master`
- [ ] My pull request has `master` as the base

**General items:**

- [ ] I have updated the relevant documentation files under docs/
- [x] My code follows the:
    - [x] [Google Java style guide](https://google.github.io/styleguide/javaguide.html) (for .java files)
    - [x] Indentation has been doubled checked
[...]

PR #3988, opened 2026-06-12
Pass Locale.US to Clock date parsing and formatting in Dates.java

apexpdl:

*Description*

Android half of #3984. The iOS half is #3989.

`Dates.java` creates `SimpleDateFormat` instances without a `Locale` in three places: `tryParseDate` (which backs `Clock.MakeInstant`), `FormatDateTime`, and `FormatDate`. On a Thai-locale device, `SimpleDateFormat` resolves against the Buddhist calendar, so `Clock.MakeInstant("06/10/2026")` reads 2026 as a Buddhist Era year and returns Gregorian year 1483. Any app that parses a saved or hardcoded date string silently gets an instant about 543 years in the past. Formatting has the mirror problem, which also breaks round-trips (format a date, store it, parse it later).

This PR passes `Locale.US` to all three, the same approach approved and merged in #3891 for the Chart components. The documented format strings (`MM/dd/yyyy`, `HH:mm`, etc.) fully specify component order, so the locale only pins the calendar and digit script. It does not change date layout for anyone.

It also fixes a debuggability problem in the same function: `tryParseDate` swallowed `ParseException` with an empty catch, so a failed parse threw `IllegalArgumentException` with no detail. The exception now includes the input value and carries the last parse error as its cause.

`WeekdayName`/`MonthName` style functions are intentionally untouched. They use locale display names by design and should keep doing so.

*Testing Guidelines*

1. Set an Android device or emulator to Thai (Settings > System > Languages > Thai).
2. In a test app, call `Clock.MakeInstant("06/10/2026")` then `Clock.Year(...)` on the result.
3. Before this PR: `1483`. After this PR: `2026`.
4. Round-trip check: `Clock.FormatDate` an instant, parse it back with `Clock.MakeInstant`, confirm the same instant on any locale.
5. Regression on a US-locale device: all Clock blocks behave identically to before.

Addresses the Android portion of #3984.

**Context for the changes**

This is Android Companion runtime code, so it targets the `ucr` branch per the contributing guidelines.

- [x] I branched from `ucr`

For all other changes:

**General items:**

- [ ] My code follows the:

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/util/Dates.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/util/Dates.java
@@ -10,6 +10,7 @@ import com.google.appinventor.components.annotations.SimpleFunction;
 import com.google.appinventor.components.annotations.SimpleObject;
 
 import java.util.Date;
+import java.util.Locale;
 import java.text.DateFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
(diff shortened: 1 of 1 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

apexpdl:
@ewpatton this is the Android half of #3984. The iOS half (#3989) merged on Aug 27, so Clock.FormatDate and FormatDateTime now use en_US_POSIX on iOS while Android still follows the device locale. The two platforms disagree until this one lands.

The branch still merges cleanly into ucr, and CI has never run on it, so an "ok to test" would help.

PR #3982, opened 2026-06-10
Chart iOS: pass en_US_POSIX to DateFormatters and fix hh -> HH (#3894)

apexpdl:
**What does this PR accomplish?**

*Description*

Closes #3894 — the iOS counterpart to #3891.

The six `DateFormatter` instances in `AxisChartView.swift`, `BarChartDataModel.swift` and `PointChartDataModel.swift` are created without an explicit `locale`, so iOS resolves them against `Locale.current`. On a Thai-locale device the default calendar is Buddhist, so the `yyyy` pattern produces `2569` instead of `2026`; on locales with non-Western digit shaping (Arabic, Persian, Myanmar) the numeric pattern characters render in the local script.

This PR sets `dateFormatter.locale = Locale(identifier: "en_US_POSIX")` on every affected formatter, and fixes the second issue from the bug report: `AxisChartView.swift` used `"hh:mm:ss"` (12-hour) for the time axis while every other site (and Android) uses `"HH:mm:ss"` (24-hour). It is now `"HH:mm:ss"`.

**Why `en_US_POSIX` and not `Locale.US`:** Apple Technical Q&A [QA1480](https://developer.apple.com/library/archive/qa/qa1480/_index.html) recommends `en_US_POSIX` for fixed-format date code — it is guaranteed Gregorian, Latin-digit, and invariant across user settings and OS versions, whereas `en_US` can still pick up user regional overrides. Because the format strings (`yyyy-MM-dd`, `HH:mm:ss`) fully specify component order, the locale only affects calendar and digit script — it does **not** impose US `mm/dd/yy` layout on anyone. This directly addresses @ewpatton's concern on #3891 that `Locale.US` might make dates appear "wrong" for `dd/mm/yy` users.

*Testing Guidelines*

1. Set an iOS device/simulator locale to Thai.
2. In Designer, create a Bar chart and a Point chart with date-type X values.
3. Run in the iOS Companion.
4. **Before this PR:** axis date labels show Buddhist-year numbers (e.g. `2569-04-17`); time axis shows 12-hour values.
5. **After this PR:** axis labels show Gregorian-year Western digits, time axis is 24-hour, regardless of device locale.
6. Regression on a US-locale device: date labels unchanged; time-axis labels are now 24-hour (matching Android).

Closes #3894.

**Context for the changes**

This is iOS Companion runtime code, so it targets the `ucr` branch per the contributing guidelines.

If your code changes how something works on the device (i.e., it affects the companion):

- [x] I have made no changes that affect the companion

[...]

PR #3989, opened 2026-06-12
Clock iOS: use en_US_POSIX for date parsing and pattern formatting

apexpdl:

*Description*

iOS half of #3984. The Android half is #3988.

`Clock.swift`'s `makeFormat` helper set every `DateFormatter`'s locale to the user's preferred language. For the `DATE_FORMATTERS` parsing array that backs `MakeInstant`, this means a Thai-language iPhone resolves `yyyy` against the Buddhist calendar: `MakeInstant("06/10/2026")` reads 2026 as a Buddhist Era year and returns Gregorian year 1483. Any app that parses a saved or hardcoded date string silently gets an instant about 543 years in the past. `FormatDate`/`FormatDateTime` have the mirror problem, which also breaks round-trips (format a date, store it, parse it later).

The fix: `makeFormat` now takes a `locale` parameter defaulting to `en_US_POSIX`, which is Apple's documented locale for fixed-format date code ([Technical Q&A QA1480](https://developer.apple.com/library/archive/qa/qa1480/_index.html)) and consistent with the Chart fix in #3982. Parsing and pattern-based formatting pick up the deterministic default with no call-site changes.

Three functions intentionally keep the device language, now passed explicitly: `WeekdayName`, `MonthName`, and `FormatTime`. These are localized display functions, and the Android fix (#3988) also left their Android counterparts untouched. Their behavior is unchanged by this PR.

Because the format strings fully specify component order, `en_US_POSIX` only pins the calendar and digit script. It does not impose US date layout on anyone.

*Testing Guidelines*

1. Set an iPhone or simulator language to Thai.
2. In a test app, call `Clock.MakeInstant("06/10/2026")` then `Clock.Year(...)` on the result.
3. Before this PR: 1483. After this PR: 2026.
4. Round-trip check: `Clock.FormatDate` an instant, parse it back with `Clock.MakeInstant`, confirm the same instant on any locale.
5. Confirm `WeekdayName` and `MonthName` still return Thai names on a Thai device (unchanged behavior).
6. Regression on a US-locale device: all Clock blocks behave identically to before.

Addresses the iOS portion of #3984.

**Context for the changes**

This is iOS Companion runtime code, so it targets the `ucr` branch per the contributing guidelines.

- [x] I branched from `ucr`

For all other changes:

**General items:**

- [ ] My code follows the:

### DIFF
--- a/appinventor/components-ios/src/Clock.swift
+++ b/appinventor/components-ios/src/Clock.swift
@@ -9,11 +9,14 @@ import Foundation
  * Creates a `DateFormatter` object for the given `str`.
  *
  * @param str The date format string
+ * @param locale The locale used to resolve pattern symbols. Defaults to
+ *     en_US_POSIX so parsing and pattern formatting are deterministic
+ *     across device locales (see #3984).
  * @returns A new DateFormatter object configured with the given string
  */
-fileprivate func makeFormat(_ str: String) -> DateFormatter {
+fileprivate func makeFormat(_ str: String, locale: Locale = Locale(identifier: "en_US_POSIX")) -> DateFormatter {
   let date = DateFormatter()
-  date.locale = Locale.preferredLanguage
+  date.locale = locale
   date.dateFormat = str
   return date
 }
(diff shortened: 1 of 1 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

ewpatton:

ewpatton (approved):

PR #4002, opened 2026-06-19
Add convenience overload for exporting a project's source zip

zikunz:

*Description*

This adds a two argument exportProjectSourceZip(userId, projectId) overload to FileExporter and FileExporterImpl. It performs the standard source only export and returns the project's .aia archive, delegating to the existing exportProjectSourceZip with the defaults the bulk exporters already use. The two bulk exporters, exportSelectedProjectsSourceZip and exportAllProjectsSourceZip, which spelled the same eleven argument call out inline, now route through the overload as its first callers in the tree. The goal is to give callers that only need a project's .aia a single entry point, rather than repeating the full set of export flags. The new method follows the existing exportProjectOutputFile overload pattern, and the only existing call sites it touches are the two bulk exporters, whose behavior does not change.

*Testing Guidelines*

A new unit test, testExportProjectSourceZipWithDefaults in FileExporterImplTest, builds a project with source and output files, calls the new two argument overload, and asserts that the returned zip contains only the two source files and no remix history. This mirrors the existing testExportProjectSourceZipWithoutHistory test. A second test, testExportProjectSourceZipWithDefaultsNonExistingProject, calls the overload for a project that does not exist and asserts that it throws, mirroring the existing testExportProjectSourceZipWithNonExistingProject test. It can be run with the AiServerLibTests target, for example ant AiServerLibTests with test_name set to com.google.appinventor.server.FileExporterImplTest.

**Context for the changes**

This is a server side change to the appengine module. It does not affect the companion or the device, and it does not change any blocks or designer API. It branches from master.

- [ ] I branched from `ucr`

For all other changes:

**General items:**

- [x] My code follows the:

**Update, August 2026**

This branch now sits on master. The rebase applied with no conflicts, the three files it touches have not changed upstream since the branch was created, and the full server suite passes on the result, including the two tests this change adds. A small tidy commit brings the two new tests in line with the project checkstyle rules, so the files carry no warnings beyond what master already has. A follow up commit then routes the two bulk exporters through the overload, which removes two argument lists that ran past the line length limit and gives the overload its callers in the tree, and the server suite passes again after that change.

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/server/FileExporter.java
+++ b/appinventor/appengine/src/com/google/appinventor/server/FileExporter.java
@@ -43,6 +43,11 @@ public interface FileExporter {
    * keystore, Yail, and screenshots. This suits callers that only need the
    * project's .aia archive.
    *
+   * <p>This method applies no access control. Like the other export methods it
+   * returns the project that belongs to {@code userId}, so any caller that
+   * exposes it to a request must first confirm that the requesting user owns
+   * the project or is an administrator.
+   *
    * @param userId the userId
    * @param projectId the project id belonging to the userId
    * @return the source files as a zip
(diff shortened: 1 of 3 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

PR #4004, opened 2026-06-21
Fix/listview image scroll lag

prabhasg5:

*Description*

Reduces the jerky/laggy scrolling reported for `ListView` rows that use an image layout.

**Root cause:** the image ListView adapters load each row's image inside `onBindViewHolder` via the **synchronous** `MediaUtil.getBitmapDrawable`. This caused lag because every time a row appeared on the screen, the image was loaded, decoded, and resized again from storage. Since this happens on the UI thread, fast scrolling makes the RecyclerView do a lot of heavy work repeatedly, leading to dropped frames and a less smooth experience.

**Fix:** I added an LruCache<String, Drawable> in the shared ListAdapterWithRecyclerView class along with a getImageDrawable(imageName) helper method.

Now, when an image is loaded for the first time, it is stored in the cache. If the same image is needed again while scrolling, it is fetched directly from the cache instead of being decoded from storage again.

The cache is created only when an image-based ListView is used, so text-only ListViews do not use extra memory.

> **Validation note:** I was **not able to reproduce the original jank on my own device** — it scrolled smoothly even with the stock (unmodified) companion and heavy images. The report is on a **Pixel 8a / Android 16**, and the symptom looks device/OS-specific . So I can't provide a measured before/after myself. Per discussion with the maintainer, this change is being put on a **test server** so the **original reporters can try it on the device that actually reproduces the issue**. The change is low-risk and a sound optimization regardless of whether it fully resolves that specific device's symptom.

Reported at: https://community.appinventor.mit.edu/t/listview-with-images-jerky-lagging-on-swipe-scroll/173790

**Context for the changes**

This changes code that runs on the device (the ListView runtime adapters), so it affects the companion 

- [x] I branched from `ucr`

*(No blocks/designer API change — internal runtime optimization only, so no version bump or upgrader is needed.)*

For all other changes:

**General items:**

- [x] My code follows the:

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/ListAdapterWithRecyclerView.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/ListAdapterWithRecyclerView.java
@@ -5,6 +5,12 @@
 
 package com.google.appinventor.components.runtime;
 
+import android.graphics.Bitmap;
+import android.graphics.drawable.BitmapDrawable;
+import android.graphics.drawable.Drawable;
+
+import android.util.LruCache;
+
 import android.view.View;
 import android.view.ViewGroup;
 
@@ -17,7 +23,10 @@ import androidx.core.view.ViewCompat;
 
 import androidx.recyclerview.widget.RecyclerView;
 
+import com.google.appinventor.components.runtime.util.MediaUtil;
 import com.google.appinventor.components.runtime.util.YailDictionary;
+
+import java.io.IOException;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

patryk84a:
I tested this PR, and in my opinion it does not bring any improvement to the behavior of ListView with multiple images.

prabhasg5:
It is adding cache for same images and also different images still matter because decode-and-scale on the UI thread at every rebind, not the duplication, scroll a row and back and the same distinct image gets decoded and resized again from scratch. so the cache holds images which are already decoded and ready!

patryk84a:
Check using this project:

[jjj.aia.zip](https://github.com/user-attachments/files/31600909/jjj.aia.zip)

prabhasg5:
Thanks, that project was a good catch. Tried it on an emulator. The cache is working, but your project still freezes. On a miss it goes and downloads the image, the download already runs on a background thread, but the UI thread then just sits there waiting for it to finish. That wait is what freezes the scroll. The real fix is to stop waiting: let the row draw straight away and put the image in when the download comes back. I'm looking at that now.

patryk84a:
This is a new test project.
[test.aia.zip](https://github.com/user-attachments/files/31660217/test.aia.zip)

PR #4014, opened 2026-06-25
fix to listview AddItem and AddItemAtIndex

cindyloo:


ewpatton:
@cindyloo I think this PR supposed to just be a88fd67? Right now it includes all of the AR stuff, which is not what we would want for this simple fix.

ewpatton:
The fixes to AddItem and AddItemAtIndex were captured in #4048

PR #4018, opened 2026-06-26
List view items not list data

cindyloo:
it's been hard to track all the different changes/fingers in ListView - it should have 5 cases in the TableView, and have _items instead of _listData.

### DIFF
--- a/appinventor/components-ios/src/ListView.swift
+++ b/appinventor/components-ios/src/ListView.swift
@@ -686,165 +686,220 @@ let HORIZONTAL_LAYOUT = 1
     open func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
       let cell = tableView.dequeueReusableCell(withIdentifier: kDefaultTableCell) ??
       UITableViewCell(style: .subtitle, reuseIdentifier: kDefaultTableCell)
+
       let hasElements = _elements.count > 0
-      
-      let listDataIndex = indexPath.row - _items.count
-      
-      if _listViewLayoutMode == 0 { // assume only strings (no dicts]
-        if hasElements {
-          let item = _elements[indexPath.row]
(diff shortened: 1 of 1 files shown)


appinventor/components-ios/src/ListView.swift
```diff
@@ -686,165 +686,220 @@ let HORIZONTAL_LAYOUT = 1
...
+                
+
+            } else {
+              tableView.rowHeight = UITableView.automaticDimension
+              tableView.estimatedRowHeight = 44
+              cell.textLabel?.text = _items[listDataIndex]["Text1"] as? String
+            }
+              
+      cell.textLabel?.numberOfLines = 0
+      cell.textLabel?.lineBreakMode = .byWordWrapping
```
ewpatton, line 900: This logic is already on lines 905-906. Why does it need to be duplicated here? Given the use of optional chaining, it seems like the detail logic could also be moved there.

appinventor/components-ios/src/ListView.swift
```diff
@@ -686,165 +686,220 @@ let HORIZONTAL_LAYOUT = 1
...
+                  verticalStackView.addArrangedSubview(cell.detailTextLabel!)
+                  
+                  // Add the horizontal stack view to the cell's content view
+                  cell.contentView.addSubview(horizontalStackView)
+                  
+                  // Set up constraints
+                  horizontalStackView.translatesAutoresizingMaskIntoConstraints = false
+                  NSLayoutConstraint.activate([
+
+
```
ewpatton, line 839: There are a bunch of extra blank lines in this file that seem unnecessary.

appinventor/components-ios/src/ListView.swift
cindyloo: I was just focused on resolving the listData issue and did a copy/paste from what was previously elsewhere. feel free to remove

PR #4028, opened 2026-06-30
Apply text changes to all DropDownButton faces

ewpatton:
Change-Id: I82bf16633ba41bbe783b5bd3c88063840c3fecc6

This PR fixes a small issue where occasionally the screen dropdown shows the wrong screen name when hovered. The reason being is that `setText(String)` only modifies the current state of the button and there are 6 possible states. Rather than relying on the current face, which could be the hover state, we just change all 6 options simultaneously.

**General items:**

- [ ] My code follows the:


josmas (approved):
LGTM, thanks!

PR #4031, opened 2026-07-01
LTI 1.3 integration so an LMS can assign and grade App Inventor work (design branch)

zikunz:
Please do not merge this. It is exploration, not finished work. I am sharing it so we can look at the direction together.

**Moodle**

https://github.com/user-attachments/assets/a745f871-f235-46da-882c-edae39e44e15

The screenshot below shows what the teacher sees after a student submits, the Moodle grader report with the submission ready to grade. When these recordings were made the teacher side stopped there, and the open question I raised was how a teacher should reach the work itself. That question is settled and the answer is in the branch now. Submit freezes a snapshot of the project, and the teacher opens that snapshot read only from the same gradebook, which is how the classroom portal already handles it. The recordings predate that work, so they do not show the review step.
Screenshot 2026-07-14 at 9 56 32 PM

**Canvas**

https://github.com/user-attachments/assets/1d5d2727-cf13-4539-833f-d1b280c9fdee

*Description*

This adds an experimental integration that connects App Inventor to a Learning Management System through the LTI 1.3 standard, tested end to end against a local Moodle. The flow is that a teacher creates an assignment in Moodle, a student opens it and lands in App Inventor already signed in with the IDE opened on their own project for that assignment, the student works and submits from inside App Inventor, and the grade shows up in the Moodle gradebook.

What is in the change
- An LTI 1.3 tool in the App Inventor server, with OIDC login, launch validation against the platform keys (signature, issuer, audience with authorized party, protocol version, message type, nonce, expiry, and deployment id), and a JWKS endpoint
- Signing the student in during the launch by reusing the existing encrypted App Inventor session cookie, so the security code does not change
- Deriving the account from the platform issuer and subject, so a launch reaches only an LTI provisioned account and never a real login whatever email the platform asserts, and so two first launches that race converge on one account rather than each creating one
- Giving the student their own project on launch, either a copy of a teacher template or a blank project, named after the activity (for example Exercise_2) and kept separate per assignment even when two activities share a title, and opening the IDE on that project by reusing the current project user setting
- A small LTI section in the datastore holding a platform registry, the per user account link, the assignment to project link, the grade line item for each forked project, and one time launch nonces, so several LMS platforms can share one server, a relaunch finds the same project after the teacher renames the activity, and a submission still works after a server restart or on another server instance
- The tool RSA key pair generated on first use and kept in the datastore rather than in the deployed artifact, with a key id in the published key set so a new key can be rotated in
- LTI Dynamic Registration behind a flag, so an administrator can register App Inventor with a platform by pasting one URL rather than copying the endpoints by hand, with a newly registered platform stored disabled until an administrator enables it, and the tool refusing to fetch a link local, private, or metadata host
- Deep Linking, so the teacher picks one of their own projects as the template while adding the assignment. The chosen project is verified to belong to that teacher, the platform return context is held server side under a one time token, and a replayed token is rejected. The template is carried to the student as a reference the tool signs and verifies at fork time, so a custom parameter set outside this flow carries no valid signature and is ignored
- Grade passback to the LMS through Assignment and Grade Services, targeting the line item of the assignment the student is submitting, so a student with several graded assignments never posts to the wrong one
- A Submit to LMS item in the Project menu. It posts the current project with a custom header that a page on another site cannot set, so only a submission started inside App Inventor is accepted, the project is checked to belong to the submitting student, and a plain browser visit only shows a short page pointing back to the menu
- A frozen copy of the template taken when the teacher picks it, owned by a reserved account nobody can sign into, and an assignment fixed to that copy by its first learner, so every learner on one assignment starts from the same project however the teacher edits or reselects afterwards. This follows what the classroom portal does
- A frozen copy of the student project taken at the moment of submit, owned by a reserved account the student cannot sign into, so the teacher grades a fixed artifact even if the student keeps working afterwards
- A read only review launch, so the teacher opens that frozen copy from the place where they are grading. Only a frozen copy is ever opened, never the learner's own project, because a session for that would have to run as the learner's own account, which holds their work for every course on the platform. It is the Submission Review message of the LTI specification, and it carries the learner named by the for_user claim. Only an instructor may use it, the project is checked to belong to that learner, and the session it builds carries a read only flag and a one project limit that the client honors, while what actually confines the session is the reserved account it runs as. This follows what the classroom portal already does
- The item is shown only on a project that came from an LTI launch, read from a marker the launch writes on the forked project, so it does not appear in any other user's Project menu

This is a spike. The in flight OIDC launch state lives in memory. The platform registry is a datastore table, seeded on first use from the single platform still described by flags, so an existing flag setup keeps working and more platforms can be added to the table. The pages the tool serves, the template picker and the short status pages, share one accessible and mobile ready style, and each page opened in an LMS popup can dismiss itself. The teacher cannot yet open the template read only, which the classroom portal offers and which is not in this change. A teacher can still pick a different template in the LMS after a student has started, because the specification does not let a tool refuse that, so the tool holds the starting point steady instead and every learner on one assignment keeps the copy the first learner fixed. I have run the whole loop against a local Moodle and a local Canvas. The same tool serves both with no code change, because LTI is an open standard, and the recording for each is above.

*Testing Guidelines*

Run a local Moodle and register this tool in it as an LTI 1.3 external tool, using the login, launch, and JWKS URLs from the App Inventor server. Set the tool client id and the platform endpoints as flags on the dev server. As a teacher, create an assignment in Moodle and pick a template through Select content. As a student, open the assignment, confirm the IDE opens on the assignment project, work, and use Submit to LMS from the Project menu. As a teacher, enter a grade and confirm the student sees it.

JUnit tests cover the token signing and verification, the naming of the student project, the namespaced account key, and the LTI datastore section, which is the platform registry, the account link, the one time nonce, the assignment to project link, the grade context, and the tool key store. They also cover the dynamic registration request, the address filter that refuses an internal or metadata host, the instructor gate and audience check, and the grade passback endpoint that keeps the line item query string.

On top of those come the hardening tests added since the last review (the exact instructor role match, the issued at freshness check, the two hundred only response rule, the state spent only after the token verifies, the loopback and transport policy, the review subject and its identity mapping, and the submission snapshot records), together with tests that drive the request paths themselves rather than their helpers, covering which project a read only review is allowed to open, every refusal on the submit endpoint, and the launch marker that decides whether Submit to LMS appears in the Project menu, 132 tests in total, and they pass.

A further hardening round moved the cross account copy filter into the copy service itself, so the generated Yail never reaches another account and every caller is covered, made the platform hear about a submission only when its frozen copy is stored and read back as the copy on record, retired the project an assignment stops pointing at so it can no longer be handed in, and moved the bookkeeping of what is saved and what is still on its way into one browser free coordinator the submission waits on. Each of those refusals was checked by removing the guard it covers and confirming the matching test fails.

The launch, the fork idempotency across renames and restarts, the routing of two same titled assignments to separate projects, the submit, the Deep Linking ownership check, and the one time token were also exercised against a real Moodle, and the launch, the submit, and the grade passback were exercised against a real Canvas as well. The review path was exercised against a real Moodle by following the link the gradebook offers, which reaches the frozen copy read only and refuses the live project. Canvas does not send the Submission Review message today, so that entry point does not appear there. The package is clean under the project checkstyle configuration.

**Context for the changes**

This branch is rebased onto the Java 17 master. Upstream replaced the project level trash flag with a folder in the user settings while this was in review, so the LTI code now asks the storage layer which projects are in the trash. That lookup had to answer for an account whose settings hold no folder tree at all, which is exactly the account an LTI launch creates, because otherwise a learner reopening an assignment gets an error instead of their project.

This is exploration built directly on master. It adds an LTI package under the appengine server, a small LTI section in the storage layer (new datastore entities and the methods that read and write them, following the existing nonce and password data patterns), the client half of Submit to LMS (the menu item and its gate, a browser free save coordinator the editor manager delegates to, and the settings plumbing that carries the assignment marker), and the servlet registrations in web.xml. It also teaches the shared project copy service to leave generated Yail behind on any copy that crosses accounts, which covers every caller of that service. It does not touch the components module or anything that runs on the device.

- [ ] I branched from `ucr`

For all other changes:

**General items:**

- [x] My code follows the:
- [ ] `ant tests` passes on my machine (the server, shared, and client suites pass on the Java 17 master, the blockly suite needs a browser this machine does not have)

I have followed the Java style guide, double checked the indentation, kept the diff free of unnecessary formatting changes, and verified the package is warning free under lib/checkstyle/appinventor-checks.xml. The LTI JUnit tests pass and a setup page for administrators was added under docs. This branch now sits on the Java 17 master, and the whole server test suite passes there, 28 classes and 225 tests with no failure and no error. The client suite passes at 44 tests, 20 of them added by this branch as plain JUnit around the save coordinator and its arithmetic, which is possible because the coordinator holds no reference to the browser. A GWT test case still cannot ride along, because the client target compiles every GWT test case into one shared module and only the first one is reachable from the generated runner, so the lines that hand the coordinator to the editor are verified by reading and by the end to end run. The marker the Project menu reads is covered on the server, where the launch writes it and a copy arrives without it.


AppInventorWorkerBee:

AppInventorWorkerBee:

PR #4023, opened 2026-06-29
Redesign ListView ListData editor with a thumbnail-forward layout

prabhasg5:
**What does this PR accomplish?**

*Description*

Redesigns the ListView **List Item Editor** — the dialog behind a ListView's
**ListData → "Click to Add/Delete Data"** button — into a cleaner,
thumbnail-forward layout. This is a UX/visual refinement that builds on the
earlier UiBinder migration of the editor rows (#3990); it changes only how the
dialog is presented, not what it saves.

- **Thumbnail-forward rows:** each item shows an image thumbnail with its
  filename on the left, a prominent main-text field over a lighter detail-text
  field, and a compact delete control.
- **Click-to-pick image:** clicking a row's thumbnail opens an asset picker that
  previews each project image and updates the thumbnail, instead of editing a
  bare filename.
- **Dialog chrome:** a live item count in the (draggable) title bar, an
  "Add an item" button, and a Save / Cancel footer.

The saved JSON contract `[{Text1, Text2, Image}]` is unchanged, and per-layout
behavior is preserved (the detail/image fields still show or hide according to
the selected ListViewLayout). The accessible `Dialog` wrapper (ARIA role, modal,
Esc-to-close, focus traps, focus restoration) is reused unchanged, and the new
thumbnail and picker controls are keyboard operable.

**Before**
Screenshot 2026-06-29 at 8 02 58 PM

**After**
Screenshot 2026-06-29 at 8 05 59 PM

**Behavior (video)**

https://github.com/user-attachments/assets/a3f64b32-4bbd-4d00-9d86-c9f8a3d837f9

This video also demonstrates the A11y accessibility features are not overridden.

Resolves # .

**Context for the changes**

Designer/appengine-only change — it does not affect the companion, so this PR is
based on `master`.

- [x] I have made no changes that affect the companion

- [ ] I branched from `ucr`
- [ ] My pull request has `ucr` as the base

Further, if you've changed the blocks language or another user-facing designer/blocks API (added a SimpleProperty, etc.):

- [ ] I have updated the corresponding version number in appinventor/components/src/.../common/YaVersion.java
- [ ] I have updated the corresponding upgrader in appinventor/appengine/src/.../client/youngandroid/YoungAndroidFormUpgrader.java (components only)
- [ ] I have updated the corresponding entries in appinventor/blocklyeditor/src/versioning.js

For all other changes:

- [ ] I have made no changes that affect the master branch

- [x] I branched from `master`
- [x] My pull request has `master` as the base

[...]

PR #4032, opened 2026-07-01
Add keyboard navigation to the ListView ListData editor

prabhasg5:

*Description*

Adds full keyboard navigation to the ListView **ListData** editor — the pop-up
opened by the *Click to Add/Delete Data* button on a ListView's `ListData`
property — so item data can be entered and moved through entirely from the
keyboard, with no mouse.

This builds on the earlier migration of the editor's rows from a virtualized
`CellTable` to real focusable widgets (`TextBox`/`FocusPanel`/`Button`). With
real widgets in place, this PR wires up the key handling:

| Key | Where | Action |
|---|---|---|
| Tab | any control | next control (forward) |
| Shift+Tab | any control | previous control (back) |
| Enter | main / detail text field | append a new blank row + focus its main field |
| ↓ Arrow Down | text field or thumbnail tile | same column, next row |
| ↑ Arrow Up | text field or thumbnail tile | same column, previous row |
| ← / → | text field | move caret in the field |
| Enter / Space | thumbnail tile | open image picker |
| Enter / Space | delete button | delete the row |
| Esc | anywhere in dialog | close dialog |

Notes:
- Tab/Shift+Tab already worked once the rows became real widgets; this PR adds
  Enter (append + focus), Up/Down (column-preserving row navigation), and makes
  the **Add Row** button also focus the new row.
- Up/Down are safe to repurpose because the fields are single-line text boxes,
  where those keys are otherwise unused; Left/Right are left to move the caret.
- The logic lives in the two files that own the widgets — no new files.
  `ListViewDataRow` detects the keys on its own widgets and reports them through
  a small `Callbacks` interface; `ListViewAddDataDialog` makes the cross-row
  decisions (append a row, focus a sibling).
- **No data-contract change:** the saved `ListData` JSON (`[{Text1, Text2,
  Image}]`) is byte-for-byte identical. Designer/appengine only — no companion
  or runtime impact.

**Context for the changes**

Designer/appengine-only change to a property editor. It does not touch the
components runtime or the companion, and adds no new blocks/designer API (no
`SimpleProperty`, so no `YaVersion`/upgrader/`versioning.js` changes). Branched
from `master`, base `master`.

- [ ] I branched from `ucr`

For all other changes:

**General items:**

- [x] My code follows the:

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
@@ -5283,6 +5283,8 @@ public interface OdeMessages extends Messages, ComponentTranslations {
     "<tr><td>Reset Connection</td><td>Alt + Shift + R</td></tr>" +
     "<tr><td>Refresh Companion Screen</td><td>Alt + R</td></tr>" +
     "<tr><td>Navigate Components in components tree</td><td>↑/↓</td></tr>" +
+    "<tr><td>Add a row in the ListView data editor</td><td>Enter</td></tr>" +
+    "<tr><td>Navigate rows in the ListView data editor</td><td>↑/↓</td></tr>" +
     "<tr><td>Open this dialog</td><td>Alt + ?</td></tr>" +
     "</tbody>" +
     "</table>")
--- a/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
@@ -5285,6 +5285,7 @@ public interface OdeMessages extends Messages, ComponentTranslations {
     "<tr><td>Navigate Components in components tree</td><td>↑/↓</td></tr>" +
     "<tr><td>Add a row in the ListView data editor</td><td>Enter</td></tr>" +
     "<tr><td>Navigate rows in the ListView data editor</td><td>↑/↓</td></tr>" +
+    "<tr><td>Move to the next / previous field in the ListView data editor</td><td>Tab / Shift+Tab</td></tr>" +
     "<tr><td>Open this dialog</td><td>Alt + ?</td></tr>" +
     "</tbody>" +
     "</table>")
--- a/appinventor/appengine/src/com/google/appinventor/client/editor/youngandroid/properties/ListViewAddDataDialog.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/editor/youngandroid/properties/ListViewAddDataDialog.java
@@ -111,13 +111,26 @@ public class ListViewAddDataDialog {
     updateCount();
   }
 
-  /** Creates a data row wired to refresh the item count when it deletes itself. */
+  /** Creates a data row wired to the dialog's count refresh and keyboard navigation. */
   private ListViewDataRow newRow(JSONObject item) {
     return new ListViewDataRow(item, showDetail, showImage, imageChoices, imageUrls,
-        new Runnable() {
+        new ListViewDataRow.Callbacks() {
           @Override
-          public void run() {
+          public void onRowDeleted() {
             updateCount();
+          }
+
+          @Override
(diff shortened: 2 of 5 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

SusanRatiLane:
Things I noticed in testing: 
1. Is there a reason not to use the css class for the green button we are using in the other dialogs? It looks like that might just be gwt-button.
2. The data entry flow doesn't quite work. When I hit enter to start a new row, the focus is on the first text field. This means I have to shift-tab to back up to the icon if I'm using a layout that has one. I realize that I asked for the focus to be on the first text field, but I was imagining the current wizard with the icon on the far right. Yours looks better, but perhaps focus should start on the icon. I realize that is a drop-down rather than a text field, but maybe it will work.

Other thoughts: Enter opens and closes the drop-down for the icon selection with that is in focus, but it creates a new row when a text field is focus. This does seem like what you expect to happen intuitively, so it's probably fine. It just means you can only create a new row with enter if you have a text field highlighted. If we can get some testers to give feedback on the usability, that would be great. That's on me. I'll see what I can do.

prabhasg5:
> 1. Is there a reason not to use the css class for the green button we are using in the other dialogs? It looks like that might just be gwt-button.

Agreed - I should have checked for an existing one before styling this, I dug in and found the shared one is " ode-TextButton " which looked to be grey but turned out that the neo theme actually making it green and I was momentarily confused. that button styling actually came in with the thumbnail-forward redesign (#4023), not the keyboard-nav work. The fix itself is small - switch Save to ode-TextButton and delete the now-unused lie-btn-primary. but i also have 2 more questions on how to handle it: 
1) fold this into the current keyboard-nav PR, or do it as a small separate follow-up ?
2) Cancel button — should I also move Cancel to ode-TextButton for full consistency or keep its current ?

prabhasg5:
> 2\. The data entry flow doesn't quite work. When I hit enter to start a new row, the focus is on the first text field. This means I have to shift-tab to back up to the icon if I'm using a layout that has one. I realize that I asked for the focus to be on the first text field, but I was imagining the current wizard with the icon on the far right. Yours looks better, but perhaps focus should start on the icon. I realize that is a drop-down rather than a text field, but maybe it will work.

Done — new-row focus now starts on the first cell: the icon for layouts that
have one, otherwise the main text field

SusanRatiLane:
>     1. fold this into the current keyboard-nav PR, or do it as a small separate follow-up ?

I think it's fine to do here. I did notice when I looked at the previous PR but forgot about it, do that's on me. No reason to make it extra complicated

>     2. Cancel button — should I also move Cancel to ode-TextButton for full consistency or keep its current ?

I think this would match the rest of the system, so let's make both buttons the same.

prabhasg5:
Yeah on it. I'll push ASAP

prabhasg5:
@SusanRatiLane  I just pushed the modified code

SusanRatiLane:
This is working very well. 

One issue: if I tab down to the Cancel/Save buttons, I'm stuck in a tab-loop between Cancel, Save, and Add Row. The up and down arrows don't navigate back to the data rows. Is there a way to get back to the rows from the buttons via keyboard? If not, there should be. The arrow buttons seem logical, but let me know if you have other ideas.

prabhasg5:
> This is working very well. 
> 
> 
> 
> One issue: if I tab down to the Cancel/Save buttons, I'm stuck in a tab-loop between Cancel, Save, and Add Row. The up and down arrows don't navigate back to the data rows. Is there a way to get back to the rows from the buttons via keyboard? If not, there should be. The arrow buttons seem logical, but let me know if you have other ideas.

The tab + shift will get to navigate those back up to rows then we can use arrow keys. I didnt touched that because its already existing before this, if we have to change it to something else we can work on it.

SusanRatiLane (approved):
LGTM

PR #4041, opened 2026-07-20
Fix ListView search filtering and selection (iOS + Android)

prabhasg5:

*Description*

Fixes ListView search-filtering and selection bugs on both platforms.

**iOS** (`components-ios/src/ListView.swift`)
- The search box now actually filters the list. Previously the filtered results
  were computed but never read by `numberOfRows`/`cellForRowAt`, so typing had no
  effect on the vertical list. Matching is now case-insensitive and matches
  anywhere in the text (was prefix-only + case-sensitive) and covers both the
  main and detail text.
- Selection now maps the tapped visible row back to its real position, so
  selecting while filtered picks the correct item — on both the table and the
  horizontal collection view.
- Also enables tapping plain-text items in the horizontal list (previously
  commented out) and fixes a fallback layout branch that indexed with a value
  that could go negative.

**Android** (`components/src/.../ListAdapterWithRecyclerView.java`)
- Filtering no longer clears the current selection — it changes what is visible,
  not what the user picked. Selection is stored against original item indexes, so
  it survives filtering and stays correct whether or not the selected item is
  currently shown. A small pair of helpers (`toDisplayPosition` /
  `toOriginalPosition`) handles the display↔original mapping in one place.

**Selection-on-filter policy:** filtering keeps the selection (does not clear it),
matching the common data-table pattern where selection is tracked independently
and survives filtering. Applied consistently on both platforms. (Out of scope:
data changes — add/remove/set Elements — still clear selection on Android via
`updateData`; unchanged here.)

No user-facing designer/blocks API changed (no new properties or blocks), so no
version bump or upgrader is required.

*Testing Guidelines*

- **iOS:** build the `AIComponentKit` scheme, run the Companion in the simulator
  with a ListView. Type in the search box → the vertical list filters
  (case-insensitive, substring); filter then tap → the correct item is selected;
  repeat with horizontal `Orientation`. Manually verified.
- **Android:** run the Companion on an emulator with a ListView (ShowFilterBar
  on). Select an item → type in the search box → the selection stays; filter then
  tap → correct item selected; clear the filter → selection still highlighted.
  Manually verified. `ant tests -Dskip.ios=true` passes (including `ListViewTest`).

**Context for the changes**

Both changes are to the ListView component runtime that runs on the device (the
companion), so this targets the `ucr` branch.

- [x] I branched from `ucr`

For all other changes:

**General items:**

- [x] My code follows the:

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/ListAdapterWithRecyclerView.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/ListAdapterWithRecyclerView.java
@@ -79,7 +79,9 @@ public abstract class ListAdapterWithRecyclerView
     @Override
     protected void publishResults(CharSequence charSequence, FilterResults filterResults) {
       items = new ArrayList<>((List<Object>) filterResults.values);
-      clearSelections();
+      // Filtering changes what is visible, not what the user picked, so the selection is left
+      // alone here. Selection is stored against original item indexes, so it stays correct
+      // whether or not the selected item is currently shown.
       notifyDataSetChanged();
       // We store the original data in the originalItems variable
       // We store the original item indexes in the originalPositions variable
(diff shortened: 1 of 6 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

appinventor/components/src/com/google/appinventor/components/runtime/ListAdapterWithRecyclerView.java
```diff
@@ -79,7 +79,9 @@ protected FilterResults performFiltering(CharSequence charSequence) {
     @Override
     protected void publishResults(CharSequence charSequence, FilterResults filterResults) {
       items = new ArrayList<>((List<Object>) filterResults.values);
-      clearSelections();
+      // Filtering changes what is visible, not what the user picked, so the selection is left
+      // alone here. Selection is stored against original item indexes, so it stays correct
+      // whether or not the selected item is currently shown.
```
SusanRatiLane: Does this mean that someone could have a selected item that is invisible because it's filtered out? I'm not sure if that's a good idea.

prabhasg5: I made that with the intention of keeping the selection and filtering independent but this raises another issue which u've mentioned now. so, maybe a middle ground between these - Keep the selection if the item is still visible and clear it only if the filter hides it. does that work ?

SusanRatiLane: > I made that with the intention of keeping the selection and filtering independent but this raises another issue which u've mentioned now. so, maybe a middle ground between these - Keep the selection if the item is still visible and clear it only if the filter hides it. does that work ?

I like that idea. Go for it.

prabhasg5: Pushed. Selection now survives filtering while the item is still on screen, and clears only when the filter hides it. — on both Android and iOS.

SusanRatiLane: > Pushed. Selection now survives filtering while the item is still on screen, and clears only when the filter hides it. — on both Android and iOS.

In testing, I saw the expected behavior on Android, but not on iOS. Here's the test app.
[ListView_randomized.aia.zip](https://github.com/user-attachments/files/30233391/ListView_randomized.aia.zip)

SusanRatiLane (approved):
LGTM

PR #4037, opened 2026-07-09
Introduce ListDataModel to own the ListView item list

prabhasg5:
**What does this PR accomplish?**

*Description*

This PR is a pure internal refactor of the ListView component on **both Android
and iOS**. It introduces a `ListDataModel` that becomes the single owner of the
ListView's list data, so the data lives in one place instead of being scattered
across the component.

**There are no behavioral changes — only the *ownership* of the data moves.** In
particular, **the filter and selection behave exactly as before** (including
their current quirks) — this PR does not touch the filtering or selection logic
on either platform. Only *where the data lives* changed. Consolidating filtering
and selection into the model is a follow-up PR.

**Android** — `ListView.java` + new `ListDataModel.java`
- Previously `ListView.java` held its own `ArrayList` while the row adapter kept
  its own copy of the same data; there was no single source of truth.
- New `ListDataModel.java` owns the item list and its mutations (add / addAt /
  addAll / addAllAt / remove / setItems / clear) plus reads (get / size /
  isEmpty / getItems).
- `ListView.java` now accesses the list through `dataModel.*`; the adapter is
  handed `dataModel.getItems()` — the same live reference as before.
- `ListAdapterWithRecyclerView` and the six layout adapters are untouched.

**iOS** — `ListView.swift` + new `ListDataModel` class (within `ListView.swift`)
- Previously the backing arrays (`elements`, `listData`) and the filtered
  `results` were fields on the ListView view class.
- A new `ListDataModel` class now owns those three, and `ListView` routes all
  access through a single `_model` instance.
- The two arrays, the filter, the selection, and the `reloadData()` update path
  are all unchanged — a straight relocation of storage. (Unifying the arrays and
  moving filtering/selection into the model is the follow-up.)

No user-facing designer/blocks API changed (no new SimpleProperty, etc.), so no
version bump or upgrader is required.

*Testing Guidelines*

- **Android:** `ant tests -Dskip.ios=true` passes, including `ListViewTest`
  (`testFilter`, `testSelectionRemovalWithDictBasedElements`,
  `testMultipleListViewEvents`). In the Companion, exercise each data path: set
  Elements / ElementsFromString / ListData; AddItem / AddItemAtIndex / AddItems /
  AddItemsAtIndex / RemoveItemAtIndex; type in the filter bar; select rows —
  data and UI stay consistent and behave exactly as before.
- **iOS:** builds and runs (AIComponentKit / AICompanionApp schemes); the same
  data paths behave identically in the simulator, with the filter and selection
[...]

PR #4053, opened 2026-07-27
Move Android ListView filtering and selection into ListDataModel

prabhasg5:

*Description*

The ListView keeps three things that aren't visual: the list of items, the
search filter, and which row is selected. Until now those were split across
different classes — the item list lived in `ListDataModel` (added in #4037),
but the filter and the selection still lived inside the row adapter. This PR
finishes that consolidation by moving the filter and the selection into
`ListDataModel` too, so there is a single place that owns all of it. The
adapter is now just a thin view that reads from the model and draws the rows.

This is an internal cleanup, not a feature change. There are no new blocks or
properties, and filtering and selection behave the same as before: the search
bar narrows the list, tapping selects a row, and a selection is kept while its
item is still visible under a filter and cleared once the filter hides it.

There is one small, intentional improvement: the selected row's highlight now
stays put when a color or font is changed at runtime. Previously, changing a
property rebuilt the list and quietly dropped the highlight even though
`SelectionIndex` still reported the old value; now the highlight and the
reported value stay in agreement.

Because the filter and selection logic now live in one plain class instead of
inside the adapter, they can be tested on their own. This PR adds
`ListDataModelTest`, which checks the filter (including matching on both the
main and detail text), the mapping from a filtered row back to its real item,
and the "keep selection while visible, clear it when hidden" behavior. The
existing `ListViewTest` still passes unchanged.

*Testing Guidelines*

Automated: `ant tests` (the relevant classes are `ListDataModelTest` and
`ListViewTest`).

Manual (Android companion): add a ListView with several items, turn on the
filter bar, and set a Label to `ListView.Selection` in `AfterPicking`.
- Type in the filter bar → the list narrows to matching rows (matches anywhere
  in the main or detail text).
- Tap one of the filtered rows → the label shows that exact item.
- Select a row, then keep typing so it's filtered out → the highlight clears;
  delete back so it's shown again → it comes back unselected.
- Select a row, then change BackgroundColor or FontSize at runtime → the
  highlight stays.

**Context for the changes**

This changes code that runs on the device (the companion), so it is based on
`ucr`.

- [x] I branched from `ucr`

(No user-facing API changed — no new properties or blocks — so no version bump
or upgrader is needed.)

For all other changes:

**General items:**

- [ ] I have updated the relevant documentation files under docs/ (no user-facing change, so no docs update needed)
- [x] My code follows the:

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/ListAdapterWithRecyclerView.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/ListAdapterWithRecyclerView.java
@@ -17,98 +17,46 @@ import androidx.core.view.ViewCompat;
 
 import androidx.recyclerview.widget.RecyclerView;
 
-import com.google.appinventor.components.runtime.util.YailDictionary;
-
-import java.util.ArrayList;
-import java.util.List;
-
 public abstract class ListAdapterWithRecyclerView
     extends RecyclerView.Adapter<ListAdapterWithRecyclerView.RvViewHolder> implements Filterable {
   protected static final String LOG_TAG = ListView.LOG_TAG;
-  
+
   protected ClickListener clickListener;
 
   protected int backgroundColor;
   protected int selectionColor;
   protected float radius;
-  protected List<Object> items = new ArrayList<>();
-  protected List<Object> originalItems = new ArrayList<>();
-  protected List<Integer> originalPositions = new ArrayList<>();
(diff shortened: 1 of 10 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

SusanRatiLane:

SusanRatiLane (approved):
LGTM

PR #4058, opened 2026-07-31
Force CustomAlertView title color to be black on iOS

ewpatton:
Change-Id: I59e4d7db641d864facc455fc4b8cc26de6a098f5

As reported [via the community](https://community.appinventor.mit.edu/t/how-to-change-notifier-title-text-color-for-ios/174953?u=ewpatton): The title text of the Notifier is white on a light background. Specifically, I think this only occurs when the phone is in dark mode, since the default background color would be darker. We use a custom UIView hierarchy for our alert view, and that is always lighter in color. Now, it might be better to dynamically adjust the background based on the fact that the phone is in dark mode, but for the sake of simplicity I figured it was easier to force the title text to be reasonable.


mark-friedman (approved):
LGTM

PR #4062, opened 2026-08-06
Stop rebuilding the ListView adapter on Android, and let iOS rows size themselves

prabhasg5:

PR #4064 should be merged before this PR.

*Description*

Two fixes to the same component, one per platform. Both are about view state being
thrown away or ignored when it did not need to be.

**Android — an appearance change no longer rebuilds the adapter**

Right now, changing any appearance property of a ListView (TextColor, FontSize,
ElementColor, and so on) throws away the whole adapter and builds a new one. The
RecyclerView loses every row view and starts drawing again from the first item,
so the user's scroll position jumps back to the top of the list. If you are 30
rows down and a block changes the text color, you end up back at row 1.

The reason is where the styling happens. Each of the six row adapters sets the
colors, fonts and alignments inside `onCreateViewHolder`. RecyclerView only calls
that about ten times in total, because it reuses row views as you scroll. So
there was no way to restyle the rows that already exist, and replacing the whole
adapter was the only thing that worked.

This PR moves the styling to `onBindViewHolder`, which runs every time a row is
drawn. The appearance values now live in a small new class, `ListViewStyle`. The
ListView owns one of them and the adapter keeps a reference to that same object,
so changing a value and asking RecyclerView to re-bind the rows is enough. No new
adapter, and the scroll position stays where the user left it.

**iOS — rows size themselves from their own content**

The iOS list pinned every row to a height derived from the font size, through
`heightForRowAt`. A row's own content had no say, so a tall image or wrapped text
was squeezed into a fixed box instead of making the row taller. Rows now size
themselves: `rowHeight` is `UITableView.automaticDimension` and `heightForRowAt`
is gone, because implementing it makes UIKit ignore `rowHeight` entirely. Every
layout already pins its content to the contentView's top and bottom, so Auto
Layout has what it needs. `estimatedHeightForRowAt` stays — an estimate is
allowed to be approximate.

**Context for the changes**

This changes component runtime code that runs on the device, so it is based on
`ucr`.

- [x] I branched from `ucr`

(No properties were added, removed or changed, so no version bump or upgrader is
needed.)

For all other changes:

**General items:**

- [x] My code follows the:

(The component runtime suite is green — `ListViewTest` 5/5 and the iOS
`ListViewTests` 31/31. The appengine and buildserver suites fail locally for
setup reasons unrelated to this change, and `MapTest.testLoadFromURL` needs
network.)

### DIFF
--- a/appinventor/components-ios/src/ListView.swift
+++ b/appinventor/components-ios/src/ListView.swift
@@ -129,6 +129,8 @@ fileprivate final class ListViewRootView: UIView {
     _view.tableFooterView = UIView()
     _view.backgroundView = nil
     _view.backgroundColor = argbToColor(_backgroundColor)
+    // Rows size themselves from their own constraints, whatever the layout mode.
+    _view.rowHeight = UITableView.automaticDimension
 
     // Auto height for the table (existing)
     _automaticHeightConstraint = _view.heightAnchor.constraint(equalToConstant: kDefaultTableCellHeight)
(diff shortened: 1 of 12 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

SusanRatiLane:

prabhasg5:
PR #4064 should be merged before this PR.

patryk84a:
Will we still be able to add our own adapter from the extension after this change?

prabhasg5:
> Will we still be able to add our own adapter from the extension after this change?

Yes, setListAdapter is still public, so a custom adapter can still be installed. But you're right that there's a gap: the adapter now needs ListDataModel and ListViewStyle, and both are private, so an extension can't create it. I'll add public getDataModel() and getStyle() here so the hook is usable again. tell me if they need more than these two.

PR #4063, opened 2026-08-06
Add a Submit to LMS menu item and track pending editor saves so a submission waits for them

zikunz:

*Description*

This adds a Submit to LMS item to the Project menu, together with the machinery that decides when it is allowed to appear. It is the client half of submitting an assignment from inside App Inventor, split out of the design branch #4031 so it can be taken on its own.

The item fixes the project at the click, saves the open editors, and posts that project id to a server submit endpoint only once nothing at all is still on its way to the server. Saving first matters because work in an open editor reaches the server on a timer, five seconds after the learner stops typing and within about thirty five seconds of the first change, so posting straight away would hand in whatever the last automatic save had caught rather than what the learner is looking at.

What still has to reach the server and what is already on its way now lives in a small browser free coordinator, PendingSaves, that the editor manager delegates to. It sends one save per file at a time, because two saves of one file can land in either order and the older one landing last would undo the newer one. It holds a file back while an earlier save of it is still running and restarts the idle timer, and the next save pass sends the held back save once the earlier one has answered. A closing window is the one exception, where everything waiting goes out beside the unanswered save, because no later pass can come for it there. It puts a failed save back in the queue without a clock of its own, the way the file path has always retried, and it ignores a completion that arrives after its save was discarded, so a late answer cannot clear a newer save of the same file. Settings saves answer their completion on every path, including failure and read only, and a settings edit made while a save is in flight is no longer lost when that save succeeds.

The request carries a custom header that a page on another site cannot set without a preflight, which the server side work refuses to grant, so once that work lands only a submission started from inside App Inventor is accepted, and a guard keeps a double click from sending two submissions. The request carries a timeout, which the underlying builder leaves off by default, a longer deadline covers the save that runs before it, and ending an attempt also ends its request, so a call that never answers cannot leave the item unusable and a retry can never overlap a request this client still holds open. Every path that can end a submission goes through one place that releases the guard and reports exactly once, and each attempt carries a number so a late answer from an attempt that already ended cannot report over a newer one.

The user sees a short progress message and then success or failure, reported through the existing ErrorReporter.

The item is gated on an invisible per project setting, LtiLaunched, that the server side assignment work writes on the project it creates for a learner at launch. The gate requires a writable project carrying that marker to be open in the designer, and it is refreshed when a project loads, when the file menu updates, and when the view or the open editor changes. Nothing in this change writes the marker, so the item stays hidden on every project until the server side work lands. The save path it goes through does change for everyone, which is where review is most useful, and the behaviour it changes is described above.

Splitting the client half out also gives the classroom portal work a base to reuse. The item, the gate, and the request guard stay the same whatever sits behind the endpoint, so a different server side flow only needs to point the same item at its own address.

*Testing Guidelines*

The full build passes with this change applied, including the GWT compile and link of the client module. The server suite reads ninety three tests, the shared suite twenty two, and the client suite forty four, with no failure and no error in any of them, and sixteen of the client tests cover the coordinator. The coordinator tests are mutation tested, meaning each rule was removed or inverted in a copy of the code and the suite went red each time, so the tests hold the behaviour rather than mirror the code. The lines that hand the coordinator to the editor and the menu run only inside the GWT client, which no unit test here can start, so those lines are verified by reading and by the end to end run, and the test javadocs say which is which. The touched files carry no checkstyle warning that master does not already have. On a build that includes the server work, the item appears only on the project a launch created, which was exercised end to end against a real Moodle on the design branch.

**Context for the changes**

This is exploration split out of #4031. The endpoint the item posts to arrives with the server side work, and until then the menu item is invisible everywhere, so this change has no user visible effect on its own.

- [ ] I branched from `ucr`

For all other changes:

**General items:**

- [x] My code follows the:
- [ ] `ant tests` passes on my machine (the server, shared, and client suites pass, the blockly suite needs a browser this machine does not have)

**Update.** A review pass found that the first version posted before the open editors were saved, and that the request had no timeout. Both are fixed in the second commit. A further pass found holes in how the second commit waited. A submission could go out while an automatic save that began earlier was still running, two saves of one file could land in either order and let the older content arrive last, and a failed settings save never answered its completion. The third commit moves the bookkeeping into the coordinator described above and closes each of these. One residual is worth naming rather than hiding. If the deadline expires while the server is still working, the learner is told the submission failed even though it may yet succeed, and a retry then produces a second frozen copy. The attempt now cancels its own request when it ends, which narrows that window, the deadline is generous enough that it should only follow a genuine fault, and the server side work treats a later submission as superseding an earlier one.

A late review pass found one more hole worth naming. The window closing save went through the same hold back, and a closing page has no later save pass, so an edit held back at that moment was never sent at all, where the old code would still have issued its racy request. Saving on a closing window now takes everything, held back or not, since no later pass exists there and the identity checks already keep a late answer from an older save from clearing the newer one. Two coordinator tests cover it, and the moved javadoc that had drifted from its method was put back where it belongs.

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/Ode.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/Ode.java
@@ -475,6 +475,7 @@ public class Ode implements EntryPoint {
     hideChaff();
     hideTutorials();
     currentView = TRASHCAN;
+    getTopToolbar().updateSubmitToLmsMenuItem();
     ProjectListBox.getProjectListBox().loadTrashList();
     projectToolbar.enableStartButton();
     projectToolbar.setProjectTabButtonsVisible(false);
@@ -490,6 +491,7 @@ public class Ode implements EntryPoint {
     hideChaff();
     hideTutorials();
     currentView = USERADMIN;
+    getTopToolbar().updateSubmitToLmsMenuItem();
     deckPanel.showWidget(userAdminTabIndex);
   }
 
@@ -1289,6 +1291,7 @@ public class Ode implements EntryPoint {
    */
   public void setCurrentFileEditor(FileEditor fileEditor) {
     currentFileEditor = fileEditor;
+    getTopToolbar().updateSubmitToLmsMenuItem();
     if (currentFileEditor == null) {
       // nothing more we can do
       LOG.info("Setting current file editor to null");
--- a/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/OdeMessages.java
@@ -543,6 +543,22 @@ public interface OdeMessages extends Messages, ComponentTranslations {
   @Description("Name of Export Project menuitem")
   String exportProjectMenuItem();
 
+  @DefaultMessage("Submit to LMS")
+  @Description("Menu item for submitting the current work to the LMS that launched it over LTI.")
+  String submitToLmsMenuItem();
+
+  @DefaultMessage("Submitting to your LMS...")
+  @Description("Shown while a submission to the LMS over LTI is in progress.")
+  String submittingToLms();
+
+  @DefaultMessage("Submitted to your LMS. Your teacher will grade it there.")
+  @Description("Shown after the current work is submitted to the LMS over LTI.")
+  String submitToLmsSuccess();
+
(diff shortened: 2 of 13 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

PR #4059, opened 2026-08-01
[iOS] Make ListDataModel own the ListView's list mutations and selection mapping

prabhasg5:
**What does this PR accomplish?**

*Description*

Part of the ListView `ListDataModel` consolidation series, following #4037 and
the Android consolidation. iOS side, one source file plus tests.

**The model owns mutations.** `ListDataModel`'s two backing arrays are now
`private(set)`, and it gained the mutator API its Android counterpart already
has — `clear` / `setItems` / `append` / `insert` / `remove`. `ListView` no
longer reaches in and mutates the arrays directly at ~14 call sites. The model
also gained `item(at:)` / `mainText(at:)` / `detailText(at:)`, so callers no
longer branch on whether a row came from the string array or the dictionary
array; that branch is gone from every read site, including the collection-view
cell renderer.

No blocks-language or designer API changes — no new properties, methods or
events, so no version bump or upgrader is needed.

**Context for the changes**

If your code changes how something works on the device (i.e., it affects the
companion):

- [x] I branched from `ucr`
- [x] My pull request has `ucr` as the base

For all other changes:

- [x] I have made no changes that affect the master branch

**General items:**

- [x] My code follows the:
    - [x] [Swift style guide](https://google.github.io/swift/) (for .swift files)
    - [x] Indentation has been doubled checked
    - [x] This PR does not include unnecessary changes such as formatting or white space
- [x] `ant tests` passes on my machine

PR #4064, opened 2026-08-09
Sync the iOS ListDataModel consolidation from master into ucr

prabhasg5:
The two iOS files from #4059, brought to ucr unchanged.

#4059 was merged into master, which did not yet carry the Android consolidation. My branch was cut from ucr, so the PR diff against master also showed the ten Android files ucr already had — they were the base showing through, not new work. Ten of those twelve files are byte-identical on ucr today; only ListView.swift and ListViewTests.swift are missing.

The effect is that the iOS ListView on ucr is still pre-consolidation, so the change isn't reaching the companion. No new code here — the content was reviewed and approved as #4059. The unrelated one-line Notifier.swift change from a different master commit is deliberately excluded.

### DIFF
--- a/appinventor/components-ios/src/ListView.swift
+++ b/appinventor/components-ios/src/ListView.swift
@@ -23,10 +23,11 @@ let HORIZONTAL_LAYOUT = 1
  * model instead of holding its own fields.
  */
 class ListDataModel {
-  /// Simple string items (populated for string lists).
-  var elements = [String]()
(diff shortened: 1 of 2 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

SusanRatiLane (approved):
LGTM. Keeping iOS ListView changes synced on ucr is important for this project. We need to reexamine our process for that.

SusanRatiLane:

Issue #2620, opened 2021-12-16
Convert internationalization of blocks to interpolated strings


SusanRatiLane:
After most of our custom blocks were in production, Blockly rolled out a change to support internationalization with interpolated strings. This allowed a block with several inputs to be translated as a single string, and the inputs could be rearranged to suit the grammar of the language. 

Dictionaries is translated in this way, so it can be a model for future work.

If the strings for blocks are changed to interpolated strings, all our existing translations will need to be converted. One way to do this would be to run a script on our existing translations to stitch the separate strings together into an interpolated string that represents what displays on the block now. Translators can then rearrange things to suit the grammar of the language.

Yash621:
@SusanRatiLane can i work on this ?

PR #4065, opened 2026-08-09
 Convert single-input Lists blocks to interpolated strings

lyszt:

*Description*

Converts the single-input Lists blocks to Blockly interpolated message strings, following the Dictionaries blocks as the model.

Blocks converted:

- is empty
- pick a random item
- copy
- is a list?
- reverse
- list to csv row
- list to csv table
- list from csv row
- list from csv table
- all but first
- all but last

For each block:

- Replaced the appendField chain with interpolateMsg.
- Added the stitched interpolated string (existing fragments joined with the %1 placeholder, preserving current on-block wording) across all 22 messages files under `appinventor/blocklyeditor/src/msg/ai_blockly/`.
- Pointed typeblock at the existing clean title key.
- Removed the now-dead single-use fragment keys (verified 0 remaining references).

Scope is limited to the simple Lists blocks to keep this reviewable. This is part of #2620.

For all other changes:

**General items:**

- [ ] My code follows the:


AppInventorWorkerBee:

AppInventorWorkerBee:

josmas:
Hi there @lyszt, thanks for the contribution. Can you please explain the reasoning behind the changes?
Also, how can you verify that all the language translations are correct?

lyszt:
Hello, @josmas !

I based this on issue #2620 by SusanRatiLane. Using interpolation allows us to use strings like **"is a list? thing
%1"**, where **%1** can receive a socket. So translators can move words around for their language and it isn't so rigid.

 As for translations, I did not translate or reword anything, each string is just the existing pieces joined the
 same way they already were before. I did notice some are a bit off, like "est une liste?" in French, but
 I left those alone to avoid making changes too broad. If you wish I can take a look at those specific lines.

ewpatton:
I think this is technically correct given that the way the block is constructed ends up causing the input (represented by "%1") to be at the end. Eventually, translators can move the input if it is needed but for the time being the blocks shouldn't change shape nor semantics at all.

PR #4067, opened 2026-08-11
Feature/listview Multiselect (Android + iOS)

prabhasg5:
**What does this PR accomplish?**

**PR #4062 should be merged before this PR**

*Description*

Makes the ListView's `MultiSelect` property visible in the designer and the blocks
on both Android and iOS, and adds a `SelectedItems` property so an app can read
everything the user picked.

**Context for the changes**

This changes component runtime code that runs on the device on both platforms, so
it is based on `ucr`.

If your code changes how something works on the device (i.e., it affects the companion):

- [ ] I have made no changes that affect the companion

- [x] I branched from `ucr`
- [x] My pull request has `ucr` as the base

Further, if you've changed the blocks language or another user-facing designer/blocks API (added a SimpleProperty, etc.):

- [x] I have updated the corresponding version number in appinventor/components/src/.../common/YaVersion.java
- [x] I have updated the corresponding upgrader in appinventor/appengine/src/.../client/youngandroid/YoungAndroidFormUpgrader.java (components only)
- [x] I have updated the corresponding entries in appinventor/blocklyeditor/src/versioning.js

For all other changes:

- [x] I have made no changes that affect the master branch

- [ ] I branched from `master`
- [ ] My pull request has `master` as the base

**General items:**

- [x] I have updated the relevant documentation files under docs/
- [x] My code follows the:
    - [x] [Google Java style guide](https://google.github.io/styleguide/javaguide.html) (for .java files)
    - [x] [Google JavaScript style guide](https://google.github.io/styleguide/jsguide.html) (for .js files)
    - [x] [Swift style guide](https://google.github.io/swift/) (for .swift files)
    - [x] Indentation has been doubled checked
    - [x] This PR does not include unnecessary changes such as formatting or white space
- [ ] `ant tests` passes on my machine

PR #4068, opened 2026-08-14
Feature/ListView replace an item in place (Android + iOS)

prabhasg5:

**PR #4067  should be merged before this PR**

*Description*

Adds `UpdateItemAtIndex` to the ListView on both Android and iOS, so an app can
change one row without rebuilding the list.

**Selection behaviour**

The replaced row stops being selected. A different item occupies that position
afterwards, so a selection pointing there no longer refers to what the user
picked. Nothing moves, so the selections around it are untouched.

`MultiSelect` needed a second rule, found while testing on the device. Dropping
the replaced row and clearing `Selection` / `SelectionIndex` with it left the
component reporting nothing selected while `SelectedItems` still held the rest
of the set — and a follow-up update driven by `SelectionIndex` then hit the
bounds check and raised an error. Reporting now falls back to the most recent
pick that remains, so `SelectionIndex` reads 0 only when `SelectedItems` really
is empty. Single selection is unchanged: replacing the one picked row leaves
nothing to fall back to.

**Both platforms**

`ListDataModel` gains a replace operation on each side — `set(index, item)` in
Java, `replace(at:with:)` for both backing arrays in Swift — along with
`lastSelection` next to the existing `firstSelection`. The unit tests mirror
each other, so the two platforms are held to the same contract rather than
merely looking alike.

**One refactor**

On Android, `AddItem` and `AddItemAtIndex` each carried the same block deciding
whether a new row should be a plain string or a dictionary, and
`UpdateItemAtIndex` would have been a third copy. That decision now lives in a
single `newRowFrom` helper, which also mirrors the `usesPlainStrings` /
`normalizedListItem` pair the iOS side already had.

- [x] I branched from `ucr`

For all other changes:

**General items:**

- [x] My code follows the:

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/client/youngandroid/YoungAndroidFormUpgrader.java
+++ b/appinventor/appengine/src/com/google/appinventor/client/youngandroid/YoungAndroidFormUpgrader.java
@@ -1674,6 +1674,10 @@ public final class YoungAndroidFormUpgrader {
       // Added SelectedItems property.
       srcCompVersion = 12;
     }
+    if (srcCompVersion < 13) {
+      // Added UpdateItemAtIndex method.
+      srcCompVersion = 13;
+    }
     return srcCompVersion;
   }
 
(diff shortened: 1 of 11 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

PR #4076, opened 2026-08-26
Do not send vestigial YAIL files to the buildserver

### LINKED ISSUE #3939: Do not send vestigial YAIL files to buildservers
**Describe the bug**

[From the forum](https://community.appinventor.mit.edu/t/unable-to-compile-because-screen-doesnt-exist/173136?u=ewpatton): Deleting a screen involves removing 3 files: the designer (scm), the blocks (bky), and the generated YAIL (yail). However, if a delete operation partially succeeds the scm and bky files will be gone, but not the yail. If the screen in question wasn't compilable, then there is no recourse to delete the YAIL file since as far as the client is concerned there isn't a screen by that name. At a minimum, we should filter out YAIL files without a corresponding scm/bky file. It would also be prudent to clean up the vestigial YAIL files for cost savings but not critical to addressing the negative user experience of being unable to remove the YAIL files themselves.

**Affects**

- [ ] Designer
- [ ] Blocks editor
- [ ] Projects Explorer
- [ ] Android Companion
- [ ] iOS Companion
- [ ] iOS Compiled IPA
- [x] Buildserver
- [ ] Debugging

**Expected behavior**

Projects that contain a YAIL file in the server but without the corresponding scm/bky files should not cause a build to fail. Instead, we should filter out the YAIL file when creating the ZIP bundle to be sent to buildserver.

**Steps to reproduce**

Not sure. This happens rarely and is likely due to some hiccup in the production systems that cause some of the files to be removed but not others. It might be possible to edit an AIA file exported as an admin to remove the scm/bky and then reimport it but this needs to be investigated.

aarushkandukoori:

*Description*

A partially failed screen deletion can leave a project with a vestigial `.yail` file whose corresponding `.scm`/`.bky` files are gone. The client does not show such a screen, so the user has no way to remove the file, and the build fails because the buildserver receives a YAIL file for a screen that does not otherwise exist (see the forum thread linked in #3939).

Following the approach suggested in the issue discussion, `YoungAndroidProjectService.build()` now checks, before packaging the project for the buildserver, that each `src/**/*.yail` file still has a corresponding `.scm` or `.bky` file. Vestigial YAIL files are deleted (with a log warning) before the export:

- Because the check runs at `build()` time, it protects every `StorageIo` implementation. (`ObjectifyStorageIo.exportProjectSourceZip` already filters incomplete screen file sets for buildserver exports, but that logic is not shared by other storage backends.)
- Deleting the file, rather than only filtering it out of the ZIP, also cleans up the vestigial YAIL files left over from before, per the cost-savings note in the issue.
- A YAIL file is only treated as vestigial when *both* the `.scm` and the `.bky` files are missing — the "(or .bky)" reading of the suggestion in the issue — so a screen that still has either of its editable files is never touched. This is deliberately narrower than the screen-count filter in `ObjectifyStorageIo`.

*Testing Guidelines*

- `ant AiServerLibTests` — new unit tests in `YoungAndroidProjectServiceTest` cover the helper: an orphaned YAIL file is flagged; complete screens are kept; a YAIL file is kept when either an `.scm` or a `.bky` file still exists; files outside `src/` are ignored.
- Manual: remove `ScreenX.scm` and `ScreenX.bky` from a project at the storage level while leaving `ScreenX.yail` (the failure mode @ewpatton describes in the issue). Building the project previously failed; with this change the build succeeds and the vestigial `ScreenX.yail` is deleted from the project.
- Local `ant tests` note (macOS): every suite passes except the GWT client test `OdeTest.testCompareLocales`, which fails identically on unmodified `master` in my environment (JDK 11 and 17), so it is unrelated to this change.

Fixes #3939.

**Context for the changes**

- [ ] I branched from `ucr`

For all other changes:

**General items:**

- [x] My code follows the:
- [x] `ant tests` passes on my machine (1533 JUnit + 140 Karma tests, `BUILD SUCCESSFUL`; the only exception is the pre-existing `OdeTest` environment error noted above, which is identical on unmodified `master`)

### DIFF
--- a/appinventor/appengine/src/com/google/appinventor/server/project/youngandroid/YoungAndroidProjectService.java
+++ b/appinventor/appengine/src/com/google/appinventor/server/project/youngandroid/YoungAndroidProjectService.java
@@ -82,10 +82,13 @@ import java.net.ConnectException;
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Properties;
+import java.util.Set;
 import java.util.TreeSet;
 import java.util.logging.Logger;
 
@@ -548,6 +551,18 @@ public final class YoungAndroidProjectService extends CommonProjectService {
     for (String buildOutputFile : buildOutputFiles) {
       storageIo.deleteFile(userId, projectId, buildOutputFile);
     }
+
+    // A partially failed screen deletion can leave behind a vestigial YAIL file
+    // with no corresponding .scm/.bky files. The client does not show such a
+    // screen, so the user has no way to remove the file, and sending it to the
+    // buildserver makes the build fail. Delete any orphaned YAIL files before
+    // packaging the project.
+    List<String> sourceFiles = storageIo.getProjectSourceFiles(userId, projectId);
+    for (String yailFile : findVestigialYailFiles(sourceFiles)) {
+      LOG.warning("Deleting vestigial YAIL file " + yailFile + " in project " + projectId);
+      storageIo.deleteFile(userId, projectId, yailFile);
+    }
+
     URL buildServerUrl = null;
     ProjectSourceZip zipFile = null;
     try {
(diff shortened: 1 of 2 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

jisqyv:

PR #4078, opened 2026-08-29
Added RAM and disk caching for images in ListView.

patryk84a:

*Description*
ListView has problems with large numbers of images when scrolling. The view freezes. I tested another fix for this problem, but I don't think that fix fixed it. I've released my own fix, and in my tests, it works fine.

*Testing Guidelines*

Compile this project on AppInventor Server and in the one compiled from this PR and compare the scrolling quality.
[test.aia.zip](https://github.com/user-attachments/files/31660132/test.aia.zip)

**Context for the changes**

- [x] I branched from `ucr`

For all other changes:

**General items:**

- [ ] My code follows the:

### DIFF
--- a/appinventor/components/src/com/google/appinventor/components/runtime/Form.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/Form.java
@@ -328,6 +328,12 @@ public class Form extends AppInventorCompatActivity
   }
 
   @Override
+  public void onLowMemory() {
+    super.onLowMemory();
+    MediaUtil.clearImageCache();
+  }
+
+  @Override
   public void onCreate(Bundle icicle) {
     // Called when the activity is first created
     super.onCreate(icicle);
--- a/appinventor/components/src/com/google/appinventor/components/runtime/ListViewImageSingleTextAdapter.java
+++ b/appinventor/components/src/com/google/appinventor/components/runtime/ListViewImageSingleTextAdapter.java
@@ -1,11 +1,10 @@
 // -*- mode: java; c-basic-offset: 2; -*-
-// Copyright 2025 MIT, All rights reserved
+// Copyright 2026 MIT, All rights reserved
 // Released under the Apache License, Version 2.0
 // http://www.apache.org/licenses/LICENSE-2.0
 
 package com.google.appinventor.components.runtime;
 
-import android.graphics.drawable.Drawable;
 import android.view.Gravity;
 import android.view.View;
 import android.view.ViewGroup;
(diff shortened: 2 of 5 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

josmas:
Hey @prabhasg5 any chance you can take a look at this PR? would appreciate your input here.

prabhasg5:
> Hey @prabhasg5 any chance you can take a look at this PR? would appreciate your input here.

sure, will take a look and share my thoughts.

prabhasg5:
I opened a Pr against the same issue a couple of weeks back upon discussion with @SusanRatiLane but as i can't reproduce the issue i just added cache and susan agreed to put that on a test server so users can give feedback when they have issue. ig @patryk84a is mentioning about that pr in his description. the picsum is currently down rn, so i cant actually see images and also can't access the test app he mentioned as github says "not found". anyways i saw the code and what i observed is that the 5 second time out is too aggressive, the current comment explicitly mentioned that "we don't cache image files from the internet to temp files. The image at a particular URL, such as an image from a web cam, may change over time." and final thing cache writes aren't atomic - we can use .tmp file until the cache is downloaded fully then we can remove the .tmp. These are just my observations from the code. I haven't been able to reproduce the original issue or test the changes with the test app, so there may be some context I'm missing.

patryk84a:
During testing, I made several videos of the application in action:
- this is a test of the current version of ListView:

https://github.com/user-attachments/assets/12a3b953-b56d-4f46-848d-14c32aeea4fa

(ListView with PR 4004 behaves similarly)

- this is a test with ListView with my fix when we open the application for the first time:

https://github.com/user-attachments/assets/f43db807-3a3e-485f-94fb-caf120d605d5

- and when we open the application again:

https://github.com/user-attachments/assets/0d4ccc36-dca0-414c-86db-7ef80bad21fb

There's a significant difference between the caching and non-caching versions. The example contained 2,000 photos downloaded from Picsum. With in-memory caching, the difference is only slight, so it's negligible, but the images I downloaded were 300x300px, not huge by today's standards. I think that with larger files, caching in internal memory would have a noticeable advantage.

Downloading to a TMP file first and then renaming it is a good idea. But if the file on the server changes, that's a problem. We can add a property in the ListView that gives the user a choice, as applications can vary and so can caching requirements.

Picsum actually doesn't work, I updated the test project using another graphics site.

prabhasg5:
I tested this with the test app provided, and it genuinely works. I just have one question: has `HttpResponseCache` been considered? It would provide correct revalidation and atomic writes for free, and it would address the “what if the file changes?” question without a need of any user choice. other than that it is working and good in my perspective.

patryk84a:
Thanks for this class. I'll take a closer look at it. I see it's available from API 13, so for a minimum API 14 in appinventor it should be fine.

patryk84a:
I implemented HttpResponseCache. It requires less code than manually saving. It works well.

prabhasg5:
Sorry for the late response! I just tested the behavior and the code, and everything is working fine based on my testing.

patryk84a:
@ewpatton Is this fix acceptable, or should something be changed so that it can be accepted? The reason is that the ListView with multiple images is currently very slow.

PR #4097, opened 2026-09-11
Include comment bubbles when downloading blocks as an image

### LINKED ISSUE #892: Comment bubbles no longer captured in PNG with Blockly update
[See the forum post](https://groups.google.com/forum/#!category-topic/mitappinventortest/8X_r-XuEiQ4). Previously blocks and bubbles were in the same group element. With the Blockly update we are only capture the blocks group and not the bubble group. Ideally we should capture both to provide consistent behavior with older versions of App Inventor.

preetvadaliya:

*Description*

"Download Blocks as Image" only captured the block canvas. Since the Blockly update, comment bubbles live on a separate bubble canvas, so every comment was silently missing from the PNG (#892). Comment text is also an HTML `<textarea>` inside a `<foreignObject>`, which browsers do not reliably draw into a canvas (Safari refuses), so adding the bubble canvas alone would not have been enough.

| | Before | After |
|---|---|---|
| Block comment bubbles (Download Blocks as Image) | missing | included, with tail and text |
| Block comment bubbles (single block: right-click → Download Blocks as PNG, or drag out) | missing | included, for the block and the blocks attached to it |
| Workspace comments | box with a black top bar, no text | top bar, body and text in the theme colours |
| Image bounds | blocks only | grow to fit bubbles and tails |
| Mutator bubbles, warnings, backpack | not exported | still not exported |
| Comment text in the SVG | `<foreignObject>` | `<text>`/`<tspan>` lines wrapped to the box |

How it works:

- `workspaceExportGroup()` builds the export from the block canvas plus the comment bubbles from the bubble canvas. The two canvases share one transform, so their children line up once it is removed; the union of their bounds becomes the image size.
- Each `<foreignObject>` is replaced with a background rect in the theme's comment colour plus SVG text wrapped to the same width, measured with the textarea's real font. Colours are inlined because the theme exposes them as CSS variables the exported SVG cannot see.
- Dangling `filter` references are removed (a missing filter makes browsers skip the element).
- The old page-wide `textarea.innerHTML = value` workaround is gone.
- Comment bubbles are found through each block's comment icon, so the whole-workspace export and the single-block export (`exportBlockAsPng`) share one routine. The whole-workspace export includes every open bubble; the single-block export includes the bubbles of that block and its children only.
- All three callers use the new path: the download menu, the single-block PNG, and the project screenshot (`getUri`).

*Testing Guidelines*

1. Open a project, go to Blocks, add a comment to a block and keep the bubble open. Right-click the workspace → Download Blocks as Image. The PNG contains the bubble, its tail and the text.
2. Right-click the workspace → Add Comment and type a long line. Download again: the note appears with its top bar and wrapped text.
3. Right-click a block that has a comment → Download Blocks as PNG (or drag the block out to the desktop). The PNG contains the block and its bubble; a comment on a *different* block is not included. Dragging the PNG back onto the workspace still recreates the block.
4. Close the bubbles and download: blocks only, as before. Open a mutator and download: the mutator is not in the image.
5. `ant tests`: 10 new cases in `exportBlocksImage_tests.js` (bubble included, one tspan per line, wrapping stays inside the box, bounds cover the tail, stray bubbles excluded, no-comment bounds unchanged, no `foreignObject` or `filter` in the SVG, the SVG loads as an image, and the single-block export includes only that block's bubbles).

Fixes #892.

**Context for the changes**

Branched from `master`; no companion changes.

For all other changes:

**General items:**

- [ ] I have updated the relevant documentation files under docs/ (not applicable: the image export has no page under docs/)
- [x] My code follows the:

### DIFF
--- a/appinventor/blocklyeditor/src/exportBlocksImage.js
+++ b/appinventor/blocklyeditor/src/exportBlocksImage.js
@@ -66,6 +66,52 @@ goog.provide('AI.Blockly.ExportBlocksImage');
     return css;
   }
 
+  var SVG_NS = 'http://www.w3.org/2000/svg';
+
+  /**
+   * Assembles the part of a workspace that belongs in an exported image: the
+   * block canvas plus the comment bubbles from the bubble canvas. The two
+   * canvases share one transform, so with it removed their children line up.
+   * Mutator bubbles, warning bubbles and the backpack flyout also live on the
+   * bubble canvas and are deliberately left out.
+   * @param {!Blockly.WorkspaceSvg} workspace The workspace to export.
+   * @return {{group: !Element, bbox: {x: number, y: number, width: number,
+   *     height: number}}} A detached group and its bounds in workspace units.
+   */
+  out$.workspaceExportGroup = function(workspace) {
+    var blockCanvas = workspace.getCanvas();
+    var bubbleCanvas = workspace.getBubbleCanvas();
+    var group = document.createElementNS(SVG_NS, 'g');
+    group.setAttribute('class', 'blocklyBlockCanvas');
+    var blocks = blockCanvas.cloneNode(true);
+    blocks.removeAttribute('transform');
+    group.appendChild(blocks);
+
+    var box = blockCanvas.getBBox();
+    var bounds = {left: box.x, top: box.y, right: box.x + box.width,
+                  bottom: box.y + box.height};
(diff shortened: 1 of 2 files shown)


AppInventorWorkerBee:

AppInventorWorkerBee:

ewpatton:

